PR #4, opened 2020-05-07
Fixed memory leaks

alexnask:


alexnask:
By the way, let me know if you have some style guides, dos and do nots etc.  
I personally use `zig fmt` but this repo doesn't seem to so I'm refraining from letting the muscle memory take over.

alexnask:
This seems to have fixed the vast majority of leaks, I will test editing more files and seeing how many allocations are counted by the leak counter.  

Also we no longer leak with every RPC :D

alexnask:
Also, I don't think it's a good idea to use an ArenaAllocator as our base allocator for this program, since we don't have clear points where we want to clear it all and refill it but we rather need to shrink/realloc memory frequently, this will waste a lot of memory.  

In practise you can see that by cutting and pasting text around or rearranging lines, we use more and more memory even without the leaks fixed by this PR.  

Ideally we would use some good general purpose allocator (with a freelist etc) for the base allocator but one is not provided by std yet.  
I propose we use std.heap.page_allocator as the base allocator for now, since most of the allocation happen by local Arenas (e.g. zig.parse, some of the new code on this PR etc.) and switch to a better allocator when the Generic Composable allocators PR is merged in zig.  

I tested this and the memory usage is much better, even for more files open the base memory usage is down and adding a lot of text + moving things around requires barely more memory (e.g. I copied std.mem.Allocator.alloc for about 1000 lines of code and needed 0.1 MB more memory, from 0.8 to 0.9 MB while with less files open and just cut pasting and minor edits I went up to 40 MB of memory fast using the page allocator).  

Please let me know if that sounds reasonable, I can add it to this PR.

SuperAuguste:
That sounds very reasonable, so you can definitely add it to the PR! Thanks so much for your hard work fixing everything!

alexnask:
Cheers, I will add it then.  
Please merge the other PR if it works for you first, since there will be a couple of conflicts I will have to merge with this one :)

SuperAuguste:
Looks perfect, merging now!

PR #5, opened 2020-05-07
Switched to anonymous struct and enum literals

alexnask:
Also added trailing commas to struct literal fields that are followed by newline.


SuperAuguste:
This look great! I'm gonna test it and merge it.

PR #6, opened 2020-05-07
Separate document text and memory.

alexnask:
Text is now a substring of memory, starting at index 0.
We now realloc memory when needed and only copy the existing data around, instead of allocating a new string, copying all the data and freeing the old string.
This leads to less allocation calls and less copying.  

We also now check duplicate open calls, free the memory of the previously open text document if they happen and log it.


SuperAuguste:
How much does this improve memory usage? I just want to make sure this pull request's complexity is a fair trade off to its improvements.

alexnask:
@SuperAuguste  

This mainly reduces the time for executing changes, especially small ones, at the cost of a bit of extra memory (how much memory we preallocate can be tuned).  
By preallocating more memory when a change needs a bigger buffer, subsequent range changes do not need to allocate at all and only need to do 2 copies (as opposed to 1 allocation, 3 copies and 1 deallocation each time for every range change like before).  

I think this is worth it since changes are usually typing or small copy pastes which would cause a lot of allocation calls in a row, as opposed to a bigger upfront allocation followed by "just" copying and less than in master.

SuperAuguste:
OK, sounds good! Merging!

PR #9, opened 2020-05-08
Added position context detection

alexnask:
Added a function that takes a document and a position and returns a PositionContext.  
Currently we recognize comment, string literal, builtin, variable access and field access contexts.  

This fixes the completions showing up in comments and string literals (at least they do in vscode) and will probably be useful later.  
It could be enhanced by becoming a `union(enum)` that returns context specific data (for example, we could return a string of what we are dot accessing for the field access context to do some analysis).  

It also correctly detects `@"..."` variable/field accesses.  

Also removed a `*const Self` parameter, zig passes all arguments as const so leaving it as `Self` will cause the compiler to choose the optimal way to pass it (either const ref or by value if it is small enough).

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -387,6 +387,8 @@ fn documentContext(doc: types.TextDocument, pos_index: usize) PositionContext {
                     string_pop_ctx = access_ctx;
                 }
                 context = .builtin;
+            } else {
+                context = .other;
             }
             continue;
         }
--- a/src/main.zig
+++ b/src/main.zig
@@ -368,7 +368,7 @@ fn documentContext(doc: types.TextDocument, pos_index: usize) PositionContext {
             continue;
         }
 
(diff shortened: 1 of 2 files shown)


SuperAuguste:
This is fantastic! Gonna be really useful for the analysis engine, too!

alexnask:
Forgot to rename `documentContext` to `documentPositionContext`, can you commit that (so that I dont open a trivial PR)?  
Thanks a lot.  
I tested the code with `const io_mode = .evented;` on windows btw and it works (obviously it just turns read and write calls to basically `await async <read/write>()` but it will be trivial to rewrite some stuff to take advantage of async io when compiled with it)

alexnask:
We also need to support the multiline string literal syntax `\\ sring literal`, this should be the same as the whole comment line case but checking for backslash.

PR #13, opened 2020-05-09
Removed no_snippets build option, added configuration parsing.

alexnask:
For now, we only have a single configuration option, `enable_snippets`, which defaults to true.  

I added the infrastracture so we can easily add new options as we build features and don't have to switch a big number of build options or add a number of cases to the code later on.  
I am passing the config around as a parameter rather than a global, that way we may extend functionallity to allow different configs per workspace/project or client.  

I am using std.debug.warn here for an error diagnostic instead of log, since this is not a response to an rpc request but should still be visible to the end user somehow, idk if this is the best way.  

Note that I use `config.zig` directly as a struct type, it seems better than defining a Config struct in this file.  

Please check this PR out ffirst if you can since the other one is a bit more involved and will require some changes to adapt with this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,10 +16,18 @@ zig build
 
 | Option | Type | What it Does |
 | --- | --- | --- |
-| `-Dno_snippets` | `bool` | Enables "no snippets" mode for compatibility with language clients that do not support snippets (such as ale). |
 | `-Ddata_version` | `string` | The data file version. Any files in the `src/data` file that correspond with the Zig version you want the language server to build for (0.6.0, master).
 
 Then, you can use the `zls` executable in an editor of your choice that has a Zig language server client!
+
+### Configuration options
+
+You can configure zls by providing a zls.json file in the same directory as the executable.  
+The following options are currently available.  
+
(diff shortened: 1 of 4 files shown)


alexnask:
I tried to integrate `std.fs.Watch`, unfortunately it's out of date (at least for Windows but I think for all platforms), fixed all of its compile errors but somehow that crashed the compiler.  
I will need to rewrite it for std at some point so we can use it here :D

SuperAuguste:
This looks great! I think that making `config` a global variable (like `allocator`) might prove useful in reducing argument bulk in our functions. Thoughts?

alexnask:
@SuperAuguste  

I made it non-global in case we want to support multiple configs for different workkspaces etc. but I can make it global if you like.

SuperAuguste:
No, good call for multiple workspaces, didn't really think of that.

alexnask:
Cool, I'm going to go ahead and resolve conflicts now.

SuperAuguste:
The branch has a few conflicts, so I can't really merge it rn.

PR #16, opened 2020-05-12
Snippet improvements

alexnask:
Correctly render comptime, noalias and vararg arguments.  
Only render spaces in return type tokens after commas, this allows us to generate types like `anyframe->void` better, although it probably is still not 100% ideal.


SuperAuguste:
Looks good!

PR #18, opened 2020-05-13
Unified completion generation into a single function

alexnask:
Did that this independently of your last commit :P  
Also added yet another small snippet improvement, starting to work on imports now.


SuperAuguste:
This looks amazing!

PR #19, opened 2020-05-13
Basic import support

alexnask:
🥳

Issue #7, opened 2020-05-07
Improve tests


alexnask:
Sooner or later we will need to add tests to check for regressions.  

I propose the following three kinds of tests:
- Function unit tests (mainly for the parsing functions)  
- I/O testing: simulate a client sending requests from a different process and check the responses are correct  
- Internal state testing: simulate a client sending requests from a different process and check our internal state (number of documents, text of documents, latest sane_text of documents etc.)

SuperAuguste:
Looks interesting. Not sure if it's a priority right now, but definitely down the line.

alexnask:
Started adding session tests, coverage going up 🎉

palash25:
hi i recently started learning zig and this seems like a beginner-friendly issue to get started with but when i try to run unit tests using `zig test unit_test.zig` i get an error saying 

```
./unit_tests.zig:2:15: error: unable to find 'types'
const types = @import("types");
              ^
./unit_tests.zig:23:43: note: referenced here
fn makeUnnamedDocument(text: []const u8) !types.TextDocument {
                                          ^
./unit_tests.zig:31:21: note: referenced here
    const doc = try makeUnnamedDocument(final_line);
                    ^
./unit_tests.zig:68:5: note: referenced here
    try testContext(
    ^

```

am i running the tests the wrong way ?  i tried looking for docs on testing and also tried looking in the ci workflow for a test job but couldnt find anything.

PS i am able to build the repo locally

data-man:
> am i running the tests the wrong way ?

Yes. Try execute `zig build test` in zls root folder.

SuperAuguste:
I think #609 closed this. Closing in favor of #623.

PR #20, opened 2020-05-14
github actions - build and test

markfirmware:


alexnask:
Thanks for the PR, I will revew it after I finish with #19.  
Note that we plan to add a test suite that includes unit testing functions as well as RPC tests, see #7.

markfirmware:
This is a starting point for rpc testing.

alexnask:
@markfirmware  
Thanks again for the PR, but we decided we will go with regular `zig test` tests that will spawn child processes etc.  

CI builds will be added soon™

PR #22, opened 2020-05-14
Parsing now always results in an AST

Vexu:
Update for https://github.com/ziglang/zig/pull/5336.

### DIFF
--- a/src/analysis.zig
+++ b/src/analysis.zig
@@ -78,7 +78,8 @@ fn collectDocComments(allocator: *std.mem.Allocator, tree: *ast.Tree, doc_commen
 pub fn getFunctionSignature(tree: *ast.Tree, func: *ast.Node.FnProto) []const u8 {
     const start = tree.tokens.at(func.firstToken()).start;
     const end = tree.tokens.at(switch (func.return_type) {
-        .Explicit, .InferErrorSet => |node| node.lastToken()
+        .Explicit, .InferErrorSet => |node| node.lastToken(),
+        .Invalid => |r_paren| r_paren,
     }).end;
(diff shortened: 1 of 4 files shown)


alexnask:
Cool, this looks good, thanks for the style fixes as well.

alexnask:
@Vexu It appears the parsing goes into a presumably infinite loop for some error cases (it consumes more and more memory until I have to kill the process).  

Here is a simple repro:  
```zig
const std = @import("std");

fn local_fn(comptime T: type, comptime V: type) type {
    return T;
}

pub fn main() void {
    // std.hash_map.AutoHashMap
    local_fn(comptime T: type, comptime V: type)
}

```

Launching zls for this document or calling `zig fmt` will block on the handle.tree call (aka `try std.zig.parse(allocator, handle.document.text);`)

Vexu:
Ah nice, I also noticed the memory increasing sometimes but couldn't make a repro.

SuperAuguste:
It seems that the infinite loop issue has been fixed in ziglang/zig#5336. I'll test this pull request out now!

SuperAuguste:
Just tested this, it leaks over 1 gigabyte after a few seconds, as opposed to the normal `0.6 mbs` of memory on the current `zls` master. @alexnask believes this is an issue with your PR, #5336, as `zig fmt` has the same memory leak behavior.

SuperAuguste:
The leak seems to be an edge case. Looking into it a little more.

SuperAuguste:
I cannot recreate the memleak. I'll merge this and take it care of it when it (rarely) reappears later on!

PR #23, opened 2020-05-14
Implement completion for error sets

Vexu:
I also implemented `getDocComments` for all named nodes with them but I have no idea where those are supposed to show up.

### DIFF
--- a/src/analysis.zig
+++ b/src/analysis.zig
@@ -214,6 +214,9 @@ pub fn resolveTypeOfNode(analysis_ctx: *AnalysisContext, node: *ast.Node) ?*ast.
             const field = node.cast(ast.Node.ContainerField).?;
             return resolveTypeOfNode(analysis_ctx, field.type_expr orelse return null);
         },
+        .ErrorSetDecl => {
+            return node;
+        },
         .SuffixOp => {
             const suffix_op = node.cast(ast.Node.SuffixOp).?;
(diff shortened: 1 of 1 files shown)


src/analysis.zig
```diff
@@ -171,6 +182,10 @@ pub fn getChild(tree: *ast.Tree, node: *ast.Node, name: []const u8) ?*ast.Node {
                 const field = child.cast(ast.Node.ContainerField).?;
                 if (std.mem.eql(u8, tree.tokenSlice(field.name_token), name)) return child;
             },
+            .ErrorTag => {
+                const tag = child.cast(ast.Node.ErrorTag).?;
+                if(std.mem.eql(u8, tree.tokenSlice(tag.name_token), name)) return child;
+            },
```
Vexu, line 188: What does `getChild` do? It doesn't seem to matter whether `.ContainerField` and `.ErrorTag` are here.

SuperAuguste: `getChild` gets the child of a node by name. This is useful for completion of values that are children of a variable or field's types, of a function's return value. I don't think `.ErrorTag` in `getChild` is necessary, as far as I know.

SuperAuguste:
This pull request looks great! Fantastic work on the comments, by the way. I think you can remove `.ErrorTag` in `getChild`, but other than that, this pull request is ready to be merged!

PR #32, opened 2020-05-14
Trigger global and field completions correctly in more contexts.

### LINKED ISSUE #28: Field completions do not trigger in all contexts
```zig
const std = @import("std");

pub fn main() void {
    const u13_max = std. // I want std.math.maxInt, why you no complete :( 
}
```

alexnask:
I also ran `zig fmt` by instinct, @SuperAuguste let me know if you want me to reopen with the changes only.  
Closes #28.


SuperAuguste:
This looks good! Merging!

PR #33, opened 2020-05-15
Fixed URI.fromPath making file paths lower case

Sergeeeek:
This fixes std import resolving when std path contains upper case chars.

Also some automatic changes by zig fmt

### DIFF
--- a/src/uri.zig
+++ b/src/uri.zig
@@ -71,7 +71,5 @@ pub fn parse(allocator: *std.mem.Allocator, str: []const u8) ![]u8 {
         i -= 1;
     }
 
-    std.debug.warn("parsed uri {} into {}\n", .{ str, uri });
-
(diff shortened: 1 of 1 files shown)


alexnask:
Hi, thanks for the PR.  
I believe URIs in LSP are lowercase, which is the reason why I made this function lowercase the path.  
I will double check if this is the case.  
If it is, a better solution would be to keep the URIs lowercase like in the current codebase but to also include a filepath field with the correct capitilization in the TextDocument struct.

alexnask:
#35 fixes this and correctly handles windows paths.  
Thanks again for the PR :-)

SuperAuguste:
Thanks for the PR @Sergeeeek, as alexnask has made a PR that fixes this, I'll close this now. Thanks again for the PR.

PR #36, opened 2020-05-15
Mention that zls must be built with zig master

ifreund:
This caused some confusion for me while setting up `zls` with `kak-lsp` and I thought it would be good to mention in the readme.


SuperAuguste:
Great call! Thanks so much for the pull request!

PR #38, opened 2020-05-15
Make style guideline warnings opt-in

pixelherodev:


SuperAuguste:
Looks perfect! I agree that styling should be opt-in, especially as Zig styling is changing every day. Thanks for your contribution! :)

pixelherodev:
Yeah, no problem! I look forward to working with you further; I love this project, and I'll definitely help out a bit more!

Issue #26, opened 2020-05-14
Redefinitions of imports in variables should support field access completions


alexnask:
```zig
const std = @import("std");
const json = std.json;

pub fn main() void {
    json. // <- No completions
}
```

SuperAuguste:
This also kills things like ArrayLists, so this is pretty critical. I'm gonna get working on a fix!

alexnask:
Yes I figured it's probably the same bug but opened separate issues to be safe :)

PR #41, opened 2020-05-16
Rework position context, use AST instead

alexnask:
This seems to work at least for basic things, I'm testing at the moment.

### DIFF
--- a/src/analysis.zig
+++ b/src/analysis.zig
@@ -375,6 +375,7 @@ pub const CompletionContext = struct {
         enum_literal,
         container_field,
         empty,
+        other,
         none,
     };
 
@@ -424,7 +425,7 @@ fn visitNodesAndFindCompletionContext(tree: *ast.Tree, node: *ast.Node, source_i
     return switch (node.id) {
         .Block, .Root => .{ .id = .empty },
         .ContainerField => .{ .id = .container_field },
-        else => CompletionContext.none,
+        else => .{ .id = .other, .node = node },
     };
 }
 
--- a/src/analysis.zig
+++ b/src/analysis.zig
@@ -384,6 +384,8 @@ pub const CompletionContext = struct {
     fn toId(self: CompletionContext, id: Id) CompletionContext {
         return .{ .id = id, .node = self.node };
     }
+
+    const none = CompletionContext{ .id = .none };
 };
 
 fn nodeContainsSourceIndex(tree: *ast.Tree, node: *ast.Node, source_index: usize) bool {
@@ -422,7 +424,7 @@ fn visitNodesAndFindCompletionContext(tree: *ast.Tree, node: *ast.Node, source_i
     return switch (node.id) {
         .Block, .Root => .{ .id = .empty },
         .ContainerField => .{ .id = .container_field },
-        else => .{ .id = .none, .node = node },
+        else => CompletionContext.none,
     };
 }
 
@@ -438,10 +440,18 @@ pub fn completionContext(tree: *ast.Tree, source_index: usize) CompletionContext
     while (tree.tokens.at(token_idx).end <= source_index and token_idx < tree.tokens.len) : (token_idx += 1) {}
     const tok_before_cursor = tree.tokens.at(token_idx);
 
-    const is_dot = tok_before_cursor.id == .Period and tok_before_cursor.end - 1 == source_index;
+    switch (tok_before_cursor.id) {
+        .DocComment,
+        .LineComment,
+        .ContainerDocComment,
+        .MultilineStringLiteralLine,
+        => return CompletionContext.none,
+        else => {},
+    }
 
+    const is_dot = tok_before_cursor.id == .Period and tok_before_cursor.end - 1 == source_index;
     return if (is_dot) switch (completion_context.id) {
-        .container_field => completion_context,
+        .container_field, .none => completion_context,
         .empty => completion_context.toId(.enum_literal),
         else => completion_context.toId(.field_access),
     } else
--- a/src/analysis.zig
+++ b/src/analysis.zig
@@ -323,7 +323,16 @@ pub fn getFieldAccessTypeNode(analysis_ctx: *AnalysisContext, start_node: *ast.N
         .InfixOp => {
             const infix_op = start_node.cast(ast.Node.InfixOp).?;
             if (infix_op.op == .Period) {
-                return resolveTypeOfNode(analysis_ctx, infix_op.lhs);
+                std.debug.assert(infix_op.rhs.id == .Identifier);
+                var rhs_name = nodeToString(analysis_ctx.tree, infix_op.rhs).?;
+                rhs_name = std.mem.dupe(&analysis_ctx.arena.allocator, u8, rhs_name) catch return null;
+
+                if (resolveTypeOfNode(analysis_ctx, infix_op.lhs)) |lhs_type| {
+                    if (getChild(analysis_ctx.tree, lhs_type, rhs_name)) |child| {
+                        return resolveTypeOfNode(analysis_ctx, child);
+                    }
+                }
+                return null;
             } else {
                 return resolveTypeOfNode(analysis_ctx, start_node);
             }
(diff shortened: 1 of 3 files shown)


alexnask:
This also exposes a `.enum_literal` completion context we can use to easily implement this kind of completion (find all enum fields currently visible and it's basically done).

alexnask:
Chaining field accesses are currently broken, fixing.

alexnask:
This looks at least as solid as master to me (probably a bit more since it supports field access on function calls).

alexnask:
Also closes #26

Vexu:
The drawback with this solution is that if the node is not added to the AST then there won't be any completions available. However this is a lot cleaner and with some work we should be able to get the parser to add most if not all nodes to the AST whether or not they are valid.

Example that wont currently provide any completions:
```zig
const std = @import("std");
test "" {
    if (std.)
}
```

alexnask:
@Vexu  
Yes I noticed this too.  
I think the existing implementation also suffers from this (at least there are a tonne of edgecases) so I do think even with the current parser this should probably get merged.  
I will try to compile a list of completions that work with the tokenizer + text parsing solution but don't with this PR.  

What do you think about the code in general?  
You are much more familiar with the parser so I would appreciate feedback, if you have any (especially regarding iterating over tokens to find a trailing dot, I couldn't find a better solution).

src/analysis.zig
```diff
@@ -401,8 +361,89 @@ pub fn nodeToString(tree: *ast.Tree, node: *ast.Node) ?[]const u8 {
...
+
+    // TODO: Can we do this better?
+    // Text like `someExpr().` will not get parsed to InfixOp.
+    // As a workaround, we check if the position corresponds exactly to a
+    // period token and adjust the context.
+    var token_idx: ast.TokenIndex = 0;
+    while (tree.tokens.at(token_idx).end <= source_index and token_idx < tree.tokens.len) : (token_idx += 1) {}
+    const tok_before_cursor = tree.tokens.at(token_idx);
+
+    const is_dot = tok_before_cursor.id == .Period and tok_before_cursor.end - 1 == source_index;
```
Vexu, line 441: You should also check that `tok_before_cursor.end - 1 != .Period`.
Currently this provides completions `foo[..<cursor here>)]`.

src/main.zig
```diff
@@ -264,19 +264,23 @@ fn completeGlobal(id: i64, handle: DocumentStore.Handle, config: Config) !void {
     });
 }
 
-fn completeFieldAccess(id: i64, handle: *DocumentStore.Handle, position: types.Position, line_start_idx: usize, config: Config) !void {
+// .field_access => try completeFieldAccess(id, handle, completion_context, config),
```
alexnask, line 267: Woops, missed this comment, will remove

src/analysis.zig
```diff
@@ -401,8 +361,89 @@ pub fn nodeToString(tree: *ast.Tree, node: *ast.Node) ?[]const u8 {
...
+            }
+        },
+        .BuiltinCall => return .{ .id = .builtin, .node = node },
+        .EnumLiteral => return .{ .id = .enum_literal, .node = node },
+        else => {},
+    }
+
+    var node_idx: usize = 0;
+    while (node.iterate(node_idx)) |child| : (node_idx += 1) {
+        if (!nodeContainsSourceIndex(tree, child, source_index)) continue;
```
Vexu, line 418: This won't stop us from iterating over all the children of this node.
If we don't want to iterate over all the child nodes and collect analytics at the same time (which I think we should do), we need to make our own iterator which will skip all the child nodes if the node doesn't contain our source index.

Vexu: Collecting the analytics at the same time as we are iterating should also make collecting all in scope variables quite easy.

alexnask: Yes, I agree that if we are going to be visiting the nodes it seems like a waste not to collect symbol tables and other information.  

Can you elaborate on the custom iterator?  
This loop will iterate over the children until it finds the child that contains the cursor, it will not iterate over the rest of the children.

Vexu:
```zig
test "a" {
    // We will iterate all the nodes in here even though there
    // is no way any of them contain our index since the test 
    // declaration didn't contain it.
}
test "b" {
    std.<cursor here>
}
```

src/analysis.zig
```diff
@@ -401,8 +361,89 @@ pub fn nodeToString(tree: *ast.Tree, node: *ast.Node) ?[]const u8 {
...
+    const completion_context = visitNodesAndFindCompletionContext(tree, &tree.root_node.base, source_index);
+
+    // TODO: Can we do this better?
+    // Text like `someExpr().` will not get parsed to InfixOp.
+    // As a workaround, we check if the position corresponds exactly to a
+    // period token and adjust the context.
+    var token_idx: ast.TokenIndex = 0;
+    while (tree.tokens.at(token_idx).end <= source_index and token_idx < tree.tokens.len) : (token_idx += 1) {}
+    const tok_before_cursor = tree.tokens.at(token_idx);
+
```
Vexu, line 440:
```zig
switch (tok_before_cursor.id) {
    .LineComment, .DocComment, .ContainerDocComment => return .{ .id = .empty },
    else => ...,
}
```

src/analysis.zig
alexnask: How is this the case?  
We enter `visitNodesAndFindCompletionContext` with `node == &tree.root_node.base`, `iterate(0)` will yield the first `TestDecl`, which has `firstToken() == test` and `lastToken() == body_node.lastToken() == }`, so `nodeContainsSourceIndex` will return false and the loop will continue and yield the second `TestDecl`.  
We only recursively visit when `nodeContainsSourceIndex` is true.

Vexu: Ah, my bad. I was thinking of how the `Node.dump` function worked but I just noticed it calls `dump` on all the child nodes instead of `iterate` doing so.

alexnask: Cool, I was wondering if I missed some `iterate` that does a deep iteration :P

src/analysis.zig
alexnask: This turned out to be a separate issue, see https://github.com/zigtools/zls/pull/41/commits/687addb49e925d2a1dcdd7e28086c37f709df0e2#diff-d2841187941f93280c97ecaafd02aeb4R423-R427

src/analysis.zig
alexnask: This turned out to be a separate issue, see https://github.com/zigtools/zls/pull/41/commits/687addb49e925d2a1dcdd7e28086c37f709df0e2#diff-d2841187941f93280c97ecaafd02aeb4R423-R427

alexnask:
Update for this PR:  
Currently this works well whenever we have access to some erro-free parent node when editing.  
However, incomplete statements like

Vexu:
A reasonable solution that would preserve the meaning of optional nodes would be to introduce a new special `Invalid` node that can only exist in AST's with errors. Then handling it in `zig fmt` and stage2 would be as simple as `.Invalid => unreachable`.

I already have most of the needed parts ready, so implementing this would be a matter of updating some call sites.

Vexu:
Alright here it is in all its glory https://github.com/Vexu/zig/commit/5a0cab187f4d6a064d798974e3fa30010891c979 +68 -82 and Zig will successfully produce an AST from literally any stream of bytes you throw at it. It does however break `findNextStmt`, fixing which will require a bit more work.

With this patch applied I'm getting completion nearly everywhere.

alexnask:
@Vexu Amazing! I expected it to require much more changes lol

Vexu:
It is a bit of a nuclear option, but with a limit on errors and `findNextStmt` fixed it should be viable. @andrewrk any opinions on this?

Vexu:
After experimenting some with that branch I think the best way to get the context is to run the Zig tokenizer on all the lines until `source_index` and then iterating backwards. Doing so would avoid all the issues with the AST but would still be much easier to extend than the current method.

PR #47, opened 2020-05-17
Cleanups and rewrite of reading

daurnimator:


src/document_store.zig
```diff
@@ -67,8 +62,8 @@ fn newDocument(self: *DocumentStore, uri: []const u8, text: []u8) !*Handle {
         },
     };
     try self.checkSanity(&handle);
-    try self.handles.putNoClobber(uri, handle);
-    return &(self.handles.get(uri) orelse unreachable).value;
+    const kv = try self.handles.getOrPutValue(uri, handle);
```
alexnask, line 65: Nice, I somehow missed this function in HashMap...  
I think the errdefer is still needed here though, this function takes ownership of `uri` and `text`.

src/document_store.zig
```diff
@@ -84,7 +79,7 @@ pub fn openDocument(self: *DocumentStore, uri: []const u8, text: []const u8) !*H
     const duped_uri = try std.mem.dupe(self.allocator, u8, uri);
     errdefer self.allocator.free(duped_uri);
 
-    return self.newDocument(duped_uri, duped_text);
+    return try self.newDocument(duped_uri, duped_text);
```
alexnask, line 82: Is this a necessary change?  
I prefer the `return <maybe_erroring_expression>` instead of `return try`, is there some reasoning to choose the latter?

src/document_store.zig
alexnask: Ah, I guess it is confusing since `duped_text` and `duped_uri` are errdefer free'd in `openDocument`.  
This is a bug, they should be handled by `newDocument`.

src/document_store.zig
daurnimator: yes. it adds a new line to the error stack trace.

alexnask: Ah good to know, thanks.  
Let's go ahead with this style instead then.

src/document_store.zig
daurnimator: it's very odd to take ownership of arguments if an error is thrown.

alexnask: Yes, I see your point.  
I think this way is clearer for the caller so it should be preferred.  
Thanks for your excellent PR and your input :)

SuperAuguste:
This is amazing! Thanks so much daurnimator!

Issue #51, opened 2020-05-18
Hard crash on incorrect import code


wilsonk:
Hello all,

```zig
const std = @import("std");
const panic = std.debug.panic;
.
.
.
panic.
```
The above code crashes the server (server has crashed 5 times in the last 3 seconds)  using newest vscode-1.45.1 and zls-cb4354e494bfadd2e74d5e3e16955d214fc638c1. I realize this is malformed code but I just ran into it after forgetting that I had already aliased panic.  This error happens at any scope.

I had to restart vscode to rectify this so I thought I would mention it here, as it could be a pain in larger files if you forget what all you have aliased.

wilsonk:
Seems to be fixed now. Thanks

Issue #52, opened 2020-05-18
Add snippet completion configuration options


wilsonk:
```zig
const std = @import("std");
const panic = std.debug.pan    <<-- TAB
```
After you TAB to complete that import in the newest vscode, zls completes and fills in the parameter signatures for the panic function. However, if you just wanted to complete to 'std.debug.panic;' by trying to hit ';' after the TAB (which is the default in some other languages), then the parameter signatures don't self delete properly.

Not sure if that is the correct behavior here but thought I would mention it.

alexnask:
@wilsonk Ah, I was not aware of that behavior.  
I will try to see how we can support this, it sounds useful.

wilsonk:
Title changed as per suggestion from @alexnask after discussing this. I was using a third party completion tool called Tabnine that was giving me different options for TAB completions. This situation doesn't seem to be solvable the way I mentioned above, but Alex may have a different solution.

alexnask:
We could add options like inserting a tab stop before the parentheses.  
Another option could be choice completion of enum literals for enum arguments.  
We can do so some interesting stuff with [snippet completions](https://github.com/microsoft/language-server-protocol/blob/master/snippetSyntax.md).

wilsonk:
Options for different types of behavior like this are usually appreciated by the end users...in my experience, at least. It would be a nice addition when you get the chance. Thanks

Techatrix:
If I understand the issue description correctly then this has become configurable through `enable_argument_placeholders` option.
Solved almost 4 years later. 🫡

PR #53, opened 2020-05-18
Fixed crash when field access completing on a local variable with a type from another import

alexnask:
This code was assuming the original tree remains valid, we need to switch the scope_decls when hitting an import.


alexnask:
Closes #51, #52.

SuperAuguste:
Great catch!

PR #57, opened 2020-05-18
Completion support when a function with multiple arguments is completed

FireFox317:
When completing a function foo: `fo<tab>` -> `foo(arg1: u8, arg2: u8)` and filling in arg1 with `std.` no completion would show up, because only EOF was considered as the end.


SuperAuguste:
Thanks so much! Merging!

PR #63, opened 2020-05-20
Update README to specify --recurse-submodules

iansimonson:
ZLS now uses known-folders via submodules and no longer builds unless that submodule is initialized.

Tiny patch to make the installation instructions copy/paste again


SuperAuguste:
Great call!

PR #72, opened 2020-05-21
Check that the goto definition cursor is within bounds

alexnask:
This should fix the random crashes in VSCode.


SuperAuguste:
Fantastic! Thanks for the hard work!

PR #73, opened 2020-05-22
Add goto definition support for import strings

### LINKED ISSUE #68: Clicking the string in a @import should open the file.

Vexu:
Currently this highlights individual identifiers in the import string but I don't think there is a way to avoid that since Microsoft's C/C++ extension works the same way.

Closes #68.


SuperAuguste:
Fantastic! Thanks Vexu!

PR #75, opened 2020-05-23
Fix @This() completions, do half the work in global completions

alexnask:
Also removes onContainer.
This has currently broken alias function completions (e.g. std.ArrayList, you need to do std.ArrayList.ArrayList and it will complete as expected)  

We were calling declsFromIndex twice unnecessarily in global completions.  

EDIT: This was simply wrong ordering on arguments, fixed.


SuperAuguste:
This seems to work for me! Merging!

PR #77, opened 2020-05-23
Now compiles on the new parser branch

alexnask:
I need to merge master in this.  
The only change that affects us is the fact that param decl is no longer a node.  

The way to go is to probably make our own Decl type and return this in the relevant places, this way we also get to attach doc comments etc while doing `declsFromIndex`.


SuperAuguste:
Cool! I'll merge this once the parser branch is merged with master!

PR #83, opened 2020-05-25
Support custom package imports

alexnask:
Done

### DIFF
--- a/src/special/build_runner.zig
+++ b/src/special/build_runner.zig
@@ -1,0 +1,89 @@
+const root = @import("build.zig");
+const std = @import("std");
+const io = std.io;
+const fmt = std.fmt;
+const Builder = std.build.Builder;
+const Pkg = std.build.Pkg;
+const LibExeObjStep = std.build.LibExeObjStep;
+const ArrayList = std.ArrayList;
+///! This is a modified build runner to extract information out of build.zig
+///! Modified from std.special.build_runner
+
+// We use a custom Allocator to intercept the creation of steps
+const InterceptAllocator = struct {
+    base_allocator: *std.mem.Allocator,
+    allocator: std.mem.Allocator,
+    steps: std.ArrayListUnmanaged(*LibExeObjStep),
+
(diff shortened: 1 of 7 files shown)


src/special/build_runner.zig
```diff
@@ -0,0 +1,89 @@
...
+const fmt = std.fmt;
+const Builder = std.build.Builder;
+const Pkg = std.build.Pkg;
+const LibExeObjStep = std.build.LibExeObjStep;
+const ArrayList = std.ArrayList;
+///! This is a modified build runner to extract information out of build.zig
+///! Modified from std.special.build_runner
+
+// We use a custom Allocator to intercept the creation of steps
+const InterceptAllocator = struct {
```
alexnask, line 13: Getting a list of the `LibExeObjStep`s is a little more involved than I would like but there is no way to differentiate between steps.  
The only way would be to check the make functions afainst the step function we want but they are not `pub`.

SuperAuguste: I don't think differentiating between steps *necessarily* matters as we're just looking to resolve library names to paths, and this works flawlessly for doing so!

andrewrk: We can definitely change std/build.zig code to make this use case easier! No need for such a hack when upstream is willing to improve the API :)

andrewrk:
> Find `zig` in PATH

Zig passes the path to itself as one of the args to the build runner, so that you don't have to do this - you will get the correct path:

https://github.com/ziglang/zig/blob/3052fd84c812280334245513d1edc631ae5e4ae2/lib/std/special/build_runner.zig#L27-L30

alexnask:
@andrewrk  
We have decided that the best way to proceed is probably by shipping just this file and using `zig run build_runner.zig` in directories where builds live.  
This way we do not have to ship the whole standard library.  
Either way, I would still have to look for `zig` in `PATH`, this is not so much about getting its directory but just making sure we can run zig as a subprocess.

alexnask:
This now depends on https://github.com/ziglang/zig/pull/5431

PR #88, opened 2020-05-26
Reworked position context to use tokenizer, works in more contexts

alexnask:
Example of a completion that is now enabled:  
image

Here we see multiple functions chained with working completions, goto definition and hover support.  
I will implement the `TODO`s in `getFieldAccessTypeNode` as well as start the testsuite with tests of this function.


alexnask:
I will also add an `error` context and adapt @Vexu's PR to the latest codebase now that we have a saner position context system :)

PR #102, opened 2020-06-05
Refactor workflow commands

joachimschmidt557:
I think this works, but I'm not 100% sure.


pixelherodev:
The only behavioral difference that should be present would be the lack of `exit 0` at the end, but the default return code is zero if the end of the script is reached, so I'm fairly certain this is good.

alexnask:
Yes, this is good, thanks for the patch,

PR #103, opened 2020-06-06
Correctly skip first comma in bound function snippets

Vexu:


SuperAuguste:
Good catch, and great fix! Merging!

PR #106, opened 2020-06-07
Fix Sublime Text 3 install instructions

squeek502:
The name of the Zig package is "Zig Language": https://github.com/wbond/package_control_channel/blob/1d8871756d78e7088a685392b8a5ae4b67a498e6/repository/z.json#L156


SuperAuguste:
Thanks!

PR #107, opened 2020-06-08
Added usage instructions for emacs

kristianhasselknippe:


alexnask:
Thanks for the PR, could you add `(require 'lsp)` to the code snippet (although I guess this must be obvious to emacs users)?

alexnask:
Thanks

PR #121, opened 2020-06-17
Prototype semantic tokens implementation with actual semantic analysis

alexnask:


alexnask:
This may blow your stack up, beware.  
I need to rewrite it iteratively.

Issue #120, opened 2020-06-17
Zig "std" implementations marked as zls(BadStyle)


axzn:
This is reproducible at least in **VSCode**, with following steps to reproduce the Issue:
1. Create a `.zig` file with following contents:
```zig
const std = @import("std");
pub fn main() void {
    std.math.cos(0);
}
```
2. `Ctrl-Click` the `cos`.
3. **zls** now scans the `lib/zig/std/math/cos.zig` standard library file, and marks the line `fn cos_(comptime T: type, x_: T) T` with `Functions should be camelCase zls(BadStyle)`

Not sure if I should report this as an Issue here, but since **zls** flags *"official"* **zig** implementation code as bad style, seems pretty reasonable. Do you think this should be an Issue for [ziglang/zig](https://github.com/ziglang/zig) instead?

SuperAuguste:
Hey @axzn! Thanks for opening this issue! The Zig style guide is not official, and ZLS's warnings are just helpful indicators. ZLS has a naive implementation that simply checks for capitalization and underscores, so that is the cause of this issue. There's no real fix to this except adding a condition that considers underscores at the end of function names valid, even though that isn't in the style guide.

alexnask:
Closed by #125

PR #125, opened 2020-06-24
fix ZLS flagging std "cos_" as not camel case (see issue #120)

axzn:
This fixes the special case where ZLS flags an std implementation on cos as zls(BadStyle) (please see issue #120).


alexnask:
Thanks :)

PR #132, opened 2020-07-04
Add semantic tokens to config build step

truemedian:


alexnask:
Thanks.  
I will also add some information to the readme about semantic tokens, in what editors they work and what token types we support soon.

PR #147, opened 2020-08-13
Update logging statements for zig std.log breaking change

heidezomp:
I implemented ziglang/zig#6046 with breaking changes to std.log and then noticed that my zls wouldn't build anymore... so I made it build again :smile: 

Marked this PR as draft since the upstream zig PR isn't merged yet.

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -1,4 +1,5 @@
 const std = @import("std");
+const main_log = std.log.scoped(.main);
 const build_options = @import("build_options");
 
 const Config = @import("config.zig");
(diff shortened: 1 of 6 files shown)


src/main.zig
```diff
@@ -1,4 +1,5 @@
 const std = @import("std");
+const main_log = std.log.scoped(.main);
```
heidezomp, line 2: I don't really like that I had to call this one main_log instead of just log like in the other files, since otherwise it would clash with the custom log implementation. I argued for renaming the log implementation to "logger" to prevent this name collision here https://github.com/ziglang/zig/pull/6039#issue-466729548:
> The intended usage pattern of `const log = std.log.scoped(.main); log.err("oops", .{});` can be used everywhere _except_ in the root source file, if that file overrides the log implementation (since it would result in a naming conflict). Is this a problem? Maybe the log implementation could be renamed to "logger" instead?

Do you have any ideas to improve this, or do you think it's fine the way it is?

alexnask:
Hi, thanks for the contributioin.  
I just updated this before checking for new issues and PRs :P  
I went with logger for the main scoped log for now :)

heidezomp:
That's fine, it works again, that's the most important part :smile:

PR #166, opened 2020-10-02
Added path of zls exec to Installation section

Cattes:
For people new to zig and language servers, the hint where the .exe will be located is helpful. It is the path that has to be added to the editors.


alexnask:
Thanks for the PR, the default zig prefix can be confusing for sure

PR #173, opened 2020-10-18
Increase eval branch quota

data-man:
Because I got
> mem.zig:871:5: error: evaluation exceeded 3500 backwards branches

with updated Zig.


leecannon:
Just to note, for zig version `0.6.0+6ddb05d99` the minimum required branch quota is 3,679

PR #174, opened 2020-10-19
Remove -Dallocation_info from CI build

truemedian:
With the removal of the debug allocator, this was also removed from the build.zig, and causes the CI to fail.


SuperAuguste:
Good call, thanks! :)

PR #177, opened 2020-10-26
Add Nix derivation

arqv:
Added Nix derivation for easy installation/usage in Nix-based environments.
Uses an specific Zig version (currently `0.6.0+91a1c20e7`), but can be easily changed if needed.
Sadly in this case the `zig build config` command can't be used, so the derivation also generates a `zls.json` file using the defaults described in the `README.md` file specifying the correct Zig paths used during building.


alexnask:
Thanks!

PR #183, opened 2020-11-02
std.fmt.trim is now deprectated

leecannon:
`std.fmt.trim` was replaced with a `@compileError` in commit https://github.com/ziglang/zig/commit/d530e7f9c7e19b2c9d9117c3120cd75855f4023b

Now it looks like there is no easy string trim function...

### DIFF
--- a/src/analysis.zig
+++ b/src/analysis.zig
@@ -51,7 +51,7 @@ pub fn collectDocComments(
         switch (tree.token_ids[curr_line_tok]) {
             .LineComment => continue,
             .DocComment, .ContainerDocComment => {
-                try lines.append(std.fmt.trim(tree.tokenSlice(curr_line_tok)[3..]));
+                try lines.append(std.mem.trim(u8, tree.tokenSlice(curr_line_tok)[3..], &[_]u8{ ' ', '\t', '\n', '\r' }));
             },
             else => break,
         }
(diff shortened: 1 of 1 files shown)


alexnask:
I thought some trimWhitespace function was introduced instead, I will check it out

src/analysis.zig
```diff
@@ -51,7 +51,7 @@ pub fn collectDocComments(
         switch (tree.token_ids[curr_line_tok]) {
             .LineComment => continue,
             .DocComment, .ContainerDocComment => {
-                try lines.append(std.fmt.trim(tree.tokenSlice(curr_line_tok)[3..]));
+                try lines.append(std.mem.trim(u8, tree.tokenSlice(curr_line_tok)[3..], &[_]u8{ ' ', '\t', '\n', '\r' }));
```
ifreund, line 54: The proper replacement is using `mem.trim` with `ascii.spaces`:
```suggestion
                try lines.append(std.mem.trim(u8, tree.tokenSlice(curr_line_tok)[3..], &std.ascii.spaces));
```

leecannon: I probably should have looked around a bit more instead of hardcoding the characters like `std.fmt.trim` did.

I've update this pull request to use `ascii.spaces`

leecannon:
`std.ascii.spaces` is the same set of characters as formerly used by `std.fmt.trim` except with two control characters included.
Due to this the behaviour of this function is not exactly the same as before, but is probably more correct now.

src/analysis.zig
ifreund: well depending on when you looked it may not have been there yet, I just added it: https://github.com/ziglang/zig/pull/6931

alexnask:
Thanks again

PR #186, opened 2020-11-04
analysis.collectDocComments: Replace `std.fmt.trim` with `std.mem.trim`

fdschonborn:
This commit replaces the usage of `std.fmt.trim`, which was deprectated in [this commit][1], with `std.mem.trim` and `std.ascii.spaces`.

[1]: https://github.com/ziglang/zig/commit/50ba0182235be67f79a1d088f279d69efd06b5d2


alexnask:
Thanks for the PR, it was already fixed in #183 though :)

fdschonborn:
Oh, nuts. :grin:

PR #198, opened 2020-11-24
build config incorrectly pulls folder named zig if it appears in PATH variable before zig executable

aniljava:
pull request adds a isDirectory check on the resolved path.


alexnask:
Thanks!

PR #215, opened 2021-01-10
Fix inStream/outStream removal

vesim987:
This also requires https://github.com/ziglibs/zinput/pull/2

PR #217, opened 2021-01-11
fix inStream deprecation

truemedian:
This was missed in #215, and it causes no visible error in any output.


SuperAuguste:
Thanks! 😄

PR #221, opened 2021-01-16
Semantic Highlight changes

### LINKED ISSUE #220: Semantic Highlight: use standard semantic tokens/modifiers as much as possible
I want to propose some changes to the semantic highlight to use as much as possible the [standard semantic tokens/modifiers](https://code.visualstudio.com/api/language-extensions/semantic-highlight-guide), for better support in themes that only set up the standard ones.

## Change 1

For symbol declarations the modifier should be `declaration` instead of `definition`.

The standard semantic modifiers supported have both `declaration` and `definition` modifiers, with some difference:

declaration: For declarations of symbols.
definition: For definitions of symbols, e.g. in header files.

### Prior work
All other semantic highlighters I know also uses `declaration` for symbol declaration (rust-analyzer, pylance, Language Support for Java and TypeScript)

## Change 2

Change the `namespace` modifier to be a token. This change would improve the quality of zig code highlighted by themes that only define colors for the standard tokens/modifiers.

## Change 3

Change the name of `tagField` to `enumMember`. This change would improve the quality of zig code highlighted by themes that only define colors for the standard tokens/modifiers.

GrayJack:
Rename token `tagField` to `enumMember` and change usage of `definition` to `declaration` modifier in symbol declarations.

This should close #220 

And as commented in the issue, the `namespace` modifier didn't change.

### DIFF
--- a/src/semantic_tokens.zig
+++ b/src/semantic_tokens.zig
@@ -28,6 +28,7 @@ pub const TokenModifiers = packed struct {
     @"enum": bool = false,
     @"union": bool = false,
     @"opaque": bool = false,
+    declaration: bool = false,
     definition: bool = false,
     @"async": bool = false,
     documentation: bool = false,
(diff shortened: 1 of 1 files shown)


GrayJack:
I did not understood the error message that was shown in the CI 😞

alexnask:
Thanks, this looks good.  
The tests currently check for the expected server capabilities without nay flexibility, I will fix them :)

PR #233, opened 2021-02-13
Update inline functions to use the new syntax

faerryn:
inline fn() void -> fn () callconv(.Inline) void


faerryn:
a886ecddd98b5fd1134b9007a1d62e598a69b549 does this, closing

PR #236, opened 2021-02-21
README: describe how to quickly install ZLS on MacOS/Linux

slimsag:
I have had these commands in my personal docs for a while now because I find them useful when setting up the Zig language server, especially because remembering the right tar options and that `xz` needs to be installed can be difficult to recall. I thought maybe they'd make a good addition to the README.


SuperAuguste:
Looks good, thanks!

PR #243, opened 2021-03-08
added mage emoji

woofymax:
In the build.zig, i added a mage emoji where it said "(insert mage emoji here)" 
i understand if this is not accepted :)


SuperAuguste:
Hey @woofymax, thanks for submitting this PR! The reason I originally replaced the actual emoji with "(insert mage emoji here)" was because Windows didn't render the mage properly. This sadly still seems to be the case so I'm not sure this would be a good idea. I'm closing this PR unless you have any particularly strong feelings about adding back the cool mage emoji 🧙‍♂️.

PR #238, opened 2021-02-26
Update to Zig std's new AST format

Luukdegram:
Started updating zls to use the new AST that was introduced in https://github.com/ziglang/zig/pull/7920
I hope to get this finished quickly so people can update to the latest master again.
Perhaps we should consider creating a new binary release after this as well.

Will fix : 
- fixes #237 
- fixes #152
- fixes #157
- fixes #153
- fixes #150
- fixes #242
- fixes #244

This also introduces a new option to `zls.json`, allowing the server to skip finding for references in std. 
This means that when searching for references of a user's function, it will skip searching through std to find a reference to it, allowing for a big speedup in lookups. This is only for references. Go to definition and renaming will continue to work as is.
The option can be enabled by setting the following field to `true`:
`skip_std_references`

TODO's:
- [x] Implement errors (completions/docs/etc)
- [x] Semantic tokens (currently commented out)
- [x] makeScopeInternal
- [x] resolveTypeOfNodeInternal
- [x] usingnamespace
- [x] Implement `test ""{}` support
- [x] Make it actually compile
- [x] Ensure no more `@TODO:` inserted by me are left
- [x] Goto definition on imports.
- [x] Completion for generic types that are imported.
- [x] Completion for `Try`
- [x] Incorrect completions. i.e. https://github.com/vrischmann/zig-sqlite/blob/master/sqlite.zig#L302
- [x] Slice/Array payload completion
- [x] Unwrapping optional in `if` causes crash

PR #245, opened 2021-03-10
Add missing \n in error message output to stderr

rhysd:
Hi, today I built zls and noticed that newlines are missing in debug logs.

For example,

```
> zig-cache/bin/zls --debug-log
Enabled debug loggingContent-Length: 184
```

Newline should be inserted between `logging` and `Content`.

This PR fixes error messages output to stderr.


rhysd:
Note that CI fails because #238 is not merged yet.

SuperAuguste:
Thanks!

PR #251, opened 2021-03-15
Fix falsy value for filterText and insertText

rhysd:
`filterText` and `insertText` fields of `CompletionItem` can only have `""` or `undefined` as falsy value.

https://github.com/microsoft/language-server-protocol/issues/1222

This PR fixes falsy values for the fields from `null` to `""`.


SuperAuguste:
Thanks! I think this issue is actually as old as zls itself as I distinctly remember panicking about falsy values at some point when initially adding completion, so thank you so much for fixing it! :)

PR #253, opened 2021-03-18
Doom Emacs Changes

Ominitay:
Changed the instruction in README.md to match with other editor instructions, and added the same instructions to the build config instructions.


alexnask:
ty

PR #255, opened 2021-03-21
Emacs: update setup instructions

bnjmnt4n:
`lsp-mode` now has built-in configuration for zls, so the only configuration that might be required is setting the location of the zls
executable (or placing it in the PATH).

`doom-emacs` has not updated to use the latest version of `lsp-mode` yet, so its configuration cannot be simplified. However, I've updated the snippet for configuration to avoid loading `lsp-mode` on startup, which can cause significant slowdown to Emacs startup, and instead defer loading it until it is required.


alexnask:
Cool!  
Thanks for this

PR #259, opened 2021-03-26
Added YouCompleteMe Instructions

KevinGConyers:
Getting zls to work with YouCompleteMe was enough of a hassle that I wanted to share it here. Hope this will help someone.

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,6 +161,24 @@ Install the `zls-vscode` extension from [here](https://github.com/zigtools/zls-v
 }
 ```
 
+#### YouCompleteMe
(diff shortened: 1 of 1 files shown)


alexnask:
Thanks :)

ernesernesto:
@KevinGConyers if it's not a hassle can you share your ycm_extra_conf here? 
I tried setting up zig completion with ycm + zls and still have no luck. Ycm was built with clang-completer and added ycm_language_server in vimrc but zig files still has "Native filetype completion not supported for current file, cannot force recompilation"

Thanks!

PR #261, opened 2021-03-27
Add option to truncate long completions

InterplanetaryEngineer:
The detail entries for big structs such as `std.zig.CrossTarget` were bricking the preview window in Sublime Text.

I needed to increase the EvalBranchQuota of the json parser to add this option even though it is just a single integer. Maybe we should evaluate the build time impact of adding config options and consider alternatives (TOML?)


alexnask:
Looks good
Maybe we could use [zzz](https://github.com/gruebite/zzz), I quite like the syntax and it looks like the codebase should work at comptime, I will evaluate the effect on the compile times/branch quota.

PR #264, opened 2021-03-28
Improve textDocument/references performance

InterplanetaryEngineer:
Filtering out documents in `symbolReferences` that do not import the file defining the symbol led to an order of magnitude speedup in the projects that I tested.


alexnask:
Nice work, I will review this later.  
Personally I also want to add `semanticTokens/full/delta` and `semanticTokens/range` at some point for performance reasons.  
I'm sure there are many perfomance improvements to be had, I never tuned the code but was rather adding features nonstop.

Issue #228, opened 2021-01-30
Neovim lspconfig Integration Flakes


hspak:
Hello. I'm stuck trying to debug my neovim+lspconfig+zls config.

Zls works for me, but I see this error pop up when opening new files and saving.
```
LSP[zls]: Error NO_RESULT_CALLBACK_FOUND: {
  id = 0,
  jsonrpc = "2.0",
  result = {
    data = { 0, 0, 31, 8, 0, 1, 0, 43, 8, 0, 1, 0, 75, 8, 0, 1, 0, 78, 8, 0, 1, 0, 44, 8, 0, 1, 0, 5, 7, 0, 0, 6, 3, 0, 33, 0, 4, 1, 11,
 0, 0, 2, 7, 12, 0, 0, 8, 12, 9, 0, 1, 0, 5, 7, 0, 0, 6, 2, 0, 33, 0, 3, 1, 11, 0, 0, 2, 3, 0, 1, 0, 4, 2, 0, 1, 2, 0, 3, 7, 0, 0, 4, 2,
 7, 0, 0, 3, 14, 0, 256, 0, 15, 8, 7, 0, 0, 9, 11, 1, 32, 0, 13, 5, 0, 0, 0, 7, 8, 7, 0, 0, 9, 10, 1, 32, 0, 12, 4, 0, 0, 0, 6, 4, 0, 0,
 1, 4, 6, 7, 0, 0, 7, 6, 7, 0, 1, 8, 17, 4, 0, 0, 19, 10, 1, 0, 1, 8, 4, 4, 0, 0, 6, 8, 0, 2, 0, 9, 1, 11, 0, 0, 2, 8, 0, 2, 0, 9, 4, 6,
 0, 2, 8, 3, 7, 0, 0, 4, 5, 7, 0, 0, 6, 5, 2, 32, 0, 6, 1, 11, 0, 0, 2, 10, 1, 0, 1, 8, 3, 7, 0, 0, 4, 5, 7, 0, 0, 6, 6, 0, 34, 0, 7, 1,
 11, 0, 0, 2, 2, 0, 1, 0, 3, 6, 0, 0, 0, 7, 1, 11, 0, 0, 1, 4, 0, 2, 0, 6, 5, 2, 0, 0, 7, 5, 6, 0, 2, 8, 5, 7, 0, 0, 6, 4, 0, 34, 0, 5,
[...]

PR #269, opened 2021-03-30
Add check for semantic tokens client capability

austinclem1:
I was also having issue #228 with Neovim lsp, disabling semantic tokens in the zls config fixed it but this should now respect that the client doesn't have that capability even if you have it enabled for zls


alexnask:
Thank you for the PR :)

PR #270, opened 2021-03-30
Fix some crashes & find all @imports

InterplanetaryEngineer:
If there are parse errors, an AST can contain uninitialized nodes. Walking the tree in this case can lead to horribly nasty crashes.

### DIFF
--- a/src/analysis.zig
+++ b/src/analysis.zig
@@ -222,6 +222,8 @@ pub fn isPascalCase(name: []const u8) bool {
 pub fn getDeclNameToken(tree: ast.Tree, node: ast.Node.Index) ?ast.TokenIndex {
     const tags = tree.nodes.items(.tag);
     const main_token = tree.nodes.items(.main_token)[node];
+    if (tree.errors.len > 0)
+        return null;
     return switch (tags[node]) {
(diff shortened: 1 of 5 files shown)


alexnask:
The imports code is nice.  
However, I don't think completely giving up on methods like the semantic highlighting or document dymbols when there is a parsing error is correct.  
The crashes happen (from what I can tell) because we use std.zig.ast.Tree methods that will assume the tree is well formed and include `unreachables`.  
In ast.zig we have a collection of functions taken from std that correct this behavior, this set of functions should be increased.  
I will go through and convert all the Tree calls to safe versions today, then we can reassess wether we actually need to turn off some portions of zls when we get a parsing error.

InterplanetaryEngineer:
Nice!

I think Rename Symbol needs to be reliable to be useful (it isn't yet, I'll try to construct an import tree to be able to modify all symbol references, not just ones in open documents), and it would be better to leave a symbol unchanged than rename the wrong one.

alexnask:
Yeah, the recent ast changes really made zls unstable, hopefully we will be able to make it rock solid again in a couple of days :)

Issue #229, opened 2021-02-04
Prebuilt binaries


faraazahmad:
I was trying to install zls, but I found that there aren't any prebuilt binaries for it (yet?). Is there a plan to provide prebuilt binary releases? because sometimes people just want to download and use it out of the box

haze:
In what scenario are you using `zls` without `zig`?

faraazahmad:
People just want to download and use. For example, rust-analyzer. I can just download and use it. I don't want to deal with building it and then configuring. In the end, it only helps zig and zls adoption

InterplanetaryEngineer:
To enable this, the config wizard would need to be extracted form the build system and possibly made a subcommand of the executable.

faraazahmad:
Sounds like a plan

alexnask:
I will set up artifact uploads from CI

Issue #194, opened 2020-11-16
zig build config panic due to invalid object name on windows


bbodi:
Hi,

- zig version: 0.7.0+6e2e747b0
- zls version:  bf4f653bf985dba3e3849ba133848c37721d0445

Running ``zig build config`` run into panic on my machine due to a weird path name while it was looking for zig.exe.

Output:
```
Welcome to the ZLS configuration wizard! (insert mage emoji here)
Looking for 'zig' in PATH...
reached unreachable code
C:\Users\sharp\dev\zig\lib\zig\std\os\windows.zig:113:37: 0x7ff73e3afc59 in std.os.windows.OpenFile (build.obj)
            .OBJECT_NAME_INVALID => unreachable,
                                    ^
C:\Users\sharp\dev\zig\lib\zig\std\os.zig:4205:31: 0x7ff73e3d4bc0 in std.os.realpathW (build.obj)
                              ^
C:\Users\sharp\dev\zig\lib\zig\std\os.zig:4153:25: 0x7ff73e3beecd in std.os.realpath (build.obj)
        return realpathW(pathname_w.span(), out_buffer);
C:\Users\sharp\dev\zls\build.zig:39:75: 0x7ff73e3b9226 in .@build.config (build.obj)
            zig_exe_path = try std.mem.dupe(allocator, u8, std.os.realpath(full_path, &buf) catch continue);
                                                                          ^
        try self.makeFn(self);
                       ^
C:\Users\sharp\dev\zig\lib\zig\std\build.zig:435:19: 0x7ff73e3a94c8 in std.build.Builder::std.build.Builder.makeOneStep (build.obj)
        try s.make();
                  ^
C:\Users\sharp\dev\zig\lib\zig\std\build.zig:390:33: 0x7ff73e39e77f in std.build.Builder::std.build.Builder.make (build.obj)
            try self.makeOneStep(s);
                                ^
C:\Users\sharp\dev\zig\lib\zig\std\special\build_runner.zig:133:17: 0x7ff73e384694 in main (build.obj)
    builder.make(targets.span()) catch |err| {
                ^
C:\Users\sharp\dev\zig\lib\zig\std\start.zig:173:65: 0x7ff73e3819ae in std.start.WinStartup (build.obj)
    std.os.windows.kernel32.ExitProcess(initEventLoopAndCallMain());
                                                                ^
Unable to dump stack trace: FileNotFound
error: the following build command failed with exit code 2147483651:
```

[...]

Issue #158, opened 2020-09-15
zls crashes every time on start. (.OBJECT_NAME_INVALID => unreachable)


doongjohn:
This happens every time when I start vscode.

crash output:
```zig
reached unreachable code
D:\zig\lib\zig\std\os\windows.zig:113:37: 0x7ff6c75b7839 in std.os.windows.OpenFile (zls.obj)
            .OBJECT_NAME_INVALID => unreachable,
                                    ^
D:\zig\lib\zig\std\os.zig:4092:31: 0x7ff6c75bdd80 in std.os.realpathW (zls.obj)
        const res = w.OpenFile(pathname, .{
                              ^
D:\zig\lib\zig\std\os.zig:4040:25: 0x7ff6c75ae87d in std.os.realpath (zls.obj)
        return realpathW(pathname_w.span(), out_buffer);
                        ^
D:\zls\src\main.zig:1593:75: 0x7ff6c759481b in main (zls.obj)
            zig_exe_path = try std.mem.dupe(allocator, u8, std.os.realpath(full_path, &buf) catch continue);
                                                                          ^
D:\zig\lib\zig\std\start.zig:154:65: 0x7ff6c7591f1e in std.start.WinMainCRTStartup (zls.obj)
    std.os.windows.kernel32.ExitProcess(initEventLoopAndCallMain());
                                                                ^
Unable to dump stack trace: FileNotFound
```
platform: `Windows 10`
zls client: `zls-vscode`
zig version: `master (0.6.0+9fe4c8923)` (used to build zls)
zig is in my `PATH`

Edit:  
Fixed by using `zls.json` and providing the path to zig.

~~I think this happened because of my antivirus software.~~ It was not related to antivirus software.

Edit:

`zls.json`
```json
{
    "zig_exe_path": "D:\\zig\\zig.exe"
}
```  
└> this works.

```json
{
    "zig_exe_path": "D:\\zig"
}
```  
└> this works but can't start `zig fmt` process.

balenamiaa:
` zig_exe_path = try std.mem.dupe(allocator, u8, std.os.realpath(full_path, &buf) catch continue);`

This seems to be because of https://github.com/ziglang/zig/issues/6044#issue-678355085. Short term you can perhaps call https://docs.microsoft.com/en-us/windows/win32/api/shlwapi/nf-shlwapi-pathcanonicalizew on full_path before sending it to std.os.realpath as this only happens on windows AFAIK.

alexnask:
[...]

PR #272, opened 2021-03-30
Refactor config wizard

### LINKED ISSUE #265: zig_exe_path is set incorrectly when using the snap version of Zig
`zig build config` finds `/snap/bin/zig` from my `PATH` variable. This is a symlink: `/snap/bin/zig -> /usr/bin/snap`. This symlink is resolved using `std.os.realpath` in https://github.com/zigtools/zls/blob/3a8d4c7bfb5daeaec5ff55fa8cf39e7fe0f82f9d/build.zig#L47
Which results in zig_exe_path being set to `/usr/bin/snap`.

What is the reason for resolving a potential symlink? What other use case would it break to remove this resolution?

InterplanetaryEngineer:
As mentioned in https://github.com/zigtools/zls/issues/229#issuecomment-808065657.
Closes #265 and possibly #194 and #158.

We can't have emoji, so let's use some good-old-fashioned ASCII art!

### DIFF
--- a/src/document_store.zig
+++ b/src/document_store.zig
@@ -422,6 +422,7 @@ fn refreshDocument(self: *DocumentStore, handle: *Handle) !void {
         still_exists: {
             for (new_imports) |new| {
                 if (std.mem.eql(u8, new, old)) {
+                    handle.imports_used.items[i] = new;
                     break :still_exists;
                 }
             }


alexnask:
Very nice!
I will test this soon and merge if there are no issues :)

InterplanetaryEngineer:
Didn't know there was a Discord! Just joined it.

InterplanetaryEngineer:
?

PR #284, opened 2021-04-02
Added signature help request handling

alexnask:
TODO:  
- Fix session tests


alexnask:
Example in action:  

Screencast 2021-04-03 03_49_11

PR #301, opened 2021-04-05
Add `insertText` field for enum completions

bnjmnt4n:
Another bug with the older version of lsp-mode on Emacs prevents the label from being inserted when there is an empty `insertText`. This commit adds an `insertText` property to enum completions for consistency with the rest of the completion items.


alexnask:
Thanks, I missed this one case when I added the insertText to the rest of the completions

PR #303, opened 2021-04-07
Improve testing infrastructure, refactor makeScopeInternal and resolveTypeOfNodeInternal, and support "catch" scopes

InterplanetaryEngineer:
It should now be a bit easier to set up a test and see how it failed.

I'll start working on some some regression tests next.


alexnask:
Thanks again for the great PR!

PR #308, opened 2021-04-15
Fix inclusion of toplevel doc comments, remove @async recursion in writeNodeTokens, add a few regression tests

InterplanetaryEngineer:
I don't really have a way to try out the semantic tokens; someone should check that they still work as intended.

### DIFF
--- a/src/analysis.zig
+++ b/src/analysis.zig
@@ -17,12 +17,7 @@ pub fn deinit() void {
     resolve_trail.deinit();
 }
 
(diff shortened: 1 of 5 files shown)


InterplanetaryEngineer:
Does someone know how to set up Emacs's lsp to show semantic tokens?

alexnask:
@InterplanetaryEngineer It sseems pretty obscure, I cannot find any instructions (aside from a bit of source code)  
The easiest editor to use for testing is probably vscode, followed by kak-lsp in my experience.

PR #316, opened 2021-05-02
Simplify semantic_tokens, fix off-by-one error in identifierFromPosition

InterplanetaryEngineer:
This should allow me to next remove the async recursion in a way that actually works.

### DIFF
--- a/src/offsets.zig
+++ b/src/offsets.zig
@@ -80,7 +80,7 @@ pub fn lineSectionLength(tree: ast.Tree, start_index: usize, end_index: usize, e
 
         const codepoint = try std.unicode.utf8Decode(source[i .. i + n]);
 
-        result += 1 + @as(usize, @boolToInt(codepoint >= 0x10000));
+        result += 1 + @boolToInt(codepoint >= 0x10000);
         i += n;
     }
     return result;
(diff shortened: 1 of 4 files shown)


src/offsets.zig
```diff
@@ -80,7 +80,7 @@ pub fn lineSectionLength(tree: ast.Tree, start_index: usize, end_index: usize, e
 
         const codepoint = try std.unicode.utf8Decode(source[i .. i + n]);
 
-        result += 1 + @as(usize, @boolToInt(codepoint >= 0x10000));
+        result += 1 + @boolToInt(codepoint >= 0x10000);
```
alexnask, line 83: The coercion here is intentional, @boolToInt will return `u1` and thus the whole expression is of type `u1` and will overflow when the condition is true.

alexnask (changes requested):
The rest of this code looks fine  
I will be testing this in the morning, thanks for the contribution

alexnask:
The CI started erroring out with BrokenPipe while running the session tests a couple of days ago, I will investigate and fix this.

src/offsets.zig
alexnask: Maybe
```zig
result = result + 1 + @boolToInt(codepoint >= 0x10000);
```
is saner here for this reason

InterplanetaryEngineer: Thanks for the catch, fixed it.

alexnask (approved):

alexnask:
Looks good, thanks again.

PR #320, opened 2021-05-07
[bug-fix]hover struct field not show docs.

Luna1996:


alexnask:
Thank you for the fix :)

PR #328, opened 2021-05-13
Update docs to reflect github.com/ziglang/zig/pull/8638

InterplanetaryEngineer:
The Zig master branch now has `zig-out` as the default installation path instead of `zig-cache`


alexnask:
Thanks

PR #348, opened 2021-06-24
Update for 0.8.0

### LINKED ISSUE #347: Update the release to latest (Zig 0.8.0 build)

Jan200101:
fixes #347


leecannon:
This will break building with zig master

Jan200101:
Hmm yeah I had not consulted the newer codebase to check for changes
While personally I think its worth supporting building for the latest release I can also see why supporting 0.8.0 is pointless considering the way forward

PR #359, opened 2021-07-09
Data gen improvements

### LINKED ISSUE #358: translate <code> when generating data
Currently, `code` tags in html like `<code class="zig"><span class="tok-builtin">@asyncCall</span></code>` are ignored and translated to plain `@asyncCall`.
However, ` can be used to turn a piece of text into code, which can improve user experience.

Prince213:
Fixes #358.


alexnask:
Very nice!  
Thanks a lot, I hope the javascript wasn't too bad, I wrote it really fast without really thinking about long term maintenance.

Prince213:
Yeah, it would be nicer if we have better comments.
Anyway thanks for the great software!

PR #361, opened 2021-07-10
Introducing new method of generating data files

Prince213:
A new file src/data/generate-data.py has been added.
It is written in Python 3 and has some new features like:
- It doesn't need Chrome to work, the script fetches the page automatically;
- It does the translation of HTML tags for better user experiences in editors.

A new section has been added to README.md to reflect this change.
Also master.zig has been updated.


alexnask:
Nice, thank you :)

Mouvedia:
@alexnask where have you been mate ?

PR #362, opened 2021-07-11
Simplifies logic in generate-data.py and update master.zig

Prince213:
Depends on https://github.com/ziglang/zig/pull/9357.


SuperAuguste:
Dependent unmerged.

Issue #372, opened 2021-08-09
Problem compiling with zig 0.9.0-dev.755+799fedf61


InfRandomness:
Hi, I seem to have an issue building zls with zig 0.9.0-dev.755+799fedf61

```
./src/setup.zig:228:31: error: expected type 'type', found '[]u8'
    var it = std.mem.tokenize(env_path, &[_]u8{std.fs.path.delimiter});
                              ^
./src/setup.zig:68:24: note: referenced here
    var zig_exe_path = try findZig(allocator);
                       ^
./src/main.zig:1805:13: note: referenced here
            try setup.wizard(allocator);
            ^
./src/known-folders/known-folders.zig:149:38: error: expected type 'type', found '*[8192]u8'
        var line_it = std.mem.split(&read, "\n");
                                     ^
zls...The following command exited with error code 1:
/home/build/zig/build/zig build-exe /home/build/zls/src/main.zig --pkg-begin build_options /home/build/zls/zig-cache/zls_build_options.zig --pkg-end --cache-dir /home/build/zls/zig-cache --global-cache-dir /home/build/.cache/zig --name zls --pkg-begin known-folders /home/build/zls/src/known-folders/known-folders.zig --pkg-end --enable-cache
error: the following build command failed with exit code 1:
/home/build/zls/zig-cache/o/0d41ce75e2baa15549c0a336c4e164e9/build /home/build/zig/build/zig /home/build/zls /home/build/zls/zig-cache /home/build/.cache/zig
```

jbsolomon-fw:
pretty simple solution: add `u8` as the first parameter of `split` and `tokenize`.

PR #377, opened 2021-08-17
Update "known-folders" submodule

sagehane:
zls doesn't seem to build without updating the `known-folders` to https://github.com/ziglibs/known-folders/commit/234b589d3cd0c00f2f675f98fd9674e208fd2bd0.

The cause is https://github.com/ziglang/zig/commit/05fd20dc104b3654ce9c5d7a22a2bff66a940dba, and is the same as #372.


markstewart:
I ran into this same issue building `zls` with latest zig@f28868e8f. Can confirm this PR fixes the problem.

Issue #387, opened 2021-09-02
std-os-reorg breaks ZLS


sagehane:
https://github.com/ziglang/zig/pull/9618 breaks ZLS.

sagehane:
Resolved as of https://github.com/zigtools/zls/commit/c5414796a49dd5115a3ec398ed6ce6363ac6a8c8.

PR #388, opened 2021-09-02
Fix namespaces

sagehane:
Addresses most of #387. There's still a compile error because zinput (one of the submodules) needs to be fixed too.

If somebody has a better idea than spamming `@import("ast.zig")`, feel free to suggest one.

I had no idea what `log.debug("{s}", .{name_decl.key});` did, so I commented it out.
Maybe it was supposed to be inside this loop?
https://github.com/zigtools/zls/blob/c2cbc05135aebb4a9fa24cd60a4a091db18605fb/src/analysis.zig#L2461-L2463

### DIFF
--- a/src/analysis.zig
+++ b/src/analysis.zig
@@ -1,10 +1,22 @@
 const std = @import("std");
 const DocumentStore = @import("document_store.zig");
 const ast = std.zig.ast;
+const fnProto = ast.Tree.fnProto;
+const lastToken = ast.Tree.lastToken;
 const types = @import("types.zig");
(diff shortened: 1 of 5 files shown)


sagehane:
Actually, a bit more work has to be done on top. https://github.com/ziglang/zig/commit/3940a1be18a8312dec9d521c6e30ec4d34c44bd6 means that all instances of `ast` needs to be renamed to `Ast`, along with `Ast.Tree` being reworked, etc.

sagehane:
Closing in favor of #392 as it looks far more complete.

PR #393, opened 2021-09-04
Update zinput submodule

sagehane:
Update the `zinput` submodule, which should get ZLS to compile on the latest Zig along with #392.


Sashiri:
😂 I guess we do think alike, pushed it to #392 a while ago.
I'm not sure if it should be there tho, if you think updating submodule should be a separate PR then I'll hard reset branch by one ;D

sagehane:
Oh, I'll close this branch then. There's no need for duplicates.

PR #399, opened 2021-09-29
support new ast token changes and saturating operators

travisstaloch:
- some ast token tags were renamed: ie .bit_shift_left => .shl
- add initial support for saturating operators: +|, -|, *|, <<|, +|=, -|=, *|=, <<|=

Not sure if this is in anyway complete.  I just got the build to complete after https://github.com/ziglang/zig/pull/9679 was merged today.  I can see that zls no longer adds squiggles to +| and +|= for example.  Let me know if anything else needs to be done.


nektro (approved):
Can confirm this works with a fresh source install of Zig but going to hold off on the merge until https://github.com/kristoff-it/zig-doctest/pull/9 is handled so that the saturation arithmetic changes can make their way to the master tarball downloads

PR #411, opened 2021-11-02
Update Sublime LSP settings for Sublime Text 4

squeek502:
"selector" is now the preferred method of configuration, everything else was deprecated in SublimeLSP 4070-1.1.0: https://github.com/sublimelsp/LSP/releases/tag/4070-1.1.0

Also syncs setup.zig with the readme for the Sublime Text 3 settings (the Zig Language package name was wrong in setup.zig)


nektro:
one comment being can you put 4 after 3

squeek502:
> one comment being can you put 4 after 3

Done

PR #415, opened 2021-11-04
Change notification method in log

omaraaa:
Currently in the log function in main, the notification method used to send logs to the client is "window/showMessage". This creates an a notification on the UI on VSCode that can cause slow downs. I propose a change to use the  "window/logMessage" method instead, which is intended to be used for logging.


SuperAuguste:
I agree with this change; we should use a separate function or perhaps different log level for messages that are meant to display on the user interface.

InKryption:
Thank you again for this fix. I swear I was going nuts trying to figure out what was wrong.

PR #420, opened 2021-11-16
only send format editings when necessary

tw4452852:
If the original document is same as the formatted one, there's no need to send the unchanged
document's content back which will make the client confused.

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -1650,6 +1650,8 @@ fn formattingHandler(arena: *std.heap.ArenaAllocator, id: types.RequestId, req: 
 
         switch (try process.wait()) {
             .Exited => |code| if (code == 0) {
+                if (std.mem.eql(u8, handle.document.text, stdout_bytes)) return try respondGeneric(id, null_result_response);
+
                 return try send(arena, types.Response{
                     .id = id,
                     .result = .{


SuperAuguste:
This change is unnecessary and will also brick some clients that will wait for a format response they will never receive. The other alternatives are to send an empty or error response, but those would break clients too, thus the status quo is better. :)

tw4452852:
Hi @SuperAuguste 

According to the [spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_formatting), if there's nothing change, we could return `null` `textedits` in response, what do you think?

SuperAuguste:
That would work, but I still do not see why this is a necessary change.

tw4452852:
I found this issue with my `acme` editor, it will trigger formatting on saving, but every time it gets a non-empty response, it will mark the current file dirty and format the content although the formatted one and original one are the same. It's kind of annoying and I didn't encounter this problem when using another language server (e.g. `gopls`, `rust-analyzer`).

SuperAuguste:
I see! If you make format return a `null` instead of nothing, I'll merge this PR. Thanks!

tw4452852:
Hi @SuperAuguste 

I just rebased the patch onto the latest master and response with a null result when no changes.

SuperAuguste:
LGTM, thank you so much for your contribution! :)

tw4452852:
Yeah, feel happy if my fix could help others.

Issue #134, opened 2020-07-11
Support for workspace/configuration and workspace/didChangeConfiguration


CherryMan:
This feature would allow the client to configure `zls` itself, so users can configure `zls` through their respective editor instead of using `zls.json`.

If this feature request is approved I'll gladly work on a PR.

alexnask:
This sounds good to me, let me know if you want to work on it, I will implement it otherwise.

tamago324:
Hello.
I am interested in developing this feature.

I don't think there is currently a mechanism for the server to send requests to the client, are there any plans to implement that?
I would like to implement a mechanism for sending requests, but I'm new to zig, so I would appreciate some advice if you can help me.

PR #436, opened 2021-12-17
implement workspace/configuration

vesim987:
As far as i can tell this implementation comply to the LSP specification. I tested this with neovim and it is working fine.

Fixes: #134


SuperAuguste:
Sorry about getting to this PR so late! I'll pull this in locally and test it out real quick and then merge because this LGTM! :)

PR #437, opened 2021-12-20
c_void -> anyopaque

Sirius902:
Replaces occurrences of `c_void` to `anyopaque` as per ziglang/zig@9f9f215305389c08a21730859982b68bf2681932. The only occurrence I could find was in `src/analysis.zig`.

Edit: To compile on zig master `known-folders` will need to be updated after ziglibs/known-folders#26 gets merged.


aconbere:
Hit this today too!

PR #448, opened 2022-01-12
add type function symbols to document symbol outline

ikrima:
### Description
ZLS skips Type Functions when enumerating the document symbols. This leads to a lot of standard library files looking like this in the VSCode Document Outline:
image

With the PR, it looks like:
image
Go To Document Symbol also works: 
image

### Change
Recurse into functions if they are a type function

### Questions
I'm not too familiar with VSCode extensions or the ZLS codebase. I ran the tests and did some sanity manual testing but please review accordingly

- `getDocumentSymbolsInternal`: from looking at the code + cross referencing VSCode Api, looks like its just used to enumerate document symbols (call tree: `documentSymbol()->getDocumentSymbols()->getDocumentSymbolsInternal()`)
  - sanity check: are there any other places this modification needs to be propogated?
  - sanity check: specifically so that symbol resolve/rename/goto def/etc aren't broken?
- error handling: unsure what the pre/post conditions are of the parse tree when `getDocumentSymbolsInternal` is invoked
  so there might be superflous guards/checks
- any tests to add?

### DIFF
--- a/src/analysis.zig
+++ b/src/analysis.zig
@@ -1809,6 +1809,16 @@ fn getDocumentSymbolsInternal(allocator: std.mem.Allocator, tree: Ast, node: Ast
                 if (var_decl.ast.init_node != 0)
                     try addOutlineNodes(allocator, tree, var_decl.ast.init_node, &child_context);
             }
+            if (tags[node] == .fn_decl) fn_ch: {
+                const fn_decl = tree.nodes.items(.data)[node];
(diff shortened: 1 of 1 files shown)


SuperAuguste:
LGTM, but I'll have to manually check later when I'm available :)

SuperAuguste:
I don't believe anything else needs to be added / modified. Thanks!

PR #451, opened 2022-02-01
Fix build on zig master (0.10.0-dev.500+66cf011aa)

SebastianKeller:
std.process.args() has a slightly different api now, breaking the build
on zig master.


SuperAuguste:
Thanks!

Issue #410, opened 2021-10-30
zls cannot locate zls.json file


palash25:
hi, i was installing zls on a new machine and noticed this problem. this is what I did

built zls from source and than ran the configuration wizard. The config wizard successfully wrote the config to `/etc/zls.json`

after that I copied the zls binary that was built in zls-out directory to `/usr/local/bin`

then i installed the vscode extension for zls and set the following option in settings

```
"zigLanguageClient.path": "/usr/local/bin/zls"

```

and restarted zls from vscode and I get this in the logs

```
[info-main] No config file zls.json found.
[info-main] Server closing.
```

does zls not look for the config in the `/etc` directory?

leecannon:
zls uses https://github.com/ziglibs/known-folders to fetch locations for configuration.

below is the code that checks both the local configuration and global configuration folders for the file
https://github.com/zigtools/zls/blob/80f09831bc4be779af8a1bf8fadda1357faa302e/src/main.zig#L1701-L1726

looking at known-folders shows global config to be defined as
```zig
.global_configuration = XdgFolderSpec{ .env = .{ .name = "XDG_CONFIG_DIRS", .user_dir = false, .suffix = null }, .default = "/etc" },
```

so unless you have the XDG_CONFIG_DIRS environment variable set to something it should default to /etc

palash25:
hey thanks for the quick response, i see the env var you mentioned is indeed set, so if I understand correctly if this env var is unset zls will look for the config in `/etc/` ?


```
echo $XDG_CONFIG_DIRS
/etc/xdg/xdg-pop:/etc/xdg

```

leecannon:
Yea looking at known-folders's code it will check for the the environment variable and if not found will fallback to the default, in the case of global configuration they have the env variable as `XDG_CONFIG_DIRS` and the default as `/etc`

I dont know enough about XDG to know if that is correct, if you dont think it is quite right raise an issue over there.

PR #455, opened 2022-02-08
Fix ZLS config not found crash #410

davidgm94:
Hi there, first open source contribution to other project. Please review the PR and comment if you see anything wrong.
This fixes a crash when the ZLS configuration file could not be found and yet still the value of the string wasn't null.

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -1721,6 +1721,7 @@ pub fn main() anyerror!void {
             }
         }
         logger.info("No config file zls.json found.", .{});
+        config_path = null;
     }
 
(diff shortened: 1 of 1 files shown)


S0urc3C0de:
Is there a reason, this has not yet been merged? I've had this fix running locally for weeks now (independently from this PR, though).
The problem is simply that `config_path` is set to a path **before** it has been checked whether there's actually a config file there - and then not being reset to null if not. Ideally, it should only be set **after** the call to `loadConfigInFolder`, that is, within the If-statement.

PR #461, opened 2022-02-15
Update setup wizard for ST4

alvv-z:
In my tests, ST4 doesn't need `include_at_in_builtins` to be `true`, since the autocomplete code was updated before build 4058.


SuperAuguste:
Oops, I missed this! Can you please deconflicts the merges! Thanks; I'll merge it ASAP :)

PR #463, opened 2022-02-16
Skip char and string literals when highlighting comments

cascade256:
The comment highlighting appears to take priority over string and char literals, leading to syntax highlighting like this:
image

These changes should make it skip string and char literals when scanning for comments.


Jarred-Sumner:
nice

nektro:
can you post a screenshot of what it looks like with these changes

cascade256:
Yep! Here is what it looks like:
image

SuperAuguste:
Thanks!

PR #474, opened 2022-04-14
Add step to Kate installation instructions

029onikyu:
Without this, the editor will not attempt to launch the language server as it does not have any notion of what language a `.zig` file is for.


SuperAuguste:
Good call!

Issue #475, opened 2022-04-16
Panic when showing suggestions for error


v1nh1shungry:
If I type `error` and dot to get the suggestions, zls will sometimes (not always) panic.

https://user-images.githubusercontent.com/98312435/163659532-82a8e830-4381-4be2-884d-1a7d8e88e218.mp4

The zls which I built from source gives more detailed messages:

```
thread 5080 panic: reached unreachable code
D:\msys64\home\lenovo\.zls\src\offsets.zig:128:35: 0x7ff600d08aec in offsets.tokenRelativeLocation (zls.obj)
                const codepoint = try std.unicode.utf8Decode(source[i .. i + n]);
                                  ^
D:\msys64\home\lenovo\.zls\src\analysis.zig:1744:68: 0x7ff600d21908 in analysis.getDocumentSymbolsInternal (zls.obj)
    const end_loc = start_loc.add(try offsets.tokenRelativeLocation(
                                                                   ^
D:\msys64\home\lenovo\.zls\src\main.zig:1595:53: 0x7ff600ca1278 in processJsonRpc (zls.obj)
                    done = extractErr(method_info[2](arena, id, request_obj, config));
                                                    ^
D:\msys64\home\lenovo\.zls\src\main.zig:1870:27: 0x7ff600c6b3bb in main (zls.obj)
        try processJsonRpc(&arena, &json_parser, buf, config);
                          ^
D:\msys64\home\lenovo\.zig\lib\std\start.zig:333:65: 0x7ff600c64151 in std.start.WinStartup (zls.obj)
    std.os.windows.kernel32.ExitProcess(initEventLoopAndCallMain());
                                                                ^
```

* PC: Windows 10 21H2
* zls version: 0.9.0
* ZLS for VSCode version: 1.0.4

v1nh1shungry:
I found an interesting fact about this. If I do `const std = @import("std");` in the global scope, then every time I type `error` and a dot it will panic. However, if I didn't import it then everything goes normal.
* zig version: 0.10.0-dev.1837+4c83b11f7

v1nh1shungry:
I found that if there is anything above the function where the keyword `error` is, zls will crash.
```zig
// any of these above the function contains the keyword `error`
const std = @import("std");

// const T = struct {
//     member: i32
// };

// fn func() void {
    
// }

pub fn main() void {
    error
}
```

PR #476, opened 2022-04-20
fix ast.lastToken() does not handle empty error_value

v1nh1shungry:
Fix https://github.com/zigtools/zls/issues/475
When we type only `error`, both the `lhs` and `rhs` of  the `error_value` node are 0. However, ast.lastToken() does not handle this, and it can cause zls to crash. This commit is to fix this.


SuperAuguste:
Thanks!

Issue #291, opened 2021-04-04
Missing symbols in the MacOS build?


InterplanetaryEngineer:
Mentioned in https://github.com/sublimelsp/LSP/issues/1634#issuecomment-813037711.

I do not have a Mac to try and reproduce this tho...

ron-wolf:
I am able to reproduce. I get the same exact message, in fact:

<pre>
zig: dyld: Symbol not found: __tlv_bootstrap
zig:   Referenced from: /Users/Ron/.local/bin/zls
zig:   Expected in: /Users/Ron/.local/bin/zls
zig:  in /Users/Ron/.local/bin/zls
</pre>

agbell:
I get this as well
```
[Info  - 11:12:40 AM] Connection to server got closed. Server will restart.
dyld: Symbol not found: __tlv_bootstrap
  Referenced from: /Users/adam/sandbox/zls/zls
  Expected in: /Users/adam/sandbox/zls/zls
 in /Users/adam/sandbox/zls/zls
[Info  - 11:12:40 AM] Connection to server got closed. Server will restart.
dyld: Symbol not found: __tlv_bootstrap
  Referenced from: /Users/adam/sandbox/zls/zls
  Expected in: /Users/adam/sandbox/zls/zls
 in /Users/adam/sandbox/zls/zls
[Error - 11:12:40 AM] Connection to server got closed. Server will not be restarted.
```

NduatiK:
Same issue. 

Building from source works well and because zig is so fast at compiling, its doable.

alxiong:
same issue. (btw, I'm on a M1 chip)

Also tried building from source, got this: 
```
$ zig build -Drelease-safe
 build -Drelease-safe
./build.zig:12:26: error: no member named 'addOptions' in struct 'std.build.Builder'
    const exe_options = b.addOptions();
                         ^
/opt/homebrew/Cellar/zig/0.8.1/lib/zig/std/special/build_runner.zig:170:24: note: referenced here
        .ErrorUnion => try root.build(builder),
                       ^

$ zig version
0.8.1
```

nektro:
did you build from source? what zig version are you on?

andyfleming:
Building from `zls@master` with zig `0.9.0-dev.1352+6d6cf5984` appears to have worked. :+1:

PR #481, opened 2022-04-30
update for latest zig master branch zig-0.10.0-dev.2017+a0a2ce92c

neel-bp:
std.childprocess.init no longer returns an error union type
std.childprocess no longer has deinit member function.


damag:
Hi, it seems that it no longer allocates a struct on the heap but returns it directly, so there is no need to deinit/destroy since &process would be an address on the stack.
You can see for examples of how the usage was updated in Zig itself: https://github.com/ziglang/zig/commit/a0a2ce92ca129d28e22c63f7bace1672c43776b5

neel-bp:
okay i removed the defer, 
thank you for your patience i am pretty new to zig, and when i tried compiling zls i was having compilation errors, so i tried to fix it, and was able to do it, and decided to fork it and make PR, 
don't feel bad if you people think this pr is not up to standards.

damag:
No worries, I'm not a maintainer, just another user that ran into the same issue as you, but removing the `try` and `deinit` line should do it. ;)

damag:
Just noticed that the `try` in tests/sessions.zig line 111 is also causing the CI checks to fail.

jacmoe:
Thanks, guys! The changes in those two files allowed me to (finally!) get latest Zig and latest Zls to get along :smile:

neel-bp:
updated startZls test, thanks @damag

damag (commented):
Just a few more things for `zig build test` to run successfully:

- The return type of `startZls()` should be changed to `!std.ChildProcess` since process is no longer a pointer.
- Likewise, `process` in the Server struct should be a simple `std.ChildProcess`
- And finally `self.process.deinit();` should be removed in the `Server.shutdown()` function

truemedian:
`file://./tests/test.zig` is a valid URI, if it doesn't work it needs to be fixed too.
`file:///./tests/test.zig` is just `file:///tests/test.zig` which is an absolute path, not relative like the test (probably) intends

guidorice:
In #291 this PR is mentioned as being the best candidate to zls working again. What else needs to be fixed here? File uris changed in zig or something? This branch seems to work to me , and `zig build test` succeeds (x86_64-macos). 

cc @damag @leecannon @neel-bp

damag:
It works for me as well, although I agree with @truemedian, I don't know what the file uri change is about.

SuperAuguste:
This PR looks good, I agree that the tests failing is unusual... Could you please rebase this (just merged suirad's build fix); thanks!

PR #492, opened 2022-05-28
update build.zig for recent zig changes

suirad:
Update `build.zig` to reflect changes of ziglang/zig/pull/11557


SuperAuguste:
Thanks!

PR #501, opened 2022-06-05
Ignore unimplemented notifications

gpanders:
Responding to an unimplemented notification can cause errors for some
LSP clients that do not expect to receive a response after sending a
notification. Differentiate between unimplemented requests and
notifications and only send a generic null response for the former while
silently ignoring the latter.


SuperAuguste:
Good call! LGTM!

PR #505, opened 2022-06-09
fix off by one error in documentRange

kristoff-it:
Previously ZLS would report the wrong line number as the range end when replying to a `format` request. This is why Helix refused to apply any change.

This same problem might be present somewhere else, I only checked this function in order to fix my own problem.


SuperAuguste:
LGTM, thanks Loris!

PR #507, opened 2022-06-23
Label details support

ryuukk:
This enables label details support for the server for compatible clients

I don't really like the implementation, i left a note, it is very loosy and to be honest, i don't think we should rely on it, but it gets the job done for now, and the result is anyways, IMO, much better than before, so i guess it worth keeping for now, i made sure it sits in a separate function and is not coupled to the entire program, so it is easy to just remove

It can get very noisy though, so maybe some more work needs to be done to slim things a little bit and only show more function detail when the item is selected/hovered for example, but i'm not sure how to do that yet!

To enable it, you'll also need an updated version of the vscode plugin, i sent a PR here: https://github.com/zigtools/zls-vscode/pull/33

There is also a bug in ZLS somewhere (not related to this PR), so it should be investigated, please read the PR on the vscode plugin repo for details

Here some screenshots: 

image

image

image

image

image

And here the noisy screenshot

image

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -89,6 +89,7 @@ const ClientCapabilities = struct {
     supports_semantic_tokens: bool = false,
     hover_supports_md: bool = false,
     completion_doc_supports_md: bool = false,
+    label_details_support: bool = false,
 };
 
 var client_capabilities = ClientCapabilities{};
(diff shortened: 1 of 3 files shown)


SuperAuguste:
This is absolutely awesome, thank you so much! I believe tests still need updating and that should be it! I think it best to merge your other PR first and then rebase this one so we don't have conflicts with the `kindToSortScore` functionality. :)

PR #508, opened 2022-06-24
Ease debugging package loading from build_runner

MageJohn:
I was having some issues with getting autocompletion of packages to work correctly in my project, and discovered that ZLS was ignoring exit codes other than 0, rather than reporting a failure. This PR fixes this.

I also changed the log level for package loading failure from debug to warn. If it happens it's usually a user problem, but it can look like a ZLS problem at first, so I think it's good to be proactive in telling users that there might be a problem with their build file.

### DIFF
--- a/src/DocumentStore.zig
+++ b/src/DocumentStore.zig
@@ -190,6 +190,8 @@ fn loadPackages(context: LoadPackagesContext) !void {
                         };
                     }
                 }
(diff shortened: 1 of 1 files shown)


SuperAuguste:
Succinct but effective fix, thank you so much! I'll let the CI run and then merge!

PR #515, opened 2022-07-09
Fix build.zig typo for enable_tracy using better wording

Sepruko:
Fixes the misspelling of `to` as `of` by removing `of not` at [build.zig#L25](https://github.com/zigtools/zls/blob/5838a34101343e801f19cb080c0dc70c68dd0c93/build.zig#L25) as it is not required for the sentence to make sense.


SuperAuguste:
based comment pr

PR #516, opened 2022-07-09
Fix renaming struct fields

### LINKED ISSUE #398: Can't rename struct property
When I try to rename the property of a struct nothing happens. 

https://user-images.githubusercontent.com/8743306/135007174-d9783621-a3c7-404b-8d52-db9dd51900a2.mov

I'm using neovim with lsp, and compiled zls from HEAD two days ago.

I'm quite new to Zig and zls, so I'm not sure if this is a bug or just a current limitation of zls.

Sepruko:
Closes #398


SuperAuguste:
To continue our conversation on Discord, we've got to make sure this doesn't break anything else;
image

Sepruko:
Found a case where ZLS does not rename all instances of a field referenced by field access, however I do not know how to reproduce it at this point.

PR #527, opened 2022-07-12
stream parsing of build runner output

vesim987:
Better solution for #459


SuperAuguste:
lets go!! this is exactly what I had in mind! tysm vesim!

SuperAuguste:
This LGTM, any future changes in mind? :)

SuperAuguste:
Stale :(

Issue #468, opened 2022-03-02
Release an arm64 / macOS apple silicon binary


leighmcculloch:
The most recent release of zls (https://github.com/zigtools/zls/releases/tag/0.9.0) includes only x86 binaries. It'd be helpful for developers on the current generation of Apple computers to be able to use zls.

Could builds for arm64 macOS be included in the next release?

jiacai2050:
It seems GitHub action only provide x86 machines, I wonder if cross compile could work here
- https://github.community/t/does-github-actions-cloud-allow-running-of-arm-based-instances/119319

leighmcculloch:
I have had success cross-compiling for arm64 macOS with Rust and Go, so I assume it might be possible with Zig, but I haven't tried yet.

jiacai2050:
Sad, Zig(47c58cba5) cann't compile zls(latest ba68a4d)

```
 (master)$ ~/downloads/zig-macos-aarch64-0.10.0-dev.3024+47c58cba5/zig build -Drelease-safe
/Downloads/zig-macos-aarch64-0.10.0-dev.3024+47c58cba5/lib/zig/std/fs.zig:960:25: error: only 'IterableDir' can be iterated; 'IterableDir' can be obtained with 'openIterableDir'
    pub const iterate = @compileError("only 'IterableDir' can be iterated; 'IterableDir' can be obtained with 'openIterableDir'");
                        ^
error: zls...
error: The following command exited with error code 1:
/Downloads/zig-macos-aarch64-0.10.0-dev.3024+47c58cba5/zig build-exe /code/misc/zls/src/main.zig -OReleaseSafe --cache-dir /code/misc/zls/zig-cache --global-cache-dir /.cache/zig --name zls --pkg-begin build_options /code/misc/zls/zig-cache/options/ZN5T5qs__MtT8K02r_1S-BHCIQm36qquqqxMkxJvDpZqrwQ-wS_m2895sSwEs1B_ --pkg-end --pkg-begin known-folders /code/misc/zls/src/known-folders/known-folders.zig --pkg-end --enable-cache 
error: the following build command failed with exit code 1:
/code/misc/zls/zig-cache/o/f7ef6c24324123d6a9875be3d5afb20b/build /Downloads/zig-macos-aarch64-0.10.0-dev.3024+47c58cba5/zig /code/misc/zls /code/misc/zls/zig-cache /.cache/zig -Drelease-safe


```

jiacai2050:
[...]

PR #544, opened 2022-07-17
update to IterableDir changes in Zig std

Vexu:
Depends on https://github.com/ziglang/zig/commit/fcaeca5b0acc4cc5930756301f56f7b56dc0a020

PR #545, opened 2022-07-17
fix openDir with iterate

jiacai2050:
See https://github.com/zigtools/zls/issues/468#issuecomment-1186441812


SuperAuguste:
Closing in favor of #544.

jiacai2050:
Hi @SuperAuguste, thanks for point out.

BTW, how often will https://ziglang.org/download/ update to latest master version?

I think it's important for user to build zls with those pre-built version.

SuperAuguste:
Absolutely! I'm not an authority on this topic but I believe that that page updates once a day!

Luukdegram:
It updates when the CI completes successfully on the master branch. So technically this could be multiple times a day.

PR #546, opened 2022-07-17
Refactor out most main globals!

SuperAuguste:
Self-explanatory; now `main.zig` is `Server.zig` and has way less globals, hooray!


SuperAuguste:
Passes local, fails CI because of iterate changes.

PR #549, opened 2022-07-19
Add flake support

leroycep:
This pull request will add `flake.nix`, `flake.lock`, and a build option `-Dknown-folders=<path>` to override the path to the `known-folders` package.

Will allow zls to be run like so:

```
nix run github:zigtools/zls 
```

Or used in another flake:

```nix 
{
    inputs.zls.url = "github:zigtools/zls";
    outputs = { self, zls}: {
        # flake stuff here
    };
}
```


SuperAuguste:
LGTM, thank you!

PR #557, opened 2022-07-23
Highlight orelse as a keyword

Techatrix:
colorizing it like `or` and `and` makes more sense

```zig
const value = (foo orelse bar) and (foo or bar);
```


SuperAuguste:
LGTM, odd CI failure.

PR #558, opened 2022-07-23
docs(neovim/vim): Update instruction for Coc

xiyaowong:


SuperAuguste:
Ooo awesome work! `coc-zls` seems great!

PR #561, opened 2022-07-24
build_runner: support 0.9 and master

leecannon:
Due to the `std.build.Pkg` change the build_runner.zig that zls uses does not function with 0.9, meaning no package completion.

Since Feburary I've been stuck with 0.9 half the time due to https://github.com/ziglang/zig/issues/10847 and I made this change to allow me to swap out the zig version and still get package completion.


SuperAuguste:
Wow, absolutely amazing catch Lee! Tysm!

PR #569, opened 2022-08-02
provide build_runner command on failure

### LINKED ISSUE #566: Proposal: dump stderr when loading packages from build.zig fails.
Ever since zls has had a way to load packages from a build.zig, there have always been weird, silent bugs that have prevented some (or all in some cases) users from having *any* package completions. Most of the silence is due to the fact that when zls exec's zig to run the build runner, it only returns an error indicating failure.

It would be helpful if zls *also* kept track of stderr and would log it before erroring out to the caller.

leecannon:
A few times i've had to read the code in `loadPackages` to figure out the correct command to execute the build_runner to debug an issue.
Closes #566

### DIFF
--- a/src/DocumentStore.zig
+++ b/src/DocumentStore.zig
@@ -134,29 +134,41 @@ fn loadPackages(context: LoadPackagesContext) !void {
     defer if (context.build_file_path == null) allocator.free(build_file_path);
     const directory_path = build_file_path[0 .. build_file_path.len - "build.zig".len];
 
+    const args: []const []const u8 = &[_][]const u8{
+        zig_exe_path,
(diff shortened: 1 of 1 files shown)


truemedian:
It would also be helpful if zls also logged stderr when the command fails like #566

leecannon:
@truemedian: i've added stderr output

Closes #566

truemedian:
ah sorry if I wasn't clear, I meant logging zig_run_result's stderr

leecannon:
fixed, that does make more sense :)

SuperAuguste:
LGTM, thanks again Lee!

Issue #536, opened 2022-07-15
Globals must go!


SuperAuguste:
The title is pretty self-explanatory: global variables/constructs must go!

- [ ] `main.zig`
  - [ ] `actual_log_level`
  - [x] `stdout`, which is actually a BufferedWriter to stdout
  - [x] `allocator`
  - [x] `document_store`
  - [x] `client_capabilities`
  - [x] `offset_encoding`
- [ ] `analysis.zig`
  - [ ] `using_trail`
  - [ ] `resolve_trail`

## Ideas

- We likely need a `Server` construct similar to our test runner's.

There are probably more globals that need addressing and ideas for addressing them, so leave a comment with any feedback!

PR #570, opened 2022-08-03
Revive Session tests

Techatrix:
With this PR i have removed the dependency on stdio and instead added a ` std.io.Writer` interface in `Server.zig`.
This allows us to intercept outgoing responses by the Server and more easily perform tests without having to run `zls` as a child process.

Additional notable changes include:
 - move main to `main.zig` and expose a small public interface in `Server.zig`
 - the arena allocator is not passed as an argument to various functions in `Server.zig` but instead stored in the Server struct.
 - logging can't be done globally (because of the decoupling from stdio) #536
 - tests now use `std.testing.allocator`

I wanted to store the writer in the Server struct itself and make it generic but i am not aware of a method to achieve this without having to wrap the entirety of `Server.zig` deeper scope.


SuperAuguste:
This is absolutely amazing; LGTM! Thank you so much! Logger changes are a little jarring but slightly less intuitive code > global state every time :P

SuperAuguste (approved):

PR #578, opened 2022-08-10
Only return highlights for symbols in current document

gpanders:
This seems to only occur when highlighting "field access" symbols, e.g.
"bar" in a statement such as `foo.bar`. This is because the
`symbolReferencesInternal` function finds the reference to the field in
the container even when the container is not in the current document.


gpanders:
Alternatively I could try modifying `symbolReferencesInternal` to have it (optionally) only find symbols within the current document. That is probably “better” in the sense that it does not generate references that only get filtered out later.

PR #579, opened 2022-08-13
Text format to return edited ranges instead of entire document

jefftime:
This pull request addresses the current behavior of ZLS's text formatting. Currently, ZLS will return the result of `zig fmt` as the replacement text for the entire range of the document. This causes odd issues in certain text editors such as [Helix](https://github.com/helix-editor/helix) which lose cursor position due to the whole document being changed

This pull request adds the ability to compare the input text and the `zig fmt` output in order to get proper text edit ranges. I am not an expert on the LSP specification by any means, but I believe this is more in mind with how text formatting is supposed to behave. In most cases it should allow for smaller messages between the client and server as well

### DIFF
--- a/src/Server.zig
+++ b/src/Server.zig
@@ -18,6 +18,7 @@ const Ast = std.zig.Ast;
 const tracy = @import("tracy.zig");
 const uri_utils = @import("uri.zig");
 const data = @import("data/data.zig");
+const diff = @import("diff.zig");
 
 // Server fields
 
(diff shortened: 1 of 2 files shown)


leecannon:
You are going to need to rebase on master to get the CI passing.

jefftime:
@leecannon Done

leecannon (approved):
Seems to work fine in vscode, I'm not setup for testing in any other editor.

PR #583, opened 2022-08-18
Improve CLI, fix leak, & other.

InKryption:
The leak occurred when specifying `--config-path`, followed by a failure to parse the specified file.

The rest of this PR is effectively just adding a bunch of infrastructure for the CLI, alongside resolving a TODO.

Also tweaked the line that conditionally initialized the tracy allocator, which was relying on a mutable temporary, which will be a compile error at some point.


SuperAuguste:
LGTM!

PR #584, opened 2022-08-18
2 minor tidy-ups + fix for over-eager unused parameter error in function types

InKryption:
ZLS should no longer complain about unused parameters in function types with this change.


SuperAuguste:
this is so slay tysm ink!! also LOL @ the main anyerror fix :P

PR #585, opened 2022-08-19
Improve unused variable report accuracy

InKryption:
With unused variable warnings enabled, this code
```
const blk = blk: {
    break :blk;
};
blk: for ([_]void{}) |_| break :blk;
blk: while (false) break :blk;
_ = .blk;
_ = .{ .blk = .foo };
```
would have previously reported the `blk` variable as being used. With some extra checks for some surrounding tokens, `blk` should now be correctly reported as unused.


SuperAuguste:
Fantastic, thanks!

PR #587, opened 2022-08-20
make `zig build` work with latest master

leecannon:
set `use_stage1` in build.zig

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -7,6 +7,7 @@ pub fn build(b: *std.build.Builder) !void {
 
     const mode = b.standardReleaseOptions();
     const exe = b.addExecutable("zls", "src/main.zig");
+    exe.use_stage1 = true;
     const exe_options = b.addOptions();
     exe.addOptions("build_options", exe_options);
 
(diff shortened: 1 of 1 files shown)


nektro:
unsetting this flag is currently blocked on https://github.com/ziglang/zig/issues/12486

dyxushuai:
But, it has led to the latest release version of zig to not work as expected.

leecannon:
@dyxushuai if by latest release you mean 0.9.1 then this is not the only thing preventing zls building with it, as zls master tracks zig master and there have been many changes to the stdlib in the 3000+ commits to zig since 0.9.1

dyxushuai:
I got it. Zls doesn't support arm mac now and I needs use it with building zig and zls both from source. 😂

squarism:
@dyxushuai Build zls with zig master.

1. Clone zls [as in the wiki](https://github.com/zigtools/zls/wiki/Downloading-and-Building-ZLS)
2. Install master zig (example here with [asdf](https://github.com/asdf-vm/asdf))

```
# install asdf
asdf plugin add zig
asdf install zig master

# cd to zls source and set this repo to build with master
# you'll have a new .tool-versions file (which might be helpful in the future for remembering/updating)
asdf local zig master

# build on m1
zig build -Drelease-safe

# no more error
# error: no member named 'use_stage1' in struct 'std.build.LibExeObjStep'

# put zls in PATH, I use ~/bin
cp zig-out/bin/zls ~/bin

# fire up vim, enjoy inline errors etc
# autocomplete probably needs more setup than this, not sure
```

dyxushuai:
@squarism With a simple solution, I use homebrew and could install master just with `brew install --HEAD zig`.

PR #588, opened 2022-08-20
fix compiler errors when using self-hosted

Techatrix:
With these changes the current zig compiler (0.10.0-dev.3659+e5e6eb983) should be able to compile zls.
Unfortunately we hit out of bound panic's all over the place, which are probably compiler bugs.
Until they are resolved we should stick to stage1. #587


SuperAuguste:
Yep, ended up going with leecannon's approach! Thanks!

Issue #518, opened 2022-07-10
Labels for enums are weird


leecannon:
Seems like the labels for enums include every enum member after (but not including) the one in question.

![label_issue](https://user-images.githubusercontent.com/2286349/178158014-769a5bb4-9ac3-4b0e-9f9e-36fb0ce9b1bb.png)

PR #590, opened 2022-08-20
Better handling for enum field signatures

Techatrix:
Improvment on #518 but doesn't really fix it.
Instead it just displays nothing but it should probably the underlying type of the enum.
```zig
const E = std.builtin.Endian;
const e = E.<cursor here> // should show 'std.builtin.Endian' in completion
```


SuperAuguste:
Nice one!

PR #592, opened 2022-08-20
Fix memory leaks & other memory usage improvements

InKryption:
The aforementioned leak could be observed in the logs when stopping/restarting the server, using a debug build of ZLS.


InKryption:
Nevermind, this doesn't address everything in the way I thought it did. More logs have been revealed to me.

PR #595, opened 2022-08-22
fix: fix memory leaks related to updating config variables

nullptrdevs:
note:
adds an edge case in analysis.makeScopeInternal to prevent
leaking memory when adding duplicate container fields w/ name "other"
but there really should be a check for duplicates

edit:
```
        if (container_field) |_| {
            if (!std.mem.eql(u8, name, "_")) {
                var doc = if (try getDocComments(allocator, tree, decl, .Markdown)) |docs|
                        types.MarkupContent{ .kind = .Markdown, .value = docs }
                    else
                        null;
                var gop_res = try context.enums.getOrPut(allocator, .{
                    .label = name,
                    .kind = .Constant,
                    .insertText = name,
                    .insertTextFormat = .PlainText,
                    .documentation = doc
                });
                if (gop_res.found_existing) {
                    if (doc) |d| allocator.free(d.value);
                }
            }
        }
```

### DIFF
--- a/src/Server.zig
+++ b/src/Server.zig
@@ -2442,7 +2442,7 @@ pub fn processJsonRpc(server: *Server, writer: anytype, json: []const u8) !void 
                     []const u8 => switch (value) {
                         .String => |s| blk: {
                             var nv = try server.allocator.dupe(u8, s);
-                            server.allocator.free(@field(server.config, field.name).?);
+                            if (@field(server.config, field.name)) |prev_val| server.allocator.free(prev_val);
                             break :blk nv;
                         }, // TODO: Allocation model? (same with didChangeConfiguration); imo this isn't *that* bad but still
                         else => @panic("Invalid configuration value"), // TODO: Handle this
--- a/src/Server.zig
+++ b/src/Server.zig
@@ -2440,7 +2440,11 @@ pub fn processJsonRpc(server: *Server, writer: anytype, json: []const u8) !void 
             if (value != .Null) {
                 const new_value: field.field_type = switch (ft) {
                     []const u8 => switch (value) {
-                        .String => |s| try server.allocator.dupe(u8, s), // TODO: Allocation model? (same with didChangeConfiguration); imo this isn't *that* bad but still
+                        .String => |s| blk: {
+                            var nv = try server.allocator.dupe(u8, s);
+                            server.allocator.free(@field(server.config, field.name).?);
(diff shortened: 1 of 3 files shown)


src/analysis.zig
```diff
@@ -2542,7 +2542,7 @@ fn makeInnerScope(allocator: std.mem.Allocator, context: ScopeContext, node_idx:
         };
 
         if (container_field) |_| {
-            if (!std.mem.eql(u8, name, "_")) {
+            if (!std.mem.eql(u8, name, "_") and !std.mem.eql(u8, name, "other")) {
```
SuperAuguste, line 2545: I don't really understand this change :)

src/Server.zig
```diff
@@ -2440,7 +2440,11 @@ pub fn processJsonRpc(server: *Server, writer: anytype, json: []const u8) !void
             if (value != .Null) {
                 const new_value: field.field_type = switch (ft) {
                     []const u8 => switch (value) {
-                        .String => |s| try server.allocator.dupe(u8, s), // TODO: Allocation model? (same with didChangeConfiguration); imo this isn't *that* bad but still
+                        .String => |s| blk: {
+                            var nv = try server.allocator.dupe(u8, s);
+                            server.allocator.free(@field(server.config, field.name).?);
```
SuperAuguste, line 2445: Please add a null capture for the free here - better safe than sorry! :)

src/analysis.zig
leecannon, line 2545: What is the purpose of the "other" case?

leecannon (changes requested):

SuperAuguste (changes requested):

src/Server.zig
nullptrdevs: :1st_place_medal:

src/analysis.zig
nullptrdevs: I'm not too familiar with the codebase but I reasoned the if is to weed out duplicates given the (!std.mem.eql(u8, name, "_"))
and found that a few container_field names are just "other" which cause the overwriting of previous entries and their .documentation field which led to the loss of the .documentation.value pointer value

I've updated the code to catch any dupes

Techatrix (approved):

leecannon (approved):

leecannon (approved):

leecannon:
Thanks for this!

PR #602, opened 2022-08-23
unmanage all the code

### LINKED ISSUE #538: Unmanage all the code!
85% of our ArrayLists and HashMaps are needlessly managed; I'm not going to list them all out here but it's... pretty bad.

Techatrix:
fixes #538


leecannon (approved):
All looks good :)

PR #606, opened 2022-08-25
fix: fix server crash related to parsing incomplete function as a fun…

nullptrdevs:
…ction's parameter

fixes zigtools#567


SuperAuguste (approved):
Sorry for the delay, LGTM!

PR #609, opened 2022-08-26
Improve Test coverage

Techatrix:
implements basic tests for:
- semantic tokens (only one tests but easy to expand)
- inlay hints
- cimport

small step toward #7

### DIFF
--- a/tests/tests.zig
+++ b/tests/tests.zig
@@ -7,6 +7,7 @@ test {
 
     // LSP features
     _ = @import("lsp_features/semantic_tokens.zig");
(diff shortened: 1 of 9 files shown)


tests/tests.zig
```diff
@@ -0,0 +1,14 @@
+test {
```
leecannon, line 1: This can be `comptime {` then the count of tests displayed when running `zig build test` is not increased by one.

leecannon (changes requested):

leecannon (approved):

PR #612, opened 2022-08-29
Fix textDocument/rename

Techatrix:
magic code that makes rename work


leecannon (approved):
Looks good and works fine for me in vscode atleast.

PR #613, opened 2022-08-29
add JSON Schema

Techatrix:


SuperAuguste (approved):
This is great, thanks! This single source of truth will be useful for the VSCode extension, too! :)

PR #615, opened 2022-08-29
Only conditionally ask for global configuration in setup wizard

### LINKED ISSUE #419: Don't ask "Should this configuration be system-wide?" when not run as root
This is not a good developer experience:
image

Instead, it could use `std.fs.accessAbsoluteZ` to test if there are permissions to the system-wide configuration directory and only ask if so.  Since most developers use machines where there is only one user, this is an unnecessary question to ask explicitly

Techatrix:
this should fix #419

### DIFF
--- a/src/setup.zig
+++ b/src/setup.zig
@@ -131,7 +131,7 @@ pub fn wizard(allocator: std.mem.Allocator) !void {
         return;
     }
     var config_path: []const u8 = undefined;
-    if (can_access_global and global_path != null and try askBool("Should this configuration be system-wide?")) {
+    if (can_access_global and try askBool("Should this configuration be system-wide?")) {
         config_path = global_path.?;
     } else {
         if (local_path) |p| {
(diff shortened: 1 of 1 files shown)


src/setup.zig
```diff
@@ -121,18 +121,18 @@ pub fn wizard(allocator: std.mem.Allocator) !void {
...
     }
     var config_path: []const u8 = undefined;
-    if (try askBool("Should this configuration be system-wide?")) {
-        if (global_path) |p| {
-            config_path = p;
-        } else {
-            write("Could not find a global config directory.\n");
-            return;
-        }
+    if (can_access_global and global_path != null and try askBool("Should this configuration be system-wide?")) {
```
leecannon, line 134: As `can_access_global` is set to false above when `global_path` is null, we don't have to check if it is non-null again.

leecannon (changes requested):

leecannon (approved):

PR #628, opened 2022-09-05
fix:  ZLS does not show function signature #608

### LINKED ISSUE #608: ZLS does not show function signature
### Zig Version

0.10.0-dev.3685+dae7aeb33

8cf96fe27cfd235acdf301728f9cce7a7b265ca3

### Steps to Reproduce

Use [Helix](https://helix-editor.com) (`v22.05`),
`build_runner.zig` in the same directory as `zls` and `zls` is in `$PATH`.

ZLS config file:
```json
{
    "zig_exe_path": "/home/noiryuh/.local/bin/zig",
    "enable_snippets": true,
    "enable_unused_variable_warnings": true,
    "enable_import_embedfile_argument_completions": true,
    "warn_style": false,
    "enable_semantic_tokens": true,
    "enable_inlay_hints": true,
    "operator_completions": true,
    "include_at_in_builtins": false,
    "max_detail_length": 1048576
}
```

### Expected Behavior

ZLS should show function signatures for stdlib functions (like for builtin functions).
20220826_14h16m57s_grim

### Actual Behavior

ZLS doesn't show function signatures for stdlib.
20220826_14h09m43s_grim

nullptrdevs:
not all editors make use of item.labelDetails.detail

would close #608

edit:
It'd still be nice to have a server.config option for the rest of the cases.. wink, @Techatrix, wink ;)

### DIFF
--- a/src/Server.zig
+++ b/src/Server.zig
@@ -1557,7 +1557,9 @@ fn formatDetailledLabel(server: *Server, writer: anytype, item: *types.Completio
             return;
         }
 
-        item.detail = item.label;
+        // keep the full fn signature in item.detail as not all editors make use of item.labelDetails.detail
+        // https://github.com/zigtools/zls/issues/608
+        // item.detail = item.label;
         item.labelDetails = .{ .detail = it[s .. e + 1], .description = it[e + 1 ..] };
 
         if (item.kind == .Constant) {


SuperAuguste:
How does this affect editors that *do* support `item.labelDetails.detail`? :)

truemedian:
This should not be a config option, this is exposed as a LSP Client capability, that should be used instead to properly detect what the client can support. see [here](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionClientCapabilities)

SuperAuguste:
Sorry but I don't think this is the right solution. Thanks for your contributions!

Issue #409, opened 2021-10-27
Bug: Untrue assertion hit


MasterQ32:
I'm hitting unreachable in some cases in my source. I sadly cannot strip it down as it only happens once a day or so and isn't reproducible afterwards.

<details>
  <summary>Stack Trace</summary>
<pre><code>
thread 8321 panic: reached unreachable code
/home/felix/software/zig-linux-x86_64-0.9.0-dev.1343+75cecef63/lib/std/debug.zig:226:14: 0x25f33b in std.debug.assert (zls)
    if (!ok) unreachable; // assertion failure
             ^
/home/felix/projects/forks/zls/src/offsets.zig:180:21: 0x343f01 in offsets.tokenLocation (zls)
    std.debug.assert(token.tag == tag);
                    ^
/home/felix/projects/forks/zls/src/offsets.zig:143:36: 0x343a8c in offsets.tokenLength (zls)
    const token_loc = tokenLocation(tree, token);
                                   ^
/home/felix/projects/forks/zls/src/semantic_tokens.zig:128:43: 0x3423ac in semantic_tokens.Builder.handleToken (zls)
        const length = offsets.tokenLength(tree, tok, self.encoding);
                                          ^
/home/felix/projects/forks/zls/src/semantic_tokens.zig:82:32: 0x3ae87b in semantic_tokens.Builder.add (zls)
                try handleToken(self, i);
                               ^
/home/felix/projects/forks/zls/src/semantic_tokens.zig:185:24: 0x31e86b in semantic_tokens.writeNodeTokens (zls)
        try builder.add(ti, tok_type, tok_mod);
                       ^
/home/felix/projects/forks/zls/src/semantic_tokens.zig:315:31: 0x31dfc9 in semantic_tokens.writeNodeTokens (zls)
                    try await @asyncCall(child_frame, {}, writeNodeTokens, .{ builder, arena, store, child });
                              ^
/home/felix/projects/forks/zls/src/semantic_tokens.zig:472:27: 0x32c643 in semantic_tokens.writeNodeTokens (zls)
                try await @asyncCall(child_frame, {}, writeNodeTokens, .{ builder, arena, store, node_data[node].rhs });
                          ^
/home/felix/projects/forks/zls/src/semantic_tokens.zig:393:31: 0x31feba in semantic_tokens.writeNodeTokens (zls)
                    try await @asyncCall(child_frame, {}, writeNodeTokens, .{ builder, arena, store, child });
                              ^
[...]

Issue #381, opened 2021-08-26
Inserting period inside any token crashes zls


Vexu:
To reproduce insert a period anywhere inside `foo` in
```zig
test {
    foo;
}
```

<details>
<summary>StackTrace</summary>
<pre><code>lib/zig/std/debug.zig:226:14: 0x25cbab in std.debug.assert (zls)
    if (!ok) unreachable; // assertion failure
             ^
/usr/lib/zig/std/zig/ast.zig:109:15: 0x30fc0d in std.zig.ast.Tree.tokenSlice (zls)
        assert(token.tag == token_tag);
              ^
/home/vexu/Documents/zig/zls/src/semantic_tokens.zig:898:44: 0x326888 in semantic_tokens.writeNodeTokens (zls)
            const rhs_str = tree.tokenSlice(data.rhs);
                                           ^
/home/vexu/Documents/zig/zls/src/semantic_tokens.zig:353:31: 0x31b689 in semantic_tokens.writeNodeTokens (zls)
                    try await @asyncCall(child_frame, {}, writeNodeTokens, .{ builder, arena, store, child });
                              ^
/home/vexu/Documents/zig/zls/src/semantic_tokens.zig:835:23: 0x32c222 in semantic_tokens.writeNodeTokens (zls)
            try await @asyncCall(child_frame, {}, writeNodeTokens, .{ builder, arena, store, node_data[node].rhs });
                      ^
/home/vexu/Documents/zig/zls/src/semantic_tokens.zig:1071:24: 0x2c4918 in semantic_tokens.writeAllSemanticTokens (zls)
        writeNodeTokens(&builder, arena, store, child) catch |err| switch (err) {
                       ^
/home/vexu/Documents/zig/zls/src/main.zig:1429:71: 0x2b2ba9 in semanticTokensFullHandler (zls)
        const token_array = try semantic_tokens.writeAllSemanticTokens(arena, &document_store, handle, offset_encoding);
                                                                      ^
/home/vexu/Documents/zig/zls/src/main.zig:1719:53: 0x2932b8 in processJsonRpc (zls)
                    done = extractErr(method_info[2](arena, id, request_obj, config));
                                                    ^
/home/vexu/Documents/zig/zls/src/main.zig:1962:27: 0x286404 in main (zls)
        try processJsonRpc(&arena, &json_parser, buf, config);
                          ^
/usr/lib/zig/std/start.zig:510:37: 0x27ceba in std.start.callMain (zls)
[...]

PR #629, opened 2022-09-05
fix: "f..o.o;" crashes the server

nullptrdevs:
this is a fix for #381 and a hack for #409
(related, get triggered by "f..o.o;")

merge or use as an inspiration to fix these

PS. only covers .field_access
ideally all calls to tree.tokenSlice(token) should be reworked to use ast.tokenSlice(tree, token)

Regards


SuperAuguste (approved):
Simple but elegant.

PR #632, opened 2022-09-06
use client capabilities instead of config for inlayHint and semanticTokens

truemedian:
Pretty self explanatory, the LSP gives these config options to us every time we connect, we should use them instead of relying on the user to change the config themselves if they use an editor that does support these features.


SuperAuguste:
Thoughts? I think this is good; all editors *should* support configuring client capabilities *right?*

truemedian:
At the very least client capabilities should reflect what the client can even show or process.

SuperAuguste:
Right, but people may not want inlay hints / semantic tokens. It's not just about technical capability. Not sure how other editors do it, but VSCode settings for semantic tokens do properly configure client capabilities; not sure about inlay hints.

SuperAuguste:
tbh, I don't like this change; sorry

PR #635, opened 2022-09-07
Place build_runner.zig in cache directory

### LINKED ISSUE #634: Don't drop build_runner.zig next to executable
### Zig Version

0.10.0-dev.3685+dae7aeb33

38210d3e9183a77dfe10e0d0f4a79d2999ce3cf7

### Steps to Reproduce

Start `zls` in a project. If the `zls` binary is installed in a read-only directory, such as `/usr/bin` this will crash with a permission denied error.

### Expected Behavior

`zls` doesn't try to drop any files next to its executable, and instead uses some alternative path such as `zig-cache` or `/tmp`.

### Actual Behavior

`zls` tries to create a `build_runner.zig` file next to its executable. This often fails, for example when zls is installed in `/usr/bin` (tries to create `/usr/bin/build_runner.zig`), and needlessly pollutes the filesystem.

This was introduced in `38210d3e9183a77dfe10e0d0f4a79d2999ce3cf7`.

Once we agree on a solution, I'd very much like to implement it. Cheers!

Techatrix:
fixes #634

### DIFF
--- a/src/Config.zig
+++ b/src/Config.zig
@@ -202,10 +202,13 @@ pub fn configChanged(config: *Config, allocator: std.mem.Allocator, builtin_crea
         config.builtin_path = try std.fs.path.join(allocator, &.{ builtin_creation_dir.?, "builtin.zig" });
     }
 
+    const cache_dir_path = (try known_folders.getPath(allocator, .cache)) orelse {
+        logger.warn("Known-folders could not fetch the cache path", .{});
+        return;
+    };
+
     if (null == config.build_runner_path) {
-        var exe_dir_bytes: [std.fs.MAX_PATH_BYTES]u8 = undefined;
-        const exe_dir_path = try std.fs.selfExeDirPath(&exe_dir_bytes);
-        config.build_runner_path = try std.fs.path.resolve(allocator, &[_][]const u8{ exe_dir_path, "build_runner.zig" });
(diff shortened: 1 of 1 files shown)


src/Config.zig
```diff
@@ -202,10 +202,13 @@ pub fn configChanged(config: *Config, allocator: std.mem.Allocator, builtin_crea
         config.builtin_path = try std.fs.path.join(allocator, &.{ builtin_creation_dir.?, "builtin.zig" });
     }
 
+    const cache_dir_path = (try known_folders.getPath(allocator, .cache)) orelse {
```
LordMZTE, line 205: You should append `zls` to this. It's bad practice to drop stuff straight in `~/.cache`.

SuperAuguste (approved):
LGTM!

src/Config.zig
```diff
@@ -214,10 +217,6 @@ pub fn configChanged(config: *Config, allocator: std.mem.Allocator, builtin_crea
     }
 
     if (null == config.global_cache_path) {
```
LordMZTE, line 219: Might be best to just move this block above the one creating the build runner file instead of using `cache_dir_path`

LordMZTE (changes requested):
Sniped my PR :P

PR #644, opened 2022-09-11
Pass allocator to NativeTargetInfo.detect

leroycep:
All allocations are cleaned up before returning, and it is planned that the allocator requirement will be removed at some point.

For more details see: https://github.com/ziglang/zig/commit/9e070b653c89a9216f9dd9f78ed7c78c11460ac7


leroycep:
Looks like this is already changed back: https://github.com/ziglang/zig/commit/3ee01c14ee7ba42b484f15daeacb67da90a81c9e

PR #647, opened 2022-09-11
Only globally log to stderr

Techatrix:
Previously zls tries to log messages in `Server.zig` using `window/logMessage`. This PR gets rid of this behavior and instead simply writes to stderr like in every other file.

### DIFF
--- a/src/Server.zig
+++ b/src/Server.zig
@@ -20,6 +20,8 @@ const uri_utils = @import("uri.zig");
 const data = @import("data/data.zig");
 const diff = @import("diff.zig");
 
+const log = std.log.scoped(.server);
+
 // Server fields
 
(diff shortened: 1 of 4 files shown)


SuperAuguste:
Makes sense, not sure why I even started using logMessage in the first place to be completely honest. Quick side notes: are there any situations where logMessage would even make sense? Maybe when using LSP over a network or smth like that?

SuperAuguste (approved):

SuperAuguste:
Sure, just something to keep in mind if network support's something we were to add in the future :)
(Also you could just pipe the stdin/stdout from zls over a socket; not much rewriting is required on our side, though why someone would do this remains nebulous - maybe remote dev containers?)

PR #649, opened 2022-09-13
Improve logging output

Techatrix:
the logs now looks more like this
```
info : (config): Using zig executable /opt/zig/zig
info : (config): Using zig lib path '/opt/zig/lib'
debug: (server): Dynamically registering method 'workspace/didChangeConfiguration'
info : (server): zls initialized
info : (server): Server.ClientCapabilities{ .supports_snippets = true, .supports_semantic_tokens = true, .supports_inlay_hints = true, .hover_supports_md = true, .completion_doc_supports_md = true, .label_details_support = true, .supports_configuration = true }
info : (server): Using offset encoding: utf16
debug: (server): Took 1ms to process method initialize
debug: (server): Took 0ms to process method initialized
debug: (store ): Document is a build file, extracting packages...
info : (store ): builtin config not found, falling back to default: file:///home/techatrix/.config/builtin.zig
```

### DIFF
--- a/src/DocumentStore.zig
+++ b/src/DocumentStore.zig
@@ -3,7 +3,7 @@ const types = @import("types.zig");
 const URI = @import("uri.zig");
 const analysis = @import("analysis.zig");
 const offsets = @import("offsets.zig");
-const log = std.log.scoped(.doc_store);
+const log = std.log.scoped(.store);
 const Ast = std.zig.Ast;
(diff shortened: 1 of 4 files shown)


SuperAuguste (approved):
Doesn't look great but it's not like typical users are looking at zls' logger output much anyways; great for readability though, thanks Techatrix!!

PR #653, opened 2022-09-14
References improvements

### LINKED ISSUE #654: Rename does not track errdefers
Defers are handled [here](https://github.com/zigtools/zls/blob/master/src/references.zig#L153) but errdefers are not.

Techatrix:
This PR merges the code responsible for `textDocument/rename`, `textDocument/references` and `textDocument/documentHighlight` because they are very similar. And `references.zig` receives a large refactor. fixes #654 

These features are still very buggy and i will address that in later PRs.


Techatrix:
This is now fixed in my refactor, but many bugs remain...

SuperAuguste (approved):
👏👏👏

PR #655, opened 2022-09-15
ability to highlight global var declarations

### LINKED ISSUE #540: Diagnostic that highlights global variables
Global variables are generally shunned in Zig land, so adding an option that shows global variable diagnostics might be cool. This would be especially useful for me right now as I'm refactoring zls' globals.

nullptrdevs:
this probably looks rough but does highlight everything declared with "var" at the root level

Closes #540


Techatrix (changes requested):
nice!
But you should also update the README.

SuperAuguste:
Could you please force push on this branch to remove all the extraneous commits?

SuperAuguste:
I think we can just squash and merge, no worries :)

SuperAuguste (approved):

PR #656, opened 2022-09-15
Add target to Env struct

jcmoyer:
Fixes an issue where zls couldn't parse the output from `zig env` causing it to repeatedly print

```
debug: (store ): Cannot resolve std library import, path is null.
```

and fail to autocomplete.

See: https://github.com/ziglang/zig/pull/11741

### DIFF
--- a/src/Config.zig
+++ b/src/Config.zig
@@ -147,6 +147,7 @@ pub fn configChanged(config: *Config, allocator: std.mem.Allocator, builtin_crea
                             std_dir: []const u8,
                             global_cache_dir: []const u8,
                             version: []const u8,
+                            target: ?[]const u8 = null,
                         };
 
                         var json_env = std.json.parse(


SuperAuguste:
Website CI is broken; I'll merge this when it's fixed :)

leecannon (changes requested):
This will need to be `target: ?[]const u8 = null,` to work with older zig versions, including CI atm.

nektro:
set `ignore_unknown_fields = true` in our `std.json.ParseOptions` instead

jcmoyer:
I'm not sure which behavior is more preferrable (add `target` field or use `ignore_unknown_fields`). I only became aware of the `zig env` change explicitly because `ignore_unknown_fields` wasn't being used.

leecannon (approved):

PR #657, opened 2022-09-16
parse 'zig env' error, master zig has changed json struct

maxzhao:


nektro:
set `ignore_unknown_fields = true` in ParseOptions instead

SuperAuguste:
Duplicate of #656

PR #658, opened 2022-09-17
fix for zig master

leecannon:
`TypeInfo` is deprecated `pub const TypeInfo = @compileError("deprecated; use Type");`


SuperAuguste (approved):
Oddly enough, despite the website build not being up to date, CI passes (we probably don't use that function then). Thanks!

PR #660, opened 2022-09-17
Fix missing includes by passing -lc to translate-c

TheSpydog:
I was trying out the new cImport feature in ZLS and noticed that it was unable to offer completion for the SDL2 library's `SDL.h`. Looking at the Debug window of VSCode, I noticed it was erroring out because it could not find one of the included files (`<process.h>`, part of the Windows C standard library). It appears that translate-c does not know how to find libc header files unless `-lc` is explicitly used.

By passing `-lc` to the translate-c invocation, SDL now has full autocompletion via ZLS.

### DIFF
--- a/src/translate_c.zig
+++ b/src/translate_c.zig
@@ -134,6 +134,7 @@ pub fn translate(allocator: std.mem.Allocator, config: Config, include_dirs: []c
         config.zig_lib_path.?,
         "--cache-dir",
         config.global_cache_path.?,
+        "-lc",
     };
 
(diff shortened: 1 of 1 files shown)


SuperAuguste:
One half of me says that there should probably be a build.zig check to see if `c` is used in any `build.zig` build steps but the other half says it probably won't hurt to just universally add `-lc`. Thoughts @Techatrix?

Techatrix:
I don't think it would cause any problems if we always add it. "keep it simple stupid". We can always add checks later...

SuperAuguste (approved):

PR #669, opened 2022-09-21
Avoid config copy for document store

Techatrix:
`DocumentStore` is keeping a copy of the `Config` struct. This PR reverts back to pointing into `Server.config`
https://github.com/zigtools/zls/blob/0fa788b7275891f5bd924ca983c8ff397de23cc7/src/DocumentStore.zig#L85-L86

### DIFF
--- a/src/DocumentStore.zig
+++ b/src/DocumentStore.zig
@@ -82,8 +82,7 @@ allocator: std.mem.Allocator,
 handles: UriToHandleMap = .{},
 build_files: BuildFileList = .{},
 
-// TODO use pointer back to Server.config
-config: Config,
+config: *Config,
 std_uri: ?[]const u8,
 // TODO make this configurable
 // We can't figure it out ourselves since we don't know what arguments
@@ -95,7 +94,7 @@ zig_global_cache_root: []const u8 = "ZLS_DONT_CARE",
 
 pub fn init(
     allocator: std.mem.Allocator,
-    config: Config,
+    config: *Config,
 ) !DocumentStore {
     return DocumentStore{
         .allocator = allocator,
@@ -671,7 +670,7 @@ fn translate(self: *DocumentStore, handle: *Handle, source: []const u8) error{Ou
 
     const maybe_result = try translate_c.translate(
         self.allocator,
-        self.config,
+        self.config.*,
         include_dirs,
         source,
     );
--- a/src/Server.zig
+++ b/src/Server.zig
@@ -23,7 +23,7 @@ const log = std.log.scoped(.server);
 
 // Server fields
 
-config: Config,
+config: *Config,
 allocator: std.mem.Allocator = undefined,
 arena: std.heap.ArenaAllocator = undefined,
 document_store: DocumentStore = undefined,
@@ -1164,7 +1164,7 @@ fn completeBuiltin(server: *Server, writer: anytype, id: types.RequestId) !void 
 
     if (server.builtin_completions == null) {
         server.builtin_completions = try std.ArrayListUnmanaged(types.CompletionItem).initCapacity(server.allocator, data.builtins.len);
-        try populateBuiltinCompletions(&server.builtin_completions.?, server.config);
+        try populateBuiltinCompletions(&server.builtin_completions.?, server.config.*);
     }
 
     try send(writer, server.arena.allocator(), types.Response{
@@ -2054,7 +2054,7 @@ fn didChangeConfigurationHandler(server: *Server, writer: anytype, id: types.Req
             }
         }
 
-        try server.configChanged(null);
+        try server.config.configChanged(server.allocator, null);
     } else if (server.client_capabilities.supports_configuration)
         try server.requestConfiguration(writer);
 }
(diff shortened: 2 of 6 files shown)


src/DocumentStore.zig
```diff
@@ -671,7 +670,7 @@ fn translate(self: *DocumentStore, handle: *Handle, source: []const u8) error{Ou
 
     const maybe_result = try translate_c.translate(
         self.allocator,
-        self.config,
+        self.config.*,
```
leecannon, line 673: Is there any reason why this and other functions take `Config` by value instead of const pointer? Or is that just a leftover of it not being stored as a pointer before.

Techatrix: Please correct me if I am wrong.
If i understand the [Pass-by-value Parameters](https://ziglang.org/documentation/master/#Pass-by-value-Parameters) section correctly. Zig automatically converts parameters that are passed by value to pass by reference if deemed more efficient. Passing a parameter by const pointer would therefore only be necessary if you want to force that optimization which i deem unnecessary.

leecannon: I think that as we are de-referencing the pointer we are copying the config onto the stack, then zig might choose to pass a reference to the _stack_ copy of the config.

Techatrix: that could be the case, but i am unsure if a `.*` unwrap always implies a copy because of code like this:
```zig
var foo: SomeStruct = .{};
var foo_ptr = &foo;
foo_ptr.* = .{}; // foo_ptr.* creates a reference to foo and not a copy of foo
```

Techatrix: I have created this small code snippet. [Compiler Explorer](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:zig,selection:(endColumn:2,endLineNumber:16,positionColumn:1,positionLineNumber:1,selectionStartColumn:2,selectionStartLineNumber:16,startColumn:1,startLineNumber:1),source:'const+std+%3D+@import(%22std%22)%3B%0A%0Aconst+S+%3D+struct+%7B%0A++++data:+%5B1024%5Du8+%3D+std.mem.zeroes(%5B1024%5Du8),%0A%7D%3B%0A%0Afn+bar(s:+S)+void+%7B%0A++++std.mem.doNotOptimizeAway(s)%3B%0A%7D%0A%0Apub+fn+main()+void+%7B%0A++++var+foo:+S+%3D+.%7B%7D%3B%0A++++var+foo_ptr:+*S+%3D+%26foo%3B%0A++++std.mem.doNotOptimizeAway(foo_ptr)%3B%0A++++bar(foo_ptr.*)%3B%0A%7D'),l:'5',n:'0',o:'Zig+source+%231',t:'0')),k:45.642633228840126,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:ztrunk,filters:(b:'0',binary:'1',commentOnly:'0',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:zig,libs:!(),options:'',selection:(endColumn:28,endLineNumber:913,positionColumn:28,positionLineNumber:913,selectionStartColumn:28,selectionStartLineNumber:913,startColumn:28,startLineNumber:913),source:1,tree:'1'),l:'5',n:'0',o:'zig+trunk+(Zig,+Editor+%231,+Compiler+%231)',t:'0'),(h:output,i:(compilerName:'zig+trunk',editorid:1,fontScale:14,fontUsePx:'0',j:1,wrap:'1'),l:'5',n:'0',o:'Output+of+zig+trunk+(Compiler+%231)',t:'0')),k:54.357366771159874,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4)
```zig
const std = @import("std");

const S = struct {
    data: [1024]u8 = std.mem.zeroes([1024]u8),
};

fn bar(s: S) void {
    std.mem.doNotOptimizeAway(s);
}

pub fn main() void {
    var foo: S = .{};
    var foo_ptr: *S = &foo;
    std.mem.doNotOptimizeAway(foo_ptr);
    bar(foo_ptr.*);
}
```
Here is the generated x86 assembly in a debug build.
Im not an expert in reading assembly but i don't see an extra copy.
```
example.main:
    push    rbp
    mov     rbp, rsp
    sub     rsp, 1040                        # allocate stack memory
    lea     rdi, [rbp - 1032]
    movabs  rsi, offset .L__unnamed_2        # pointer into constant memory filled with zeroes
    mov     edx, 1024                               
    call    memcpy@PLT                       # memcpy zeroes into the stack memory
    lea     rax, [rbp - 1032]                # evaluate &foo
    mov     qword ptr [rbp - 8], rax         # foo_ptr = &foo
    mov     rdi, qword ptr [rbp - 8]
    call    mem.doNotOptimizeAway__anon_3232 # optimization boundry (just in case)
    mov     rdi, qword ptr [rbp - 8]         # setup bar's first parameter which points into stack memory (no copy)
                                             # a copy would require allocating more stack memory but that's not the case
                                             # and we only allocated 1040 bytes which is only enough for one copy
    call    example.bar                      # call the bar function
    add     rsp, 1040                        # deallocate stack memory
    pop     rbp
    ret
```

leecannon: Seems like your right :)

PR #672, opened 2022-09-24
wip: more accurate symbol kinds in analysis.getDocumentSymbols

haolian9:
hi, i'm trying to make analysis.getDocumentSymbols recognize more Symbol.Kind, do you think this is worth it?
and what must be done to make this pr mergeable?

todo:
* [x] `[pub] fn Function() type {};`
* [x] `[pub] var|const Type = struct|enum|union {}`
* [ ] `const Foo = @import("Foo.zig");`
* [ ] `pub var|const Type = std.ArrayList(u8);`

i have no idea to deal with the last two cases at now.

### DIFF
--- a/src/analysis.zig
+++ b/src/analysis.zig
@@ -1837,7 +1837,7 @@ fn resolveNodeKind(tree: Ast, node_idx: Ast.Node.Index, tag: std.zig.Ast.Node.Ta
         // [pub] var|const Type = struct|enum|union {}
         .simple_var_decl, .aligned_var_decl, .global_var_decl => {
             switch (token_tags[node_maintokens[node_idx]]) {
-                .keyword_var => .Variable,
+                .keyword_var => return .Variable,
                 .keyword_const => {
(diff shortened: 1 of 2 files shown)


haolian9:
i saw no `test {}` in analysis.zig, so i have to put tests inside `/tests/`, this make it harder to test a private function.

Techatrix:
The code you've written kind of reminded be of this section:
https://github.com/zigtools/zls/blob/0d848a71ecffe67ed930cc2e1629acc0dfec6055/src/Server.zig#L756-L798
Because they are both trying to solve the question: "What is the type of this symbol"?

The `resolveTypeOfNode` function may help you solve the other two cases
https://github.com/zigtools/zls/blob/0d848a71ecffe67ed930cc2e1629acc0dfec6055/src/analysis.zig#L1085
the return value's sub field `is_type_val` should help distinguish between variables and types.

haolian9:
thanks for your advices! but lately i found i have no enough time to learn the whole source code, so i'm closing this pr for now

PR #679, opened 2022-09-27
Adds a 'format to camelCase' codeaction for function names

NuclearPhone:
Does as the title says, this PR adds a code action that reformats non-camelCase function names to be camelCase.
 ___
camelcase1
camelcase2

### DIFF
--- a/src/code_actions.zig
+++ b/src/code_actions.zig
@@ -74,26 +74,21 @@ pub const Builder = struct {
 fn handleNonCamelcaseFunction(builder: *Builder, actions: *std.ArrayListUnmanaged(types.CodeAction), loc: offsets.Loc) !void {
     const identifier_name = offsets.locToSlice(builder.text(), loc);
 
-    const tree = builder.handle.tree;
-
-    const decl = (try analysis.lookupSymbolGlobal(
-        builder.document_store,
-        builder.arena,
-        builder.handle,
-        identifier_name,
-        loc.start,
-    )) orelse return;
-
-    const name_token_idx = decl.nameToken();
-    const name_loc = offsets.tokenToLoc(tree, name_token_idx);
+    // const decl = (try analysis.lookupSymbolGlobal(
+    //     builder.document_store,
+    //     builder.arena,
+    //     builder.handle,
+    //     identifier_name,
+    //     loc.start,
+    // )) orelse return;
 
     const new_text = try createCamelcaseText(builder.arena.allocator(), identifier_name);
 
     const action1 = types.CodeAction{
         .title = "make function name camelCase",
-        .kind = .QuickFix,
+        .kind = .SourceFixAll,
         .isPreferred = true,
-        .edit = try builder.createWorkspaceEdit(&.{builder.createTextEditLoc(name_loc, new_text)}),
+        .edit = try builder.createWorkspaceEdit(&.{builder.createTextEditLoc(loc, new_text)}),
     };
 
     try actions.append(builder.arena.allocator(), action1);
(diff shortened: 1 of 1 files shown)


src/code_actions.zig
```diff
@@ -159,7 +188,7 @@ fn handleUnusedVariableOrConstant(builder: *Builder, actions: *std.ArrayListUnma
 
     try actions.append(builder.arena.allocator(), .{
         .title = "discard value",
-        .kind = .SourceFixAll,
+        .kind = .QuickFix,
```
Techatrix, line 191: this should be keep at SourceFixAll

src/code_actions.zig
```diff
@@ -284,6 +341,10 @@ const DiagnosticKind = union(enum) {
             return DiagnosticKind{
                 .omit_discard = parseEnum(DiscardCat, msg["discard of ".len..]) orelse return null,
             };
+        } else if (std.mem.startsWith(u8, msg, "Functions should")) {
```
Techatrix, line 348: i thinks its better if you check for the entire message `Functions should be camelCase`

src/code_actions.zig
```diff
@@ -70,6 +71,28 @@ pub const Builder = struct {
...
+    const decl = (try analysis.lookupSymbolGlobal(builder.document_store, builder.arena, builder.handle, identifier_name, loc.start)) orelse return;
+
+    const name_token_idx = decl.nameToken();
+    const name_loc = offsets.tokenToLoc(tree, name_token_idx);
+
+    const new_text = try createCamelcaseText(builder.arena.allocator(), identifier_name);
+
+    const action1 = types.CodeAction{
+        .title = "make function name camelCase",
+        .kind = .SourceFixAll,
```
Techatrix: I am unsure if this should be part of autofix (indicated by `SourceFixAll`). Can you set it to `QuickFix` while i think about it being included in autofix.

src/code_actions.zig
```diff
@@ -265,18 +259,22 @@ fn handlePointlessDiscard(builder: *Builder, actions: *std.ArrayListUnmanaged(ty
 // attempts to converts a slice of text into camelcase 'FUNCTION_NAME' -> 'functionName'
 fn createCamelcaseText(allocator: std.mem.Allocator, identifier: []const u8) ![]const u8 {
     var num_separators: usize = 0;
-    
+
+    for (identifier) |c| {
+        if (c == '_') num_separators += 1;
```
Techatrix, line 270: you can refactor this into `const num_separators = std.mem.count(u8, identifier, "_")`

src/code_actions.zig
```diff
@@ -70,6 +71,34 @@ pub const Builder = struct {
...
+        builder.document_store,
+        builder.arena,
+        builder.handle,
+        identifier_name,
+        loc.start,
+    )) orelse return;
+
+    const name_token_idx = decl.nameToken();
+    const name_loc = offsets.tokenToLoc(tree, name_token_idx);
+
```
Techatrix, line 89: You don't need this code because the identifier location is in `loc`.

src/code_actions.zig
```diff
@@ -233,6 +262,37 @@ fn handlePointlessDiscard(builder: *Builder, actions: *std.ArrayListUnmanaged(ty
...
+        if (c == '_') {
+            const c2 = identifier[idx + 1];
+            new_text.appendAssumeCapacity(std.ascii.toUpper(c2));
+            idx += 2;
+        } else {
+            new_text.appendAssumeCapacity(std.ascii.toLower(c));
+            idx += 1;
+        }
+    }
```
Techatrix, line 291: this causes a index out of bound error when you have an identifier which ends with an underscore like
```zig
fn foo_bar_baz_() void {}
```

Techatrix (changes requested):
Thanks for you PR! It's great to see more code actions getting integrated. There are "a few" issues though.
If you are interested you can also add code actions for `Type functions should be PascalCase` or `A ./ is not needed in imports`.

NuclearPhone:
Thanks for the feedback; I'll get to fixing my error-ridden code. In the future i'll pay more attention to proof-reading.

src/code_actions.zig
Techatrix: I think your latest commit touched the wrong place. 
Set this location https://github.com/NuclearPhone/zls/blob/39ea5869a377eeec71fe333f2ee264fac7cf2bbb/src/code_actions.zig#L85 to `.QuickFix`.
And set this location https://github.com/NuclearPhone/zls/blob/39ea5869a377eeec71fe333f2ee264fac7cf2bbb/src/code_actions.zig#L176 to `.SourceFixAll`.

Techatrix (approved):
nice

PR #682, opened 2022-09-28
Self hosted fixes

Techatrix:
This is the first step towards using self hosted as the compiler backend. In my latest testing i didn't experience any crashes during semantic token generation or inlay hints which was previously the case. So i don't think there is much stopping zls from making the move to self-hosted. Of course more testing is needed before i can consider it stable enough.
NOTE: this PR only includes fixes and doesn't switch to self hosted.


SuperAuguste:
Conflicts need to be resolved :(

SuperAuguste (approved):

PR #683, opened 2022-09-28
Allow user defined options to be passed to build_runner

jcmoyer:
### Problem

Build options (`-Dname=value`, `-Dflag`) can change the dependency graph resulting from running `build.zig`. If `build.zig` is incorrectly configured, zls will either not be able to extract dependency information[1] or it may do so incorrectly[2].

[1] This can happen when the build script is designed to exit if a required option is not present. This will cause `build_runner` to fail. [Example from zig](https://github.com/ziglang/zig/blob/f3a1b5c481646ee35813cbe8bb2b0a3979df3ab8/build.zig#L183-L188)
[2] This can happen when options have default values that differ from those passed during actual builds. Since zls currently does not pass any user options to `build.zig`, it will always use defaults. [Example from zig](https://github.com/ziglang/zig/blob/f3a1b5c481646ee35813cbe8bb2b0a3979df3ab8/build.zig#L51-L62)

### Motivation

I have projects that use a mixture of C, C++, and zig libraries. I prefer to keep these out of tree since there are some especially large libraries like opencv and zigwin32, and they are shared between multiple projects. To accomplish this, library paths are passed on the command line via `Builder.option`:

```
> zig build -Dopencv-root-dir=D:\Scratch\opencv-4.6.0-zigbuild\install -Dzigwin32-root-dir=D:\Scratch\zigwin32 -Dautomate-root-dir=D:\Dev\Projects\ZigAutomate
```

```zig
    // somewhere in a build.zig
    const automate_root_dir = b.option([]const u8, "automate-root-dir", "root directory for automate");
    if (automate_root_dir == null) {
        std.debug.print("missing required option: -Dautomate-root-dir=[string]\n\n", .{});
        std.process.exit(1);
    }
    const automate_root_src = b.pathJoin(&[_][]const u8{ automate_root_dir.?, "src/lib.zig" });
    const automate = Pkg{
        .name = "auto",
        .source = .{ .path = automate _root_src },
        .dependencies = &[_]Pkg{ win32 },
    };
    // similar things for other packages
```

The project I'm doing this in is not getting autocomplete from zls because it does not pass these options through `build_runner`, which then fails because I use the idiom in [1].

### Implementation

This PR adds a `build_options` field to `BuildAssociatedConfig`. These can be set in a project-local `zls.build.json`. Example:

```json
{
    "build_options": [
        {
            "name": "automate-root-dir",
            "value": "D:\\Dev\\Projects\\ZigAutomate"
        },
        {
            "name": "opencv-root-dir",
            "value": "D:\\Scratch\\opencv-4.6.0-zigbuild\\install"
        },
        {
            "name": "zigwin32-root-dir",
            "value": "D:\\Scratch\\zigwin32"
        }
    ]
}
```

Options with a `name` and `value` are passed to `build.zig` as `-Dname=value`. If `value` is omitted, the option is treated as a flag and passed as `-Dname`.

### DIFF
--- a/src/DocumentStore.zig
+++ b/src/DocumentStore.zig
@@ -178,7 +178,7 @@ fn loadBuildConfiguration(context: LoadBuildConfigContext) !void {
     defer if (context.build_file_path == null) allocator.free(build_file_path);
     const directory_path = build_file_path[0 .. build_file_path.len - "build.zig".len];
 
-    const standard_args = &[_][]const u8{
+    const standard_args = [_][]const u8{
         zig_exe_path,
         "run",
         build_runner_path,
@@ -198,7 +198,7 @@ fn loadBuildConfiguration(context: LoadBuildConfigContext) !void {
     var args = try arena_allocator.alloc([]const u8, standard_args.len + if (build_file.build_options) |opts| opts.len else 0);
     defer arena_allocator.free(args);
 
-    args[0..standard_args.len].* = standard_args.*;
+    args[0..standard_args.len].* = standard_args;
     if (build_file.build_options) |opts| {
         for (opts) |opt, i| {
             args[standard_args.len + i] = try opt.formatParam(arena_allocator);
--- a/src/DocumentStore.zig
+++ b/src/DocumentStore.zig
@@ -118,7 +118,7 @@ fn loadBuildAssociatedConfiguration(allocator: std.mem.Allocator, build_file: *B
     const directory_path = build_file_path[0 .. build_file_path.len - "build.zig".len];
 
     const options = std.json.ParseOptions{ .allocator = allocator };
-    const build_associated_config = blk: {
+    var build_associated_config = blk: {
         const config_file_path = try std.fs.path.join(allocator, &[_][]const u8{ directory_path, "zls.build.json" });
         defer allocator.free(config_file_path);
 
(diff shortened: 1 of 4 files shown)


src/DocumentStore.zig
```diff
@@ -133,6 +142,21 @@ fn loadBuildAssociatedConfiguration(allocator: std.mem.Allocator, build_file: *B
...
+            for (build_opts.items) |*opt| {
+                opt.deinit(allocator);
+            }
+            build_opts.deinit(allocator);
+        }
+        for (opts) |opt| {
+            try build_opts.append(allocator, try opt.dupe(allocator));
+        }
+        build_file.build_options = build_opts.toOwnedSlice(allocator);
+    }
```
alexnask, line 159: You can probably do something like  
```zig
build_file.build_options = build_associated_config.build_options;
build_associated_config.build_options = null;
```  
To 'steal' the slice from the parsed JSON since the allocators are the same.  
In any case using an ArrayList is overkill since you can also allocate a slice of the desired size with Allocator.alloc

src/DocumentStore.zig
```diff
@@ -156,26 +180,45 @@ fn loadBuildConfiguration(context: LoadBuildConfigContext) !void {
...
-        build_file_path,
-        "--pkg-end",
-        "--",
-        zig_exe_path,
-        directory_path,
-        context.cache_root,
-        context.global_cache_root,
-    };
+    // enough capacity for arguments always passed to the build runner plus a couple build options
+    var arglist = try std.ArrayListUnmanaged([]const u8).initCapacity(arena_allocator, 32);
```
alexnask, line 192: You can calculate the exact amount of arguments that are needed and allocate them directly  
```zig
const args = try allocator.alloc([]const u8, 14 + if (build_file.build_options) |opts| opts.len else 0);
defer allocator.free(args);
args[0..14].* = [_][]const u8 {
    // ...
};
if (build_file.build_options) |opts| {
    for (opts) |opt, i| args[14 + i] = opt;
}
```

jcmoyer:
Good catches, I've made the requested changes.

PR #686, opened 2022-09-28
handle `OptionsStep` in build_runner

leecannon:
This allows generated files from `OptionsStep` to be resolved.

This is dependant on https://github.com/ziglang/zig/pull/13003 I will leave it as a draft until it is merged.

I'm not 100% happy with this due to:
- We can't know how costly the dependencies of the step might be so i've decided to only support `OptionsStep`s that have no dependencies.
Thankfully this is the most common situation, I don't think i've ever seen an `OptionsStep` have any dependencies.
- It's extremely common for the pacakge to be called `build_options`, this could easily result in multiple packages with the same name pointing to different files.
As before this is an uncommon issue to hit as multiple `OptionsStep`s is rare. 
This is one more reason the TODO in `build_runner.zig` would be nice to fix.
https://github.com/zigtools/zls/blob/8cf93fc96a705d30771843de6415ddfce097022f/src/special/build_runner.zig#L93-L95

Before:
```
$ /home/lee/bin/zig run /home/lee/src/zls/src/special/build_runner.zig --cache-dir /home/lee/.cache/zls --pkg-begin @build@ build.zig --pkg-end -- /home/lee/bin/zig /home/lee/src/fff/ zig-cache ZLS_DONT_CARE
{
    "packages": [
        {
            "name": "args",
            "path": "libraries/zig-args/args.zig"
        },
        {
            "name": "known_folders",
            "path": "libraries/known-folders/known-folders.zig"
        },
        {
            "name": "bestline",
            "path": "libraries/bestline/bestline.zig"
        },
        {
            "name": "tracy",
            "path": "libraries/tracy/tracy.zig"
        },
        {
            "name": "zriscv",
            "path": "zriscv/zriscv.zig"
        },
        {
            "name": "bitjuggle",
            "path": "libraries/zig-bitjuggle/bitjuggle.zig"
        }
    ],
    "include_dirs": [
        "libraries/bestline/bestline"
    ]
}%
```

After:
```
$ /home/lee/bin/zig run /home/lee/src/zls/src/special/build_runner.zig --cache-dir /home/lee/.cache/zls --pkg-begin @build@ build.zig --pkg-end -- /home/lee/bin/zig /home/lee/src/fff/ zig-cache ZLS_DONT_CARE
{
    "packages": [
        {
            "name": "build_options",
[...]

PR #688, opened 2022-09-29
Revamp bulid.zig discovery

mattnite:
The current method for finding a build file for a given URI involves walking up the filesystem until one is found. This is a problem when a dependency provides the root file for your executable, such as in [microzig](https://github.com/ZigEmbeddedGroup/microzig). [rp2040](https://github.com/ZigEmbeddedGroup/rp2040) is an example of a package that uses it. If I'm editing a path in rp2040, say:

```
rp2040/src/hal/adc.zig
```

with a root file in

```
rp2040/deps/microzig/src/core/microzig.zig
```

what ends up happening is that zls walks upwards and determines the build file to be:

```
rp2040/deps/microzig/build.zig
```

when we want:

```
rp2040/build.zig
```

This is a problem in rp2040 because zls ends up giving completions for an AVR based microcontroller instead of an ARM microcontroller, an artifact from the microzig build file.

### This solution:

Since there does not seem to be a notion of "project root", this patch introduces a new strategy for build.zig discovery where we walk down the filesystem from root towards our file of interest. For each build.zig it encounters, it loads the build configuration and _only associates the build file with the file of interest if that file is imported/used_, otherwise it continues walking. Finally, if no build file is technically associated, zls will used the closest build.zig.

This makes it so that higher build.zig files have precedence for completions, and should be correct in more situations.

### DIFF
--- a/src/DocumentStore.zig
+++ b/src/DocumentStore.zig
@@ -354,24 +354,63 @@ fn createBuildFile(self: *DocumentStore, build_file_path: []const u8) !*BuildFil
     return build_file;
 }
 
+fn uriAssociatedWithBuild(
+    self: *DocumentStore,
+    build_file: *BuildFile,
+    uri: []const u8,
+) bool {
+    var checked_uris = std.StringHashMap(void).init(self.allocator);
+    defer {
+        var it = checked_uris.iterator();
+        while (it.next()) |entry|
+            self.allocator.free(entry.key_ptr.*);
+
+        checked_uris.deinit();
+    }
+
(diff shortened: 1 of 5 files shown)


SuperAuguste (approved):
LGTM

leecannon:
Just wanted to add that this also hits the problem of duplicate package names like #686 does but worse as the package is *always* called `root`.

If a build.zig has multiple `LibExeObjStep`s then `root` will be mapped to the only one of them. Zig itself hits this issue as will any project with examples.

PR #690, opened 2022-09-29
inlay_hints: implement ability to remove redundant hints

vesim987:
This adds a way to disable inlay hints when they are redundant, for example: `foo: foo`, `foo: &foo`, `foo: bar.foo`

Some screenshots:
inlay hints without this feature:
image
with inlay_hints_hide_redundant_param_names:
image
with inlay_hints_hide_redundant_param_names_last_token:
image
with both enabled:
image


SuperAuguste (approved):
Simple but elegant!

SuperAuguste:
That's on me, I forgot that we had all those config option sites. I'll add them all now.

PR #675, opened 2022-09-26
support pkgconfig

leecannon:
Support getting include paths using pkgconfig, requires https://github.com/ziglang/zig/pull/12981

This allows `cImport` completion to include all the include paths that things like `exe.linkSystemLibrary("gtk+-3.0");` add.

With master zls:
```
$ /home/lee/bin/zig run /home/lee/src/zls/src/special/build_runner.zig --cache-dir /home/lee/.cache/zls --pkg-begin @build@ /home/lee/src/fff/build.zig --pkg-end -- /home/lee/bin/zig /home/lee/src/fff/ zig-cache ZLS_DONT_CARE
{
    "packages": [],
    "include_dirs": []
}%
```

With this PR:
```
$ /home/lee/bin/zig run /home/lee/src/zls/src/special/build_runner.zig --cache-dir /home/lee/.cache/zls --pkg-begin @build@ /home/lee/src/fff/build.zig --pkg-end -- /home/lee/bin/zig /home/lee/src/fff/ zig-cache ZLS_DONT_CARE
{
    "packages": [],
    "include_dirs": [
        "/usr/include/gtk-3.0",
        "/usr/include/pango-1.0",
        "/usr/include/glib-2.0",
        "/usr/lib/glib-2.0/include",
        "/usr/include/sysprof-4",
        "/usr/include/harfbuzz",
        "/usr/include/freetype2",
        "/usr/include/libpng16",
        "/usr/include/libmount",
        "/usr/include/blkid",
        "/usr/include/fribidi",
        "/usr/include/cairo",
        "/usr/include/lzo",
        "/usr/include/pixman-1",
        "/usr/include/gdk-pixbuf-2.0",
        "/usr/include/gio-unix-2.0",
        "/usr/include/cloudproviders",
        "/usr/include/atk-1.0",
        "/usr/include/at-spi2-atk/2.0",
        "/usr/include/dbus-1.0",
        "/usr/lib/dbus-1.0/include",
        "/usr/include/at-spi-2.0"
    ]
}%
```

PR #691, opened 2022-09-29
support zig 0.9.1

leecannon:
As the `build_runner.zig` is executed by the users zig compiler supporting 0.9.1 is a good idea.

Unfortunately the recent pkgconfg #675 and OptionsStep #686 support has to be sacrificed when using 0.9.1.

### DIFF
--- a/src/special/build_runner.zig
+++ b/src/special/build_runner.zig
@@ -118,7 +118,10 @@ pub fn main() !void {
     );
 }
 
-fn reifyOptions(step: *std.build.Step) !void {
+fn reifyOptions(step: *std.build.Step) anyerror!void {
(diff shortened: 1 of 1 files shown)


SuperAuguste:
I don't believe support for 0.9.1 makes sense as zls cannot be built with that version anyway (though the prebuilt binaries *mostly* work with that version).

leecannon:
I think it is an easy hurdle to jump to atleast make the `build_runner` 0.9.1 compatible that way someone using 0.9.1 can still use latest zls.

SuperAuguste (approved):

SuperAuguste:
Good point

PR #693, opened 2022-09-29
remove types.TextDocument

Techatrix:
The main reason for this change is getting rid of `borrowNullTerminatedSlice`, because it temporarily modifies the source text. This may not seem like a big problem but it is one of the thinks that needs to change before the possibility of processing request in parallel becomes possible.

I've also replaced `applyChanges` with a cleaner implementation. We may consider setting [TextDocumentSyncKind](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocumentSyncKind) to `.Incremental` which should be especially helpful if you are dealing with very large files.
https://github.com/zigtools/zls/blob/3c4535a32111fc995d805c015f8f89b9b608e01d/src/Server.zig#L1487


SuperAuguste (approved):
LGTM :)

Issue #694, opened 2022-09-30
Simplify config structure


SuperAuguste:
At the moment when someone wants to add a config option, they must modify five sites:
1. `Config.zig`
2. `requests.zig` for didChangeConfiguration to work
3. `setup.zig` for the option to appear in the configuration wizard
4. `schema.json` so `zls.json` gets proper typing
5. `README.md` so people know it exists

We should probably establish a single source of truth and automate the rest. Some quick remediations:
1. Use comptime to remove the need for manual maintenance of the `requests.zig` and `Config.zig` sites
2. Adapt / create a script based on [this script](https://github.com/zigtools/zls/blob/master/src/data/generate-vscode-config.js) so it creates more than just a VSCode config (this would handle `schema.json` and `README.md`)

PR #696, opened 2022-09-30
remove requests.Configuration

Techatrix:
Replaces `requests.Configuration` with some comptime code that derives this Struct based on `Config`. #694
I've also seen that VS Code always responds with `null` in `didChangeConfigurationHandler` instead of providing the updated settings.


SuperAuguste:
Yep, the null response is why I added the `requestConfiguration` call at the end of didChangeConfiguration! :)

SuperAuguste (approved):
Absolutely awesome! :))

PR #702, opened 2022-10-09
DocumentStore refactor

Techatrix:
Previously DocumentStore used a reference counting like approach for cleaning up unused document. This approach is unfitting because of cyclic dependencies.
As an example if `A.zig` imports `B.zig` and `B.zig` imports `A.zig` they will never be closed.
This new approach should collect all unneeded documents for cleanup.

CImports also received some care and are now managed globally instead of per document.

### DIFF
--- a/src/DocumentStore.zig
+++ b/src/DocumentStore.zig
@@ -302,6 +302,9 @@ fn garbageCollectionCImports(self: *DocumentStore) error{OutOfMemory}!void {
 }
 
 fn garbageCollectionBuildFiles(self: *DocumentStore) error{OutOfMemory}!void {
+    const tracy_zone = tracy.trace(@src());
+    defer tracy_zone.end();
+
     var reachable_build_files = std.StringHashMapUnmanaged(void){};
     defer reachable_build_files.deinit(self.allocator);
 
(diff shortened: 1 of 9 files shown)


leecannon:
This seems to have broken completion of `@import("builtin")`.

Completion will show public decls from `build.zig` and ctrl+clicking will also open `build.zig`.

Techatrix:
> This seems to have broken completion of `@import("builtin")`.

good catch! this is fixed now

leecannon (approved):

leecannon:
There is a significant performance hit associated with these changes.

_TLDR: ReleaseSafe 73x longer to reach first `textDocument/semanticTokens/full` and 7x more memory in use_

### Process:
1. Clone one of the bigger projects [zigwin32](https://github.com/marlersoft/zigwin32)
2. Modify ZLS by changing `Server.zig` line 2410 (master) / line 2439 (PR) to `log.info` so the time to do is also displayed in release modes
3. open vscode in the zigwin32 project _with no documents open_
4. open `win32.zig` in the root of the project
5. wait for the time to do `textDocument/semanticTokens/full` to be printed

### Results:
| Source  | Build Mode | Time | Memory | Time change from Master | Memory change from Master
| ------------- | ------------- | ------------- | ------------- | ------------- | ------------- |
| Master | Debug | 1,885ms | 178mb | - | - |
| PR | Debug | 49,170ms | 1,139mb | ~26x longer | ~6x more memory |
| Master | ReleaseSafe | 459ms | 150mb | - | - |
| PR | ReleaseSafe | 33,653ms | 1,057mb | ~73x longer | ~7x more memory |

Techatrix:
I didn't expect the performance penalty to be this bad. I will just continue to work on returning to "load on demand" when i have the time and then this could be merged.

Techatrix:
The memory consumption and runtime should now be close to master.

leecannon (approved):
Looks good. :+1:

PR #703, opened 2022-10-09
Separate argument placeholder completions from `enable_snippets` to `enable_argument_placeholders` option [fixes #685]

### LINKED ISSUE #685: Feature flag request: Separate argument injection from snippets
Would it be possible to separate argument injection into its own config independent of `enable_snippets`?

With `enable_snippets` enabled, accepting a function completion also injects the function's args and types,

e.g.

```zig
exe.addCS
```
is completed to

```zig
exe.addCSourceFile(file: []const u8, flags: []const []const u8)
```

instead of

```zig
exe.addCSourceFile()
```

I'm sure this is quite useful to some people and some code editors, but it seems like a separate functionality to be toggled independently (and to me it is quite annoying).

So basically change the code here to use different flag, and add docs for it & everything https://github.com/zigtools/zls/blob/8eeae0286559a5765adb55d82c4b2046786fcc58/src/Server.zig#L498

I can create a PR if you are ok with the change.

peteruhnak:
Separate argument placeholder completions from enable_snippets to enable_argument_placeholders option [fixes #685

_Edit: I have removed my comments about completing `func` to `func()` as it is is out of scope, and would require more nuance._

### DIFF
--- a/schema.json
+++ b/schema.json
@@ -9,6 +9,9 @@
             "type": "boolean",
             "default": "false"
         },
+        "enable_argument_placeholders": {
+            "description": "Enabes argument placeholder completions, e.g. `@intCast` completes to `@intCast(comptime DestType: type, int: anytype)` instead of `@intCast()`"
+        },
         "enable_ast_check_diagnostics": {
             "description": "Whether to enable ast-check diagnostics",
             "type": "boolean",
--- a/README.md
+++ b/README.md
@@ -105,6 +105,7 @@ The following options are currently available.
 | Option | Type | Default value | What it Does |
 | --- | --- | --- | --- |
 | `enable_snippets` | `bool` | `false` | Enables snippet completions when the client also supports them. |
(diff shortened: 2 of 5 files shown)


SuperAuguste:
Wow we totally missed this, so sorry! LGTM but I don't know if the default should be `false` or `true` tbh, I've grown quite fond of on-by-default argument snippets...

peteruhnak:
Yeah, makes sense not change everyone's current behavior, I'll change the default.

schema.json
```diff
@@ -9,6 +9,9 @@
             "type": "boolean",
             "default": "false"
         },
+        "enable_argument_placeholders": {
+            "description": "Enabes argument placeholder completions, e.g. `@intCast` completes to `@intCast(comptime DestType: type, int: anytype)` instead of `@intCast()`"
```
SuperAuguste, line 13: Don't forget to add the type and default! :)

jiacai2050:
@SuperAuguste Hi, I wonder why this PR is close? 

I want to disable fn args completion.

PR #709, opened 2022-10-13
build.zig: ensure user zig matches program minimum

nektro:
make sure to update this anytime breaking changes come through

exact versions can be found by checking the CI runs on the commit where a breaking change is added upstream


leecannon (approved):
Good idea :+1:

PR #712, opened 2022-10-17
fix symbol references

Techatrix:
i broke references a while ago and this should return it to how it previously was.


SuperAuguste (approved):
LGTM

PR #717, opened 2022-10-23
fix allocator mismatch in tagStoreCompletionItems

Techatrix:


SuperAuguste (approved):
oop sorry for missing this, nice leak find!

PR #720, opened 2022-10-27
Improve folding regions

InKryption:
+ a few other small formatting fixes.
This adds support for folding for `else_expr` in
```
_ = <control flow>
    then_expr
else
    else_expr;
```
where `<control flow>` can be
* `while (expr)` `|capture|`? ` : (expr)`?
* `for (expr) |capture, ...|`
* `if (expr)` `|capture|`?

Also adds folding for multiline strings, doc comments, and container doc comments.

Using the node tree is slightly more complex, but it allows for more targeted and specific folding ranges, without having to waste memory on creating a stack for each type of nest-able folding range.

I'm unsure whether it would be better to traverse the node tree as it's structured rather than as a contiguous list as done in this design. I think the benefits would be marginal for most cases.

Note: only tested with VSCode.

### DIFF
--- a/src/Server.zig
+++ b/src/Server.zig
@@ -2339,63 +2339,245 @@ fn codeActionHandler(server: *Server, writer: anytype, id: types.RequestId, req:
     });
 }
 
-fn foldingRangeHandler(server: *Server, writer: anytype, id: types.RequestId, req: requests.FoldingRange) !void
-{
-    const Tag = std.zig.Token.Tag;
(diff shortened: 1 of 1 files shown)


InKryption:
Should be ready for merging, if these changes are acceptable.

SuperAuguste (approved):
LGTM :)

I think this contextless iteration model works just fine tbh; hopefully this works in other editors 🙏

PR #732, opened 2022-11-01
Add CI workflow to check build_runner works with a range of zig versions

leecannon:
This workflow is only run if `build_runner.zig` is modified, should it also have a schedule to catch changes in master?

Currently the versions checked are: 
- 0.9.1 (as the 0.10.0 [release notes](https://ziglang.org/download/0.10.0/release-notes.html#Is-it-time-to-upgrade) have a section about waiting for 0.10.1, I think we should still check 0.9.1)
- 0.10.0
- master

### DIFF
--- a/.github/workflows/build_runner.yml
+++ b/.github/workflows/build_runner.yml
@@ -1,0 +1,41 @@
+name: BuildRunner
+
+on:
+  push:
+    paths:
+      - "src/special/build_runner.zig"
+  pull_request:
+    paths:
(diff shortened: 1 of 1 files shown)


SuperAuguste:
Now this is cool :)

Maybe a cron job for like a once a day job could be nice because the build system is still changing so quickly? Just something to consider.

leecannon:
Added daily scheduled run as well.

SuperAuguste (approved):
LGTM

PR #735, opened 2022-11-02
Add builtin data for 0.9.1 and 0.10.0

BratishkaErik:
~~Copied `src/data/master.zig` to `src/data/0.10.0.zig`~~


nektro:
the way this is meant to be done is changing the version on line 6 to be the stable one and then running `src/data/generate-data.py`

BratishkaErik:
> the way this is meant to be done is changing the version on line 6 to be the stable one and then running `src/data/generate-data.py`

Oops, didn't know this, thanks!

SuperAuguste (approved):
LGTM :)

PR #737, opened 2022-11-03
update flake.lock

sagehane:
Since the minimum supported version of Zig changed as per https://github.com/zigtools/zls/commit/bb727d263b94cc3dfd871dc638a8da49ea8b05ca.

PR #739, opened 2022-11-03
Nix: Pass `-Dcpu=baseline` to `zig build`

### LINKED ISSUE #738: Question: Should the Nix file aim to build ZLS reproducibly?
Related: https://github.com/NixOS/nixpkgs/issues/185665

Nix's selling point is reproducibility. However, Zig normally targets for the native system unless specified otherwise.
Nixpkgs addresses this by passing `-Dcpu=baseline`[0].

Our current `flake.nix` file doesn't do this[1], and I personally think it should be "fixed". This might affect setups like people using Cachix or similar to serve other users substitutes that they've built. Thoughts?

[0] - https://github.com/NixOS/nixpkgs/blob/8ff7b290e6dd47d7ed24c6d156ba60fc3c83f100/pkgs/development/tools/zls/default.nix#L21:L23
[1] - https://github.com/zigtools/zls/blob/master/flake.nix#L34:L37

sagehane:
Closes #738 and depends on #737 for successful compilation.

Also, formatted some code with `nixpkgs-fmt`.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,5 +1,6 @@
-{ pkgs ? import <nixpkgs> {},
-  system ? builtins.currentSystem }:
+{ pkgs ? import <nixpkgs> { }
+, system ? builtins.currentSystem
+}:
 
 let
   zig-overlay = pkgs.fetchFromGitHub {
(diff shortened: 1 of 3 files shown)


SuperAuguste (approved):
LGTM (from my nix-virgin eyes, but I'll trust you on this)

PR #742, opened 2022-11-06
Update the main CI workflow

nullptrdevs:
targets i386-* -> x86-*
use actions/checkout@v3 (v2 uses Node.js 12 which is about to be deprecated)


SuperAuguste (approved):
LGTM

PR #744, opened 2022-11-08
Improve Zig version mismatch error

nickcernis:
Improves the server message we send if the system Zig version is older than the Zig version that ZLS was built with:

- Correct typo (“build with” → “built with”).
- Show versions so users know which one they need to update beyond.
- Suggest step needed to fix the error (“Update Zig…”).

Example of messages in VS Code:

| Before | After |
| - | - |
| Screenshot 2022-11-08 at 18 18 49 | Screenshot 2022-11-08 at 18 10 59 |


SuperAuguste (approved):
Thanks :)

PR #747, opened 2022-11-13
panic when setting configuration

ilohmar:
### Zig Version

0.10.0

0.10.0-dev.364+88d1693

### Steps to Reproduce

Using Emacs' eglot client to connect to zls, I started getting `zls` panics after each start in every file nowadays.

This was after upgrading Emacs+eglot, but I still need to check what has changed there.  In any case, `zls` should surely not panic like that..  Attached is eglot's log for one whole initialization run and the final panic.

[zls-after-eglot-update.txt](https://github.com/zigtools/zls/files/9995169/zls-after-eglot-update.txt)

### Expected Behavior

`zls` starts and is configured as before my Emacs/eglot update.

### Actual Behavior

`zls` panics (and is restarted several times by the LSP client, repeating the cycle..)

### DIFF
--- a/src/Server.zig
+++ b/src/Server.zig
@@ -2644,6 +2644,7 @@ pub fn processJsonRpc(server: *Server, writer: anytype, json: []const u8) !void 
         // Yes, String ids are officially supported by LSP
         // but not sure how standard this "standard" really is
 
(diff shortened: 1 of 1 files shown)


SuperAuguste (approved):
LGTM

PR #750, opened 2022-11-14
README: Change '--config' to 'config'

jgollenz:


leecannon:
I'm a bit confused by this PR `zls --config` is correct.

```
$ zls --config
Welcome to the ZLS configuration wizard!
      *
       |\
      /* \
      |  *\
    _/_*___|_    x
      | @ @     /
     @     \   /
      \__-/   /

? Should this configuration be system-wide? (y/n) > ^C
```

jgollenz:
Ah, `zig build` ofc doesn't replace the old binary I had in my path. Build went through but I was still using an old version that could not deal with `--config`. I'm sorry

PR #755, opened 2022-11-16
Format with `Ast.render` instead of `zig fmt`

Techatrix:
We don't need to call `zig fmt` because we can format the code ourself with `std.zig.Ast.render`. This should provide a more stable experience with formatting because don't need to spawn a child process and don't depend on a zig executable for formatting.


nektro (approved):

SuperAuguste (approved):
Amazing.

PR #760, opened 2022-11-16
shout-out to zls for helping with zls's development

Techatrix:
ZLS has really helped me and others while working on zls so i think we should at least show some appreciation for their hard work. /s


SuperAuguste:
stop padding your commits to beat me Techatrix!! :P /j

but this is pretty funny ngl, congrats on becoming (tied with) the number two contributor!

PR #766, opened 2022-11-18
don't format files with syntax errors

Techatrix:
yes, i introduce bugs
yes, it will happen again
yes, i am ashamed of myself

### DIFF
--- a/src/Server.zig
+++ b/src/Server.zig
@@ -2098,6 +2098,11 @@ fn formattingHandler(server: *Server, writer: anytype, id: types.RequestId, req:
         return try respondGeneric(writer, id, null_result_response);
     };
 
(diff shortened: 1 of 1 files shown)


src/Server.zig
```diff
@@ -2098,6 +2098,11 @@ fn formattingHandler(server: *Server, writer: anytype, id: types.RequestId, req:
         return try respondGeneric(writer, id, null_result_response);
     };
 
+    if (handle.tree.errors.len != 0) {
+        try server.showMessage(writer, .Info, "Files with syntactic errors cannot be formatted");
```
SuperAuguste, line 2102: I think this might get spammy quick (for format on save) so an empty return would be better.

SuperAuguste:
YOLO

PR #772, opened 2022-11-21
correctly find last full text change in applyTextEdits

Techatrix:


SuperAuguste (approved):
I mean it works on my machine so...

PR #776, opened 2022-11-22
set TextDocumentSync to Incremental

### LINKED ISSUE #762: Range/incremental edits
It seems like we have the code required for range-based [textDocument/didChange](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_didChange), yet as kcbanner on our Discord mentioned, our capabilities indicate that we only support `.Full` document change notifications. Is there a specific reason for this? If not, we should take a go at enabling them and ensure that turning on range edits doesn't make anything explode. :P

Techatrix:
this relies on `diff.applyTextEdits` working properly which it likely does based on my limited testing.
fixes #762

### DIFF
--- a/src/Server.zig
+++ b/src/Server.zig
@@ -1606,7 +1606,7 @@ fn initializeHandler(server: *Server, writer: anytype, id: types.RequestId, req:
                     },
                     .textDocumentSync = .{
                         .openClose = true,
-                        .change = .Full,
+                        .change = .Incremental,
                         .save = true,
                     },
                     .renameProvider = true,


SuperAuguste:
I don't think that was the consensus, rather that it *could* take longer in certain cases. But generally this is quite a fast/low-cost operation with the real bottleneck being analysis. Regardless, I think this is a step in the right direction (especially with the new fixes)

SuperAuguste (approved):
YOLO

PR #791, opened 2022-12-01
Allocgate 2.0 slain

SuperAuguste:


SuperAuguste:
@alexnask Tuple support broke semantic token highlighting and enums it seems. Any ideas?

alexnask:
@SuperAuguste

Semantic token highlighting is working just fine for me (I'm 100% confident since I don't use any zig extention other than zls on VSCode  specifically to test it) although I am running a handful of commits behind master  
Not sure what could have caused the enum compleitons to break though.

SuperAuguste:
Yeah everything works for me too, I just mean that they fail in the test suite for some reason :(

nektro:
`build.zig` should be updated too

SuperAuguste:
Good call - I'll bump it up next time I commit.

Issue #804, opened 2022-12-04
URI escaping on windows does not correctly calculate length of resulting string


josephst:
### Zig Version

0.11.0-dev.462+af4361f57

### Zig Language Server Version

0.11.0-dev.59+c2d6b95

### Steps to Reproduce

`src/uri.zig` has various methods for converting strings to URIs as well as `parse` to convert from a URI to a string

On Windows, a file path such as `C:\main.zig` will be converted to `file:///c%3A/main.zig` with the `:` character replaced with `%3A`. When converting back to a file path, `%3A` will be replaced by `:`. However, the `parse()` method does not take this 2 character length reduction into account which results in the final string having two extra characters at the end of the string consisting of uninitialized memory. 

The malformed string results in the language server ultimately failing to initialize and most features (autocomplete, go to definition, etc) are unavailable on Windows. 

Example: 
```
====== expected this output: =========
c:\main.zigΓÉâ

======== instead found this: =========
c:\main.zig¬¬ΓÉâ

======================================
First difference occurs on line 1:
expected:
c:\main.zig
           ^ (end of string)
found:
c:\main.zig¬¬
           ^ ('\xaa')
```

PR fixing this to come (including addition of a failing test demonstrating this issue)!

### Expected Behavior

Should replace %3A with : character and properly reduce size of allocated string. 

On a side note, there may be an underlying bug in zig with allocator.resize() - it is not properly resizing the allocated string at the end of the `parse()` function. The final call to `allocator.resize(uri, i);` in the `parse()` function seems like it should resize the allocated memory but when examining this loop with a debugger, even when `i` is i.e. 40, `uri.len` remains 42

### Actual Behavior

String is not resized and two bytes of uninitialized memory are included at the end of the string

SuperAuguste:
Yep, this is the cause of #803, fixing it as I speak :)

SuperAuguste:
So I just got a fix out ASAP but the issue seems to be that `len` isn't modified by the allocator - not sure if this is intentional behavior or not (it seems quite counterintuitive, especially when `shrink` did this) but yeah I'm not sure.

josephst:
Ah thanks for getting that fixed so fast! Just sent in a PR with some tests I added that fail because of this bug. I think I was chasing the wrong thing - the colon mishandling on Windows is more a symptom of the problem with .resize() (and hard-coding in a case for `%3A` probably isn't the way to go about fixing this!).

SuperAuguste:
[...]

PR #806, opened 2022-12-04
Allocator.resize() does not adjust size of string allocated for URI to file path conversion

josephst:
Related to problem #804 - escape sequences are removed from the string as it is being built which results in the final string never filling up the memory that was allocated for it. 

This results in a URI such as `file:///c%3A/main%2B.zig` being given 4 more bytes of memory than necessary (and these bytes are left undefined/ uninitialized~). 

URI.parse() correctly calculates the size of the final string (based on inspecting `i` in URI.parse() with a debugger) but the final call to `allocator.resize(allocator.resize(uri, i)) does not actually affect the size of the array. 

This pull request currently includes tests that fail as a result of this bug. Still working on fixing the underlying bug but wanted to include some tests in case anyone else out there can fix the problem sooner!

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,4 +1,0 @@
-{
-    "zls.path": "C:/Users/Joseph/source/zig/zls/zig-out/bin/zls.exe",
-    "zls.debugLog": true
-}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,4 @@
+{
+    "zls.path": "C:/Users/Joseph/source/zig/zls/zig-out/bin/zls.exe",
+    "zls.debugLog": true
+}
--- a/build.zig
+++ b/build.zig
@@ -6,7 +6,7 @@ const zls_version = std.builtin.Version{ .major = 0, .minor = 11, .patch = 0 };
 
 pub fn build(b: *std.build.Builder) !void {
     const current_zig = builtin.zig_version;
-    const min_zig = std.SemanticVersion.parse("0.11.0-dev.323+30eb2a175") catch return; // added tuple declaration support
+    const min_zig = std.SemanticVersion.parse("0.11.0-dev.399+44ee1c885") catch return; // whereabouts allocgate 2.0
     if (current_zig.order(min_zig).compare(.lt)) @panic(b.fmt("Your Zig version v{} does not meet the minimum build requirement of v{}", .{ current_zig, min_zig }));
 
     const target = b.standardTargetOptions(.{});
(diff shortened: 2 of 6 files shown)


SuperAuguste:
You accidentally included your VSCode config in the mix by the way :P

SuperAuguste:
Wow `realloc` is exactly what we need - totally missed it, nice find! Could you please quickly rebase this (merge conflicts currently present in `uri.zig`) - thanks! :)

josephst:
Ah - so I did 😅 luckily just my testing config! 
Just pushed another commit with a call to `allocator.resize()` which appears (in brief testing) to properly reduce the length of the array!

josephst:
Anyways, you seem to be catching my commits faster than I can type them up! Going to leave the rest to you to fix the merge conflict - thanks for getting it fixed up!

SuperAuguste:
Testing this locally, `realloc` seems to be just as unreliable - it seems to return uninitialized memory, causing test failures.

josephst:
Also, from looking at the CI runners, the tests for windows-specific formatting (specifically, `file:///c/...` instead of `file://src/...` may be causing problems - these can just be commented out or something. To be honest, I was picking up Zig for doing Advent of Code exercises so I'm a bit over my head! Not sure if there's an easy way in Zig to have certain tests only run on windows CI runners

SuperAuguste:
No worries I'll handle it :)

josephst:
Investigated this for a while this morning - I think the best approach right now is just to isolate the platform-specific code to that particular test runner (test Windows-specific formatting on Windows test runners, etc)

From looking at how Python does it, the [as_uri](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.as_uri) method produces Unix or Windows style URIs but pathlib lacks any good way to convert from URI to file path. Meanwhile [urllib](https://docs.python.org/3/library/urllib.parse.html#module-urllib.parse) has various methods for converting from a `file:///` URI to a local file path although it also doesn't seem particularly robust; converting a file path => URI => back to file path results in a leading slash being appended to the path. [There's ongoing discussion about it](https://discuss.python.org/t/file-uris-in-python/15600); clearly no great way to handle this cross-platform

```python
import pathlib
from urllib.parse import urlparse
urlparse(pathlib.PureWindowsPath("C:\\foo.txt").as_uri()) # ParseResult(scheme='file', netloc='', path='/C:/foo.txt', params='', query='', fragment='')
```

Anyways, isolating the tests to each platform so that Windows-specific URI stuff only runs on Windows test platforms and Unix-specific URI stuff only runs on Unix-like platforms seems to fix the URI parsing/ encoding problems. 

Additionally, latest commit also changes some of the `uri - pathRelative` test. `file://project/zig` refers to a file `zig` on host `project` whereas `file:///project/zig` refers to a file `zig` on localhost under the directory `/project`. I think testing the latter is better here.

SuperAuguste:
Sounds good to me!

SuperAuguste:
ayo it passed, thank you so much!

SuperAuguste (approved):

PR #835, opened 2022-12-18
track builtin API changes from zig/master

skabbes:
**Context**
I was getting set up to use zls for the first time, building from zig master - I came across some compile errors. I tracked it down to a very recent API change in zig.

https://github.com/ziglang/zig/pull/13930

In this PR, std.builtin.Type.field_type is renamed to type And there is discussion that all Enum layouts are .Auto.


skabbes:
Thanks, I wasn't aware of that at ll - I bumped it up.

SuperAuguste:
CI failure caused by website build being outdated - I'll try again tomorrow :(

Thanks for this PR!

leecannon (approved):

PR #840, opened 2022-12-20
Fix Nix build, add Nix to CI

erikarvstedt:
### Commit: `Nix: Fix build by updating the Zig version`
  Fix breakage introduced in https://github.com/zigtools/zls/commit/23ef018521714b988cfca6c34a4832358e3e1309.

### Commit: `CI/main: convert line endings from CRLF to LF`
Review this commit like so:
```bash
mkdir -p /tmp/diff
git show master:.github/workflows/main.yml | sed 's|\r$||' > /tmp/diff/master_converted_to_unix
git show --name-only b692e22813
git show b692e22813:.github/workflows/main.yml > /tmp/diff/from_this_pr
# This only shows some removed trailing whitespace
diff -u --color /tmp/diff/{master_converted_to_unix,from_this_pr}
rm -rf /tmp/diff
```

### Commit `CI: Add Nix build`

  This adds a CI Nix step to test that building the Nix package succeeds.
  Currently, this includes the use of [Cachix](https://www.cachix.org/), a Nix cache service that speeds up builds by caching build results.

  We can either:
  - Don't use Cachix (remove step `Install Cachix` in `workflows/main.yml`).
    This very slightly increases build time because the implicit Zig package build isn't cached.

  - Setup Cachix like so:
    1. Sign up and create a binary cache at https://www.cachix.org/
    2. Go to Cache -> Settings -> Auth Token and create a `Write` auth token
    3. ZLS Github repo settings -> Secrets -> Actions -> New repo secret:
       Name: `CACHIX_AUTH_TOKEN`
       Value: See step 2.
    3. Post the name of the binary cache, and I'll update ths PR.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -4,9 +4,11 @@ on:
   push:
     paths:
       - "**.zig"
+      - "flake.*"
   pull_request:
     paths:
       - "**.zig"
+      - "flake.*"
   schedule:
     - cron: "0 0 * * *"
   workflow_dispatch:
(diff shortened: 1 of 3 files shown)


SuperAuguste:
Looks good to me as far as I can tell (I'm not a nix user so I can't really check any specifics). We've got a steady stream of donations but I'm not sure that spending almost half of it on just Nix caching is really worth it to be frank, so I think going with the non-caching option is better.

erikarvstedt:
@SuperAuguste, cachix is free for Open Source projects.

erikarvstedt:
Rebased to master and added cache.
Review this with:
```bash
git range-diff  121915b...
```

SuperAuguste (approved):

SuperAuguste:
Reverting this for now, hitting payment required despite this being a public cache. I guess I'll have to get in contact with the Cachix team.

PR #844, opened 2022-12-21
semantic_tokens: Fix handleComments not evaluating the last byte

### LINKED ISSUE #842: Off-by-one bug during semanticTokens response that leads to invalid UTF-8
### Zig Version

0.11.0-dev.885+698ff988f

c355a54dd1072fafa1c4b37d6f1d2b5b31255475

### Steps to Reproduce

This is the cause of http://zls-fuzzing.augustera.me/group/31dbf3809604c73790a8a40a2cc70895

The error looks like this:
```
panic: index out of bounds: index 11, len 10
/root/zig/lib/std/unicode.zig:271:24: 0x505805 in nextCodepointSlice (zls)
        return it.bytes[it.i - cp_len .. it.i];
                       ^
/root/zig/lib/std/unicode.zig:275:44: 0x494697 in nextCodepoint (zls)
        const slice = it.nextCodepointSlice() orelse return null;
                                           ^
/root/sus/repos/zls/src/offsets.zig:271:38: 0x3dc490 in countCodeUnits (zls)
            while (iter.nextCodepoint()) |codepoint| {
                                     ^
/root/sus/repos/zls/src/offsets.zig:88:31: 0x332660 in locLength (zls)
    return countCodeUnits(text[loc.start..loc.end], encoding);
                              ^
/root/sus/repos/zls/src/semantic_tokens.zig:179:114: 0x3324e6 in handleComments (zls)
            const length = offsets.locLength(self.handle.tree.source, .{ .start = comment_start, .end = i }, self.encoding);
                                                                                                                 ^
/root/sus/repos/zls/src/semantic_tokens.zig:102:59: 0x332c21 in finish (zls)
            try handleComments(self, starts[i - 1], starts[i]);
                                                          ^
/root/sus/repos/zls/src/semantic_tokens.zig:1043:23: 0x333049 in writeAllSemanticTokens (zls)
    try builder.finish();
                      ^
/root/sus/repos/zls/src/Server.zig:1917:121: 0x34006c in semanticTokensFullHandler__anon_13968 (zls)
    const token_array = try semantic_tokens.writeAllSemanticTokens(&server.arena, &server.document_store, handle, server.offset_encoding);
                                                                                                                        ^
/root/sus/repos/zls/src/Server.zig:2878:31: 0x389528 in processJsonRpc__anon_11353 (zls)
                method_info[2](server, writer, id, request_obj) catch |err| {
                              ^
/root/sus/repos/zls/src/main.zig:51:34: 0x38f7b2 in loop (zls)
        try server.processJsonRpc(writer, buffer);
                                 ^
```

I did some debugging to try to determine if this was a `std.unicode` bug or a `zls` bug and here's what I found:

If the last token of a file is a doc comment or top-level doc comment with a multibyte UTF-8 code point at the end, then `zls` will omit the last byte from the token and then try to get the length of the now-invalid UTF-8 which causes the error. I have not investigated where exactly this off-by-one is happening.
 
For example, in [this log that causes the error](http://zls-fuzzing.augustera.me/log/logs-1671525189120/stdin), the relevant portion of the generated Zig code before the error happens is this comment at the end:

```zig
//! └────┬────────────────┬────────────
```

The last codepoint is `U+2500` which is encoded in UTF-8 as `\xe2\x94\x80`. The token that `zls` tries to get the length of, however, ends with the byte sequence `\xe2\x94`. The full `\xe2\x94\x80` sequence *is* in memory, though, so that's why I think this is an off-by-one length issue.

As far as I can tell, this only occurs if all the following are true:
- The comment is a doc comment or top-level doc comment
- The comment has a multibyte UTF-8 codepoint at the end (note: I think it might be the case that the byte at the end would be omitted from the token if it was not a multibyte codepoint, but it wouldn't crash)
- The comment ends at the end of the file exactly (no trailing newline)

Minimal reproduction:

```zig
///─
```

or

```zig
//!─
```

### Expected Behavior

No panic

### Actual Behavior

Panic

### LINKED ISSUE #843: semantic_tokens: Normal comments don't get evaluated if they are alone
### Zig Version

0.11.0-dev.885+698ff988f

c355a54dd1072fafa1c4b37d6f1d2b5b31255475

### Steps to Reproduce

The following added `semantic_tokens` test fails:

```
try testSemanticTokens(
        \\// anything
    ,
        &.{ 0, 0, 11, 8, 0 },
    );
```

Note: This is only true for normal comments, doc comments and top-level doc comments do work (see #844)

### Expected Behavior

The test to pass

### Actual Behavior

```
Test [55/61] test.semantic tokens - comments...
{"data":[0,0,3,11,0]}␃

{"data":[]}␃

expected:
{"data":[0,0,3,11,0]}
         ^ ('\x30')
found:
{"data":[]}
         ^ ('\x5d')
Test [55/61] test.semantic tokens - comments... FAIL (TestExpectedEqual)
C:\Users\Ryan\Programming\Zig\zig\lib\std\testing.zig:881:9: 0x7ff6a72e482d in expectEqualStrings (test.exe.obj)
        return error.TestExpectedEqual;
        ^
C:\Users\Ryan\Programming\Zig\zls\tests\context.zig:116:9: 0x7ff6a737f8e3 in request (test.exe.obj)
        try std.testing.expectEqualStrings(expected, result_json);
        ^
C:\Users\Ryan\Programming\Zig\zls\tests\lsp_features\semantic_tokens.zig:72:5: 0x7ff6a738fe23 in testSemanticTokens (test.exe.obj)
    try ctx.request(
    ^
C:\Users\Ryan\Programming\Zig\zls\tests\lsp_features\semantic_tokens.zig:26:5: 0x7ff6a738ff62 in test.semantic tokens - comments (test.exe.obj)
    try testSemanticTokens(
    ^
```

squeek502:
Fixes #842

Does not fix #843, but does fix the last byte of all comments not being included in their semantic tokens.

### DIFF
--- a/src/semantic_tokens.zig
+++ b/src/semantic_tokens.zig
@@ -141,16 +141,16 @@ const Builder = struct {
         const source = self.handle.tree.source;
 
         var i: usize = from;
-        while (i < to - 1) : (i += 1) {
+        while (i < to) : (i += 1) {
             // Skip multi-line string literals
             if (source[i] == '\\' and source[i + 1] == '\\') {
-                while (i < to - 1 and source[i] != '\n') : (i += 1) {}
+                while (i < to and source[i] != '\n') : (i += 1) {}
                 continue;
             }
             // Skip normal string literals
             if (source[i] == '"') {
                 i += 1;
-                while (i < to - 1 and
+                while (i < to and
                     source[i] != '\n' and
                     !(source[i] == '"' and source[i - 1] != '\\')) : (i += 1)
                 {}
@@ -159,7 +159,7 @@ const Builder = struct {
             // Skip char literals
             if (source[i] == '\'') {
                 i += 1;
-                while (i < to - 1 and
+                while (i < to and
                     source[i] != '\n' and
                     !(source[i] == '\'' and source[i - 1] != '\\')) : (i += 1)
                 {}
@@ -174,7 +174,7 @@ const Builder = struct {
             if (i + 2 < to and (source[i + 2] == '!' or source[i + 2] == '/'))
                 mods.documentation = true;
 
-            while (i < to - 1 and source[i] != '\n') : (i += 1) {}
+            while (i < to and source[i] != '\n') : (i += 1) {}
 
             const length = offsets.locLength(self.handle.tree.source, .{ .start = comment_start, .end = i }, self.encoding);
             try self.addDirect(TokenType.comment, mods, comment_start, length);
--- a/tests/lsp_features/semantic_tokens.zig
+++ b/tests/lsp_features/semantic_tokens.zig
@@ -22,6 +22,16 @@ test "semantic tokens" {
     // TODO more tests
 }
 
+test "semantic tokens - comments" {
+    try testSemanticTokens(
+        \\//!─
+    ,
+        &.{ 0, 0, 4, 8, 128 },
+    );
+
+    // TODO more tests
+}
+
 const file_uri = switch (builtin.os.tag) {
     .windows => "file:///C:/test.zig",
     else => "file:///test.zig",


SuperAuguste (approved):
Absolutely amazing off-by-one catch / fix - thank you so much squeek!! :)

PR #850, opened 2022-12-25
Use zig-lsp-codegen

Techatrix:
Thanks to @SuperAuguste we can use https://github.com/zigtools/zig-lsp-codegen to automatically create zig types for the LSP Specification, which more complete and accurate. From the users point of view not much should be changed.

I've also used this opportunity to rework zls's message handler [`processJsonRpc`](https://github.com/zigtools/zls/blob/f6c15ac10c6f43ed1416911a832b4e4f1f958e23/src/Server.zig#L2728).
https://github.com/ziglibs/tres has been added as a submodule.


SuperAuguste:
So I've been working on this exact same thing in `use-zig-lsp` but with some of the issues you mention (like typechecking responses, for example) addressed. Not sure how to address both of these solutions existing. Maybe I can merge this and then try to rebase my work onto your branch?

SuperAuguste (approved):

PR #851, opened 2022-12-27
fix memory lifetime issues

Techatrix:
I've inserted a randomly failing allocator similar to `std.testing.FailingAllocator` into zls to find some leaks when `error.OutOfMemory` occured.
Once #850 is merged i could try to integrate this into zls so that it can be used with [sus](https://github.com/zigtools/sus) to find even more issues.


SuperAuguste:
That's an awesome technique for testing zls, I'll definitely have to try and integrate it!

SuperAuguste (approved):

PR #855, opened 2022-12-27
fix: tres not available in a nix environment.

acristoffers:
The last commit added tres as a depency, but did so in a way that broke the nix environment. 

This pull-request fixes that by:

- Adding a `tres` option to the build, the same way we have a `known-folders` option.
- Changing the `flake.nix` file to pull `tres` and pass the correct path through the new option.
- Changing the `Server.zig` file to use the correct import statement.

With those changes I now have both `zig build` and `nix build` working.


acristoffers:
Apparently my formatter also removed useless whitespaces and a duplicated empty line in `Server.zig`, but I bet that's now a problem.

SuperAuguste (approved):

PR #859, opened 2022-12-29
Add command to update `master.zig` on Windows Powershell

paoda:
The Cross-Platform Powershell (v6+) makes this a lot easier, but you have to go out of your way to install it. By default, an older version (v5) is installed and this way should (?) work.


SuperAuguste:
I think the correct fix here would be to add `*.py` and `*.js` files to our `.gitattributes` w/ the LF line ending option.

paoda:
Ah the problem isn't that related to what's checked into git, it's more about how I can't update `master.zig` locally becuase just piping to file will result in a UTF-8 with BOM/UTF-16_LE  encoded file with CRLF endings.

SuperAuguste:
Oh my god I am so sorry, totally missed that. Yeah this makes a lot more sense now - LGTM!

SuperAuguste (approved):

PR #862, opened 2022-12-29
add vscode config generation to `zig build gen`

Techatrix:
You can now call `zig build gen -Dvscode-config-path=/path/to/output/file.json` to generate the configuration properties for [`packages.json`](https://github.com/zigtools/zls-vscode/blob/master/package.json) in [zls-vscode](https://github.com/zigtools/zls-vscode), which is why i removed  [generate-vscode-config.js](https://github.com/zigtools/zls/compare/master...Techatrix:vscode-config-gen?expand=1#diff-e301c1554ed3ee63d4a3c875a18e6b4feee45a4a9812771c5de448db847ac7d1).

There has also been a bug where you get corrupted results when the output file becomes smaller.
This tool should now also build on windows even though the output may have CRLF LF mismatches.

Tres is also being updated.


SuperAuguste (approved):
This is a godsend, thank you so much

Issue #856, opened 2022-12-28
[Neovim] LSP diagnostics go to a split


paveloom:
### Zig Version

0.11.0-dev.947+cf822c6dd

### Zig Language Server Version

61c0981294c52820d185afe41a0965a722c3e314

### Steps to Reproduce

So, this is an awkward one. I assume you are familiar with Neovim.

1) Here's a minimal configuration file for Neovim:
	```lua
	-- Set the local root path
	local root = vim.fn.fnamemodify("./.repro", ":p")
	
	-- Set the standard paths to use the local root path
	for _, name in ipairs({ "config", "data", "state", "cache" }) do
	  vim.env[("XDG_%s_HOME"):format(name:upper())] = root .. "/" .. name
	end
	
	-- Bootstrap Lazy
	local lazypath = root .. "/plugins/lazy.nvim"
	if not vim.loop.fs_stat(lazypath) then
	  vim.fn.system({
	    "git",
	    "clone",
	    "--filter=blob:none",
	    "--single-branch",
	    "https://github.com/folke/lazy.nvim.git",
	    lazypath,
	  })
	end
	vim.opt.runtimepath:prepend(lazypath)
	
	-- Install plugins
	local plugins = {
	  "folke/tokyonight.nvim",
	  "neovim/nvim-lspconfig",
	}
	require("lazy").setup(plugins, {
	  root = root .. "/plugins",
	})
	
	-- Setup the theme
	vim.opt.termguicolors = true
	vim.cmd([[colorscheme tokyonight]])
	
	-- Setup the LSP
	local lspconfig = require("lspconfig")
	lspconfig.zls.setup({
	  root_dir = lspconfig.util.root_pattern(".git", "build.zig", "zls.json"),
	  settings = {
	    zls = {
	      enable_inlay_hints = true,
	      enable_snippets = true,
	      warn_style = true,
	    },
	  },
	})
	
	-- Open the `main.zig` file if exists
	local main_path = "src/main.zig"
	if vim.loop.fs_stat(main_path) then
	  vim.cmd(":edit " .. main_path)
	end
	```
	
	Save it as `init.lua`.
2) Run `zig init-exe`
3) Run `nvim -u init.lua`, restart after the plugins are installed
4) Add something to the code that would cause a diagnostic to appear and save the buffer

### Expected Behavior

The diagnostic should be treated as usual, that is, being displayed as virtual text:

[...]

PR #865, opened 2022-12-29
fix ast-check with single error

Techatrix:
fixes an ast-check issue that was caused in 61c0981294c52820d185afe41a0965a722c3e314

i believed that this will f i x  #856 but based on what is being said in the comments that issue is also caused in older version.
I did a `git bisect` between 44b6c4dae4ace789c5a3ebf9fdbeb65db8f1cf30 and c6d74dbca565105b49655e50c062758e6af99127. 61c0981294c52820d185afe41a0965a722c3e314 was the resulting turning point.

### DIFF
--- a/src/Server.zig
+++ b/src/Server.zig
@@ -453,6 +453,12 @@ fn getAstCheckDiagnostics(
             };
         }
     }
+
+    if (last_diagnostic) |*diagnostic| {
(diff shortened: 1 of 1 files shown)


SuperAuguste (approved):
Massive W, awesome stuff! :)

PR #871, opened 2022-12-31
Full compilation diagnostics

### LINKED ISSUE #870: Add way to build project with `-fno-emit-bin` flag set for error reporting
This would fix https://github.com/zigtools/zls/issues/869 and https://github.com/zigtools/zls/issues/868 in a maintainable way and bring the Zig developer experience closer to what mature programming languages like Rust or Go have.

SuperAuguste:
Fun lil hack :)

TODO:
- [ ] Cache results + multifile support (should be a single .eql in the diagnostic translation loop that checks errors versus current file pulling diagnostics)
- [ ] Run concurrently

This PR also updates tres and lsp.zig to the latest versions.

Closes #870

### DIFF
--- a/src/Server.zig
+++ b/src/Server.zig
@@ -20,6 +20,7 @@ const tracy = @import("tracy.zig");
 const uri_utils = @import("uri.zig");
 const diff = @import("diff.zig");
 const ComptimeInterpreter = @import("ComptimeInterpreter.zig");
+const CompilationDiagnostics = @import("CompilationDiagnostics.zig");
 
 const data = @import("data/data.zig");
 const snipped_data = @import("data/snippets.zig");
@@ -40,6 +41,8 @@ outgoing_messages: std.ArrayListUnmanaged([]const u8) = .{},
 recording_enabled: bool,
 replay_enabled: bool,
 offset_encoding: offsets.Encoding = .@"utf-16",
+compilation_diagnostics: CompilationDiagnostics,
+
 status: enum {
     /// the server has not received a `initialize` request
     uninitialized,
@@ -219,7 +222,10 @@ fn generateDiagnostics(server: *Server, handle: DocumentStore.Handle) !types.Pub
 
     if (server.config.enable_ast_check_diagnostics and tree.errors.len == 0) {
         try getAstCheckDiagnostics(server, handle, &diagnostics);
-        try getFullCompilationDiagnostics(server, handle, &diagnostics);
+    }
+
+    if (server.compilation_diagnostics.diagnostics.get(handle.uri)) |diags| {
+        try diagnostics.appendSlice(allocator, diags.items);
     }
 
     if (server.config.warn_style) {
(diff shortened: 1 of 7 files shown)


src/Server.zig
```diff
@@ -469,6 +470,101 @@ fn getAstCheckDiagnostics(
...
+        build_file_path,
+        "--pkg-end",
+        "--",
+        config.zig_exe_path.?,
+        directory_path,
+        zig_cache_root,
+        zig_global_cache_root,
+    };
+
+    const arg_length = standard_args.len + if (build_file.build_associated_config) |cfg| if (cfg.build_options) |options| options.len else 0 else 0;
```
Jarred-Sumner, line 518: this will need a way to pass:
- `-fno-emit-bin`
- custom `build.zig` arguments for diagnostics as a config option for zls.json

src/Server.zig
```diff
@@ -469,6 +470,101 @@ fn getAstCheckDiagnostics(
...
+            }
+        }
+    }
+
+    const zig_run_result = try std.ChildProcess.exec(.{
+        .allocator = arena_allocator,
+        .argv = args.items,
+    });
+
+    var line_iterator = std.mem.split(u8, zig_run_result.stderr, "\n");
```
Jarred-Sumner, line 534: the `\n` character could potentially could be part of a unicode character

Instead, you want to find the first non-ASCII character OR newline. If non-ASCII character, skip ahead by the length of the codepoint. If newline, process accordingly

Here is some example code: https://github.com/oven-sh/bun/blob/b86915c80819e61e6e1932d7802a4c31c73094e9/src/string_immutable.zig#L2919-L2966

src/Server.zig
```diff
@@ -469,6 +470,101 @@ fn getAstCheckDiagnostics(
...
+
+    var arena = std.heap.ArenaAllocator.init(allocator);
+    defer arena.deinit();
+    const arena_allocator = arena.allocator();
+
+    const build_file_path = try uri_utils.parse(arena_allocator, build_file.uri);
+    const directory_path = try std.fs.path.resolve(arena_allocator, &.{ build_file_path, "../" });
+
+    // TODO extract this option from `BuildAssociatedConfig.BuildOption`
+    const zig_cache_root: []const u8 = try std.fs.path.join(arena_allocator, &.{ directory_path, "zig-cache" });
```
Jarred-Sumner, line 496: it would be great if zig supported outputting errors as json

SuperAuguste: That'll never happen - Andrew hates JSON.

src/Server.zig
SuperAuguste: [Already handled](https://github.com/zigtools/zls/pull/871/files#diff-2b8e6e0157f1a82b0d7b30ff44364d62044cb3213f4ad91df0bf5bd72fb04983R237-R244) (diagnostics build runner, line 237-244)

src/Server.zig
SuperAuguste: I think I'll just end up using `std.unicode` like the rest of the codebase, tracked by #872

src/Server.zig
Vexu: There is a proposal for it https://github.com/ziglang/zig/issues/5456

SuperAuguste: Maybe there *is* hope :P (wonder if it'll ever get accepted though)

src/Server.zig
SuperAuguste: See conversation is Discord for why this mitigation is not necessary

Issue #867, opened 2022-12-30
Neovim lsp crashes when using latest zls


dmbfm:
### Zig Version

0.11.0-dev.1011+40ba4d4a8

### Zig Language Server Version

0.11.0-dev.97+c88562c

### Steps to Reproduce

Open any zig file in neovim with LSP enabled and 

### Expected Behavior

The servers should not crash.

### Actual Behavior

get this error message:

```
Error executing vim.schedule lua callback: ...w/Cellar/neovim/0.8.2/share/nvim/runtime/lua/vim/lsp.lua:1278: RPC[Error] code_name = unknown, code = 104, message = "InternalError"
stack traceback:
        [C]: in function 'assert'
        ...w/Cellar/neovim/0.8.2/share/nvim/runtime/lua/vim/lsp.lua:1278: in function ''
        vim/_editor.lua: in function <vim/_editor.lua:0>
```

LspLog outputs this:
```
[START][2022-12-30 16:32:01] LSP logging initiated
[ERROR][2022-12-30 16:32:01] .../vim/lsp/rpc.lua:733	"rpc"	"zls"	"stderr"	"info : ( main ): Enabled debug logging.\n\n"
[ERROR][2022-12-30 16:32:01] .../vim/lsp/rpc.lua:733	"rpc"	"zls"	"stderr"	"info : (config): Using zig executable /Users/daniel/.local/bin/zig\n"
[ERROR][2022-12-30 16:32:01] .../vim/lsp/rpc.lua:733	"rpc"	"zls"	"stderr"	"info : (config): Using zig lib path '/Users/daniel/.zigup/zig-macos-aarch64-0.11.0-dev.1011+40ba4d4a8/lib'\n"
[ERROR][2022-12-30 16:32:01] .../vim/lsp/rpc.lua:733	"rpc"	"zls"	"stderr"	"debug: (server): Took 1ms to process method initialize\n"
[ERROR][2022-12-30 16:32:01] .../vim/lsp/rpc.lua:733	"rpc"	"zls"	"stderr"	"error: (server): got error.InternalError while processing message!\n"

```
It seems that this happens after commit 61c0981294c52820d185afe41a0965a722c3e314. Previous commits seem to work correctly.

LordMZTE:
Can reproduce, I suspect its caused by this line: https://github.com/zigtools/zls/blame/master/src/Server.zig#L3049

mlugg:
[...]

PR #884, opened 2023-01-05
Fix crash when using nvim-lspconfig

mlugg:
Apparently, nvim reports its code action kinds using both the actual strings (e.g. "refactor.extract") and the enumeration names (e.g. "RefactorExtract"). I don't know why this is done - possibly an attempt at compatibility with non-compliant server implementations? Regardless, this was causing a crash on init (when tres tried to parse an initializaiton message), which is easily fixed by just supporting those enumeration values.

Resolves: #867


SuperAuguste (approved):
Wow, nice find! Yeah nvim is absolutely implementing the spec wrong (probably for compatibility reasons as you said) :(

PR #885, opened 2023-01-06
Verify minimum zig version at comptime

Techatrix:
This means no more strange compile errors when using 0.9.1 or 0.10.0.


SuperAuguste (approved):
Big brain

nektro:
it being in `build.zig` means it was already happening at compile time

PR #889, opened 2023-01-07
improve memory allocations

Techatrix:


src/main.zig
```diff
@@ -48,9 +48,18 @@ fn loop(
...
     const writer = buffered_writer.writer();
 
+    var arena = std.heap.ArenaAllocator.init(std.heap.page_allocator);
+    defer arena.deinit();
+
     while (true) {
-        var arena = std.heap.ArenaAllocator.init(server.allocator);
-        defer arena.deinit();
+        defer {
+            // Mom, can we have garbage collection at home?
```
erikarvstedt, line 56: This hits too close to home.
```suggestion
            // Mom, can we have garbage collection?
```

src/main.zig
```diff
@@ -48,9 +48,18 @@ fn loop(
     var buffered_writer = std.io.bufferedWriter(std_out);
     const writer = buffered_writer.writer();
 
+    var arena = std.heap.ArenaAllocator.init(std.heap.page_allocator);
```
nektro, line 51: switched from `server.allocator` to `std.heap.page_allocator` here

src/main.zig
```diff
@@ -338,7 +347,7 @@ const stack_frames = switch (zig_builtin.mode) {
 
 pub fn main() !void {
     var gpa_state = std.heap.GeneralPurposeAllocator(.{ .stack_trace_frames = stack_frames }){};
-    defer _ = gpa_state.deinit();
+    defer std.debug.assert(gpa_state.deinit());
```
nektro, line 350: you'll want this to be a `!gpa.deinit()`

src/main.zig
Techatrix: Thats intentional. Is that a problem?

SuperAuguste: Why is it intentional - would it be incorrect to use the server GPA? (Maybe like a metadata or perf thing?) :)

SuperAuguste: Your change sounds good to me! :)

SuperAuguste (approved):

PR #897, opened 2023-01-08
Nix: Improve debugging, support testing

erikarvstedt:
#### Copy of commit msg
- Export library deps to the nix build environment, so they are easily available inside `nix develop` shells.

- Set `XDG_CACHE_HOME` to an absolute path, because the tests require an absolute path for creating and accessing `build_runner.zig`.
  This allows running `zig build test` inside the dev shell.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -36,11 +36,16 @@
           nativeBuildInputs = [ zig ];
           dontConfigure = true;
           dontInstall = true;
+
+          inherit tres;
+          # Builder var names can't have dashes
+          known_folders = known-folders;
+
           buildPhase = ''
             mkdir -p $out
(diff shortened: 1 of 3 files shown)


Techatrix:
Can you update [config.json](https://github.com/zigtools/zls/blob/48288e02a9a5d0ba4ef5be1db7be6a64d37d2806/src/config_gen/config.json#L149) and then run `zig build gen` because [Config.zig](https://github.com/zigtools/zls/blob/master/src/Config.zig) should not be manually edited.

erikarvstedt:
Thanks! `zig build gen` is currently broken. First bad commit: 9badc745c5ea395ab85abde71b408d908990ba0d.

Techatrix:
I accidentally left some trailing commas in config.json.

SuperAuguste:
I let this go stale, that's on me; I decided to go with the latest / most common denominator solution from #975.

Issue #891, opened 2023-01-07
Can't jump to definition on first char of word


maxzhao:
### Zig Version

x86_64-0.11.0-dev.1240+24b4e643f

### Zig Language Server Version

2717b0fba19daba897f86dd6de4b8b590bfee516

### Steps to Reproduce

I use Nvim + Lsp,  When I locate on the first char of the fn name, then applied  go to definition action, it does't jump to fn source.  

### Expected Behavior

jump to right definition

### Actual Behavior

nothing happen

InKryption:
I'll attest this also seems to affect single-char identifiers (unable to jump to definition of a declaration of name `i`).

acristoffers:
From the small test I did, it looks like the code was written with VSCode in mind, in the sense that it considers the cursor to be inside a symbol only when it makes sense for a line-cursor. But in block-cursors, usually employed in terminal editors, it makes sense visually to consider the cursor to be in the word when it is right before it.

Those two cursors are on the same position in the file:

![image](https://user-images.githubusercontent.com/2191364/211194235-781d347c-cd25-4c26-aaf7-4e21eb75f7dd.png)

So for the block-cursor, it is pretty much in the symbol, while for the line-cursor it may be debatable.

This is also the reason why single-char identifiers may not work in VSCode. For me in Emacs it works only if I have the cursor after the identifier.

I tried to look at the code, but I'm not familiar enough to reason about it. It seems that the function `identifierFromPosition` (https://github.com/zigtools/zls/blob/master/src/Server.zig#L826) needs to be reworked, but then again the logic there seems right and it should get the symbol corretly. I can't see why the "look-forward" part of the code is not selecting the identifier (I expected the first loop to not run, but the second one should select the identifier). I'll see if I can find a way to debug it.

acristoffers:
I found out that the function `identifierFromPosition` is not really involved in the process. The actual culprid is `getPositionContext`, which actually does follow the line vs block logic I described in the last comment. When the cursor is right after the first char, it correctly returns `analysis.PositionContext{ .field_access }`, but returns `analysis.PositionContext{ .nothing }` when it is before the first char, as it cannot see it.

In the line https://github.com/zigtools/zls/blob/master/src/analysis.zig#L1532, we can just change

```zig
const line_loc = offsets.lineLocUntilIndex(text, doc_index);
```

to

```zig
const line_loc = offsets.lineLocUntilIndex(text, doc_index + 1);
```

[...]

PR #905, opened 2023-01-10
Fix lsp weird behaviour on block cursors (#891)

acristoffers:
I believe I got it right this time. This not only fixes the server for block-cursors but also correctly detects builtins, `foo.@"bar"`-style field access and single-letter identifiers.

For line-cursor, the following changes will be perceived: anytime the cursor touches an identifier, it is considered to be on that identifier.

I changed the test file to include a lot more positions for the same contexts we had already, always following the idea that if it touches it is in.

The change boils down to

```zig
    var new_index = doc_index;
    if (new_index < text.len and isSymbolChar(text[new_index])) {
        new_index += 1;
    } else if (new_index + 1 < text.len and text[new_index] == '@') {
        new_index += 2;
    }
```

which, in plain english, (I hope it) means: 
- if the next char is a valid identifier char, include it.
- otherwise, if the next char is @ and we still have a char after that available in the buffer, add both.

### DIFF
--- a/src/Server.zig
+++ b/src/Server.zig
@@ -837,21 +837,17 @@ pub fn identifierFromPosition(pos_index: usize, handle: DocumentStore.Handle) []
     if (pos_index + 1 >= handle.text.len) return "";
     var start_idx = pos_index;
 
-    while (start_idx > 0 and isSymbolChar(handle.text[start_idx - 1])) {
+    while (start_idx > 0 and analysis.isSymbolChar(handle.text[start_idx - 1])) {
         start_idx -= 1;
     }
 
     var end_idx = pos_index;
-    while (end_idx < handle.text.len and isSymbolChar(handle.text[end_idx])) {
+    while (end_idx < handle.text.len and analysis.isSymbolChar(handle.text[end_idx])) {
         end_idx += 1;
     }
 
     if (end_idx <= start_idx) return "";
     return handle.text[start_idx..end_idx];
-}
-
-fn isSymbolChar(char: u8) bool {
-    return std.ascii.isAlphanumeric(char) or char == '_';
 }
 
 fn gotoDefinitionSymbol(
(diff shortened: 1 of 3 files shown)


Techatrix (dismissed):
This change will cause auto completion to include a character to the right of the cursor which is nonsensical in my opinion. `getPositionContext` needs another parameter to control if an identifier to the right side should be included. This can then be disabled for completions and enabled for references, goto definition and hover.

Note that this change doesn't actually change the behavior of auto completions because there is not filtering based on the identifier at the end (yet).

acristoffers:
But that would mean that you are requesting a completion in the middle of a word or right before a `@`, which is also nonsensical, no?

Techatrix:
> But that would mean that you are requesting a completion in the middle of a word or right before a `@`, which is also nonsensical, no?

i don't really understand what you mean with this sentence (maybe i'm just bad at english).
Maybe this will clear things up:

auto completions:
`getPositionContext("std.<cursor>mem")` -> `"std."` 
`getPositionContext("std.me<cursor>m")` -> `"std.me"` 
`getPositionContext("std.<cursor>@"memcopy"")` -> `"std."`
`getPositionContext("std.@<cursor>"memcopy"")` -> `"std."`
`getPositionContext("std.@"<cursor>memcopy"")` -> `"std."`
`getPositionContext("std.@"me<cursor>mcopy"")` -> `"std.me"`
goto definition:
`getPositionContext("std.<cursor>mem")` -> `"std.mem"` 
`getPositionContext("std.me<cursor>m")` -> `"std.mem"` 
`getPositionContext("std.<cursor>@"memcopy"")` -> `"std.memcopy"` 
`getPositionContext("std.@<cursor>"memcopy"")` -> `"std.memcopy"`
`getPositionContext("std.@"<cursor>memcopy"")` -> `"std.memcopy"`
`getPositionContext("std.@"me<cursor>mcopy"")` -> `"std.memcopy"`

I think you will actually get something like `"std.@"memcopy"` or `"std.@"me"` but that can be fixed later.

acristoffers:
I mean that basically all your examples of autocompletion make no sense. Why would you request completition in the middle of a word?

```zig
getPositionContext("std.mem.e<cursor>ql")
```

What is supposed to happen in this case? Rigt now I'm getting this:

image

If I chose the first options the result is

image

and if I chose the second option it becomes

image

So no matter what I chose, the editor simply inserts text in the middle, so now I have to go to the end and delete it. So I don't see why it even makes sense to request completion in the middle of a word.

I may be missing the use case for this here, but it makes no sense to me to consider this case as something that needs to be taken into consideration because it doesn't make sense to start with.

Techatrix:
It can make sense to request completion in the middle of a word. 

The fact that zls will insert the entire function signature instead of removing the characters after the cursor is the part that makes no sense. This should be fixed but that a separate issue. The changes i request are just so that we don't need to redo this later.

SuperAuguste (approved):

PR #908, opened 2023-01-11
Partially fixes #907

### LINKED ISSUE #907: Completion does not replace existing text.
For clients that support the ApplyEdit capability, the code completion can return a range that allows the editor to replace text instead of only inserting new text. For example, in the case where the completion is requested as in:

```zig
const varA = 0;
const varB = 1;
const a = var<cursor>A;
```

the completion as of now will just insert the new text, leaving everything that comes after the cursor:

```zig
const a = varAA;
const a = varBA;
```

but it should replace the whole identifier with the new value:

```zig
const a = varA;
const a = varB;
```

acristoffers:
This partially fixes the completion issue, but leaves the following cases open:

- Function call:

```zig
const a = std.mem.e<cursor>ql(u8, "a", b);
```

will replace `eql` but leave `(u8, "a", b)` behind. We need a way to identify the function call and get its range.

- Structure initialization:

```zig
const a = A<cursor>Structure{.field = value};
```

will replace `AStructure` and leave `{.field = value}` behind.

Basically it only recognizes identifiers and cannot identify complex structures where it may make sense to replace the whole thing.

### DIFF
--- a/src/Server.zig
+++ b/src/Server.zig
@@ -2144,6 +2144,29 @@ fn completionHandler(server: *Server, request: types.CompletionParams) Error!?ty
 
     const completions = maybe_completions orelse return null;
 
+    // The cursor is in the middle of a word or before a @, so we can replace
+    // the remaining identifier with the completion instead of just inserting.
+    // TODO Identify function call/struct init and replace the whole thing.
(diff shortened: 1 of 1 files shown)


acristoffers:
This pull-request depends on https://github.com/zigtools/zls/pull/905 being accepted.

SuperAuguste (approved):

PR #910, opened 2023-01-11
Goto definition works when the cursor is at the start of the identifier

matklad:
Before, the code lexed only a prefix of the line up to cursor position.
Now, we lex the entire line, and do look at the token just after the
cursor.

This subtly changes sematncih of `getPostionContext`: now it is becomes
oblivious of the _exact_ position of the cursor and returns the whole
token at cursor's position.

I believe this is semantically right approach -- _most_ of the callsite
should not worry at all about such details. Something like completion
_might_ want to know more, but it's better to make that call site's
problem.

It might be the case that some existing code relies on the past
behavior. It's hard to tell though -- we don't have a lot of tests for
_features_, and changes to unit-tests don't explain if the changes are
meaningful for user-observable behavior or not.

In general, for LSP-shaped thing, I feel that the bulk of testing should
be focused on end-to-end behaviors....

Closes https://github.com/zigtools/zls/issues/891

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -163,6 +163,11 @@ pub fn build(b: *std.build.Builder) !void {
     test_step.dependOn(b.getInstallStep());
 
     var tests = b.addTest("tests/tests.zig");
+    tests.setFilter(b.option(
+        []const u8,
+        "test-filter",
+        "Skip tests that do not match filter",
+    ));
 
     if (coverage) {
         const src_dir = b.pathJoin(&.{ b.build_root, "src" });
--- a/src/analysis.zig
+++ b/src/analysis.zig
@@ -1528,9 +1528,15 @@ fn tokenLocAppend(prev: offsets.Loc, token: std.zig.Token) offsets.Loc {
     };
 }
 
+/// Given a byte index in a document (typically cursor offset), classify what kind of entity is at that index.
+///
+/// Classification is based on the lexical structure -- we fetch the line containin index, tokenize it,
+/// and look at the sequence of tokens just before the cursor. Due to the nice way zig is designed (only line
+/// comments, etc) lexing just a single line is always correct.
 pub fn getPositionContext(allocator: std.mem.Allocator, text: []const u8, doc_index: usize) !PositionContext {
-    const line_loc = offsets.lineLocUntilIndex(text, doc_index);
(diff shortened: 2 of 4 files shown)


matklad:
builds on top of #909

acristoffers:
Doesn't it break completion? Because now you are returning the entire identifier, but the completion code still expects only part of it, up to the point where it will start completing.

matklad:
It might... Guess I need to write some completion tests first, or do we already have some?

acristoffers:
I have two open pull-requests that touch on this (#891 and #908). Don't you want to take a look at them and build upon that? I say because they close the issue without breaking changes and with better testing (imo), and your changes may allow to better implement the completion thing I discuss with @Techatrix and partially implement in #908, but also requires more testing.

matklad:
I'd love to dig into completion issue first: I am somewhat sure (0.7) that fixing `getPostionContext` to not depend on specific cursor position too much is the more principled fix. I don't think we should worry about changing the tests for `getPostionContext` -- those are unit test, they don't exercise user-visible behaviors. We should add a bunch of completion tests those. 


My memory here is fuzzy, but it also seems that, unless we are doing something very facny, completion shouldn't depend on precise cursor position either -- LSP spec requires returnining all possible candidates, and its up to the client to filter those out based on cursor position...

acristoffers:
The location returned by `getPostionContext` is used in the completion code, so if you start returning the entire identifier, it will try to complete starting at the end of that identifier. When I modified `getPostionContext` to return one more char completion started behaving as if the cursor was one position more to the right. It may be easy to fix, though: change the places in the completion handler where it uses the location returned by `getPostionContext` to use a range that starts at same place as the loc from `getPostionContext` but ends wherever the cursor actually is.

matklad:
Ah, I see #905 now. I _think_ this PR should be fine wrt that behavior, but let me write the test first :)

matklad:
Ok, so we actually do have a bunch of completion tests which I've missed somehow:

https://github.com/zigtools/zls/blob/master/tests/lsp_features/completion.zig

(though, when I open that file zls crashes due to some UAF in build file URI). 

Looking at how the completion actually uses location and doing some manual testing, it seems we should be fine here.

leecannon (approved):

PR #911, opened 2023-01-11
fix use after free

matklad:
We free build_file_uri at https://github.com/zigtools/zls/compare/master...matklad:zls:fix-uaf?expand=1#diff-9cb2be04fa3f02d77f8aa62f4254392953ec06c5a1c4063db610575ab78a3343L690

I am not entirely sure whhat I am doing heer, but it gets me rid of a crash when working with zls iteslf.


leecannon (approved):

matklad:
As a side note, this whole thing looks rather scary, memory-managing individually allocated objects isn’t really zigs forte.

What if we just intern uri and never ever free them until dropping the server itself? There’s not that much Uris we’ll see anyway. Rust-analyzer does that for file paths (mostly to get rid of those pesky lifetimes), and it works great.

PR #939, opened 2023-01-24
Use new ast helper functions

Techatrix:
i wonder who added all these useful functions...


SuperAuguste:
Poggers. "I think whoever added these useful functions these is also allergic to force pushes" I remark, as I spam GitHub with 100s of force-pushes to keep my own PRs "clean." :P

SuperAuguste (approved):

PR #955, opened 2023-02-02
Work in Zig's breaking changes (build sys apis)

nullptrdevs:
Notes:
- I tried updating all occurrences of `std.build` however I had to keep the one on line 7, `pub fn build(b: *std.build.Builder) ....` otherwise older compilers choke on the missing `std.Build` and never reach the check.
             I also tried moving the comptime block outside the fn but `zig-linux-x86_64-0.10.0` and `zig-linux-x86_64-0.10.1` segfault..
- Depending on how soon `std.build` and related get removed, types in  `build_runner.zig` would need to be conditionally aliased.


SuperAuguste (approved):
this LGTM, thanks npd!

PR #958, opened 2023-02-03
Skip incomplete fn_protos when generating folding ranges

nullptrdevs:
fuzz this :smiley:


SuperAuguste (approved):
Thank you so much for fixing this!!! Also helped me net a bug in the fuzzer which is a big W. 🫡🫡

PR #971, opened 2023-02-05
Build fixes for latest master

kcbanner:
This doesn't migrate to the package manager way (with dependencies), but just uses the existing submodules to get it compiling again with latest master.


leecannon:
The build runner will need to be modified to support both the old and new way as well as updating `min_zig`.

SuperAuguste:
I believe that this has been superseded by #976. Thanks for your contribution!

PR #982, opened 2023-02-07
Switch to Myers diff

### LINKED ISSUE #979: formatting file with zls causes infinite loop in memory allocator
### Zig Version

0.11.0-dev.1393+38eebf3c4

384f227cb752553f440f2d256e68858b65fb13fc

### Steps to Reproduce

When using zls to format files, it seems to infinite loop on certain input. Generally it seems to be more likely to happen on longer files. `zig fmt` works as expected.

[File that reproduces the issue](https://gist.github.com/Jarred-Sumner/f6eb3d35d41bbac8de84fe0f58336766) (too long to fit in an issue)

### Expected Behavior

Formats successfully, like `zig fmt`

### Actual Behavior

It infinite loops until it uses enough memory for the OOM killer to kill it. This type of issue is often caused by looping over or duplicating a slice that is undefined, in which cause the `len` field is something absurdly large (due to undefined memory)

image

SuperAuguste:
Fixes #979;

I originally ripped off gopls' Myers Difference Algorithm implementation, but realized that it did not fit our usecase, so I instead wrote about 2000 lines of code for a Zig DiffMatchPatch impl w/ Travis, which is what this PR uses! :)

Confirmed working by adding a new, correct diff patch implementation (which we should probably start using in the future) to `diff.zig` and getting tests running.


SuperAuguste:
Extremely weird CI failure...

SuperAuguste:
Sadly, while this drastically improves performance, this messes up the cursor position on format by moving it to the start of the line. Looking for a solution / better algorithm we could use :)

leecannon (approved):

PR #989, opened 2023-02-09
Ignore common editor file

matklad:
In theory, this is everyone's own business, and they should add personal excludes to `.git/info/exclude` or to global gitignore. In practice, it's just simpler to add this stuff directly to project's .gitignore


SuperAuguste:
I think this should be left to the global `.gitignore` as you said.

PR #990, opened 2023-02-09
Fix missing nodes in outline

### LINKED ISSUE #986: Outline is broken with nested items
For the following code: 

```zig
const Foo = struct {
    const Self = @This();
    pub fn foo() !Self {}
    const Bar = struct {};
};
```

the outline (`textDocument/documentSymbol`) contains neither `foo` function nor `Bar` struct. If I comment out either `Self` or `Bar` this  works though.

matklad:
Builds on top of #988.

Closes #986

### DIFF
--- a/src/analysis.zig
+++ b/src/analysis.zig
@@ -1851,6 +1851,7 @@ fn addOutlineNodes(allocator: std.mem.Allocator, tree: Ast, child: Ast.Node.Inde
         .error_set_decl,
         => return,
         .container_decl,
+        .container_decl_trailing,
         .container_decl_arg,
         .container_decl_arg_trailing,
         .container_decl_two,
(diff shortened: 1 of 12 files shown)


src/analysis.zig
```diff
@@ -1851,6 +1851,7 @@ fn addOutlineNodes(allocator: std.mem.Allocator, tree: Ast, child: Ast.Node.Inde
         .error_set_decl,
         => return,
         .container_decl,
+        .container_decl_trailing,
```
matklad, line 1854: debating whether I should replace else bellow with an explicit list of tags

SuperAuguste: I don't think that's necessary here :)

SuperAuguste:
Please rebase this and #988 :)

matklad:
rebased!

SuperAuguste (approved):

PR #999, opened 2023-02-14
sync build.zig with latest std.build API changes

travisstaloch:
fixes build.zig now that b.path_root changed from a `[]const u8` to a `std.build.Cache.Directory`.


SuperAuguste (approved):
A breaking Zig change passing CI on the first try; this must be a good omen. LGTM! :)

PR #1000, opened 2023-02-14
Embed AstGen into ZLS

Techatrix:
This removes the dependency on Zig for showing `ast-check` diagnostics.
This is also the starting pointer for integrating AstGen based analysis into ZLS.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,7 @@ The following options are currently available.
 | `include_at_in_builtins` | `bool` | `false` | Whether the @ sign should be part of the completion of builtins |
 | `skip_std_references` | `bool` | `false` | When true, skips searching for references in std. Improves lookup speed for functions in user's code. Renaming and go-to-definition will continue to work as is |
 | `max_detail_length` | `usize` | `1048576` | The detail field of completions is truncated to be no longer than this (in bytes) |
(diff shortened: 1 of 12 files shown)


g-w1:
I believe it should be `LICENCE -> LICENSE` :)

truemedian:
zls master is only intended to really work with zig master, they update in lock-step. zls gets pinned in time for the zig releases to fix this exact problem.

leecannon:
We do try to keep the runtime zig version requirements as wide as possible by making sure the build runner can run on as many versions as possible but we only do this a _best effort_.

As @nullptrdevs points out the multi-object for is a breaking change to the language (`for (a) |*item, i|` is no longer valid) this means the embeded ast-check will find valid 0.9.1/0.10.1 code as invalid which means that with this change zls will only support zig versions after the multi-object change (which is a good excuse to simplify the build_runner to remove all the compatibility stuff)

SuperAuguste:
On its own, this PR does not make sense, but in the wider picture Techatrix painted in the Discord of using ZIR for analysis and eventually comptime interpreting, it absolutely does. I believe that for now, it would make the most sense to have `prefer_ast_check_as_child_process` set to `true` by default.

SuperAuguste (approved):
I've been thinking this PR through and, at first I was a bit skeptical, but YOLO.

I guess we're getting closer and closer to becoming `zig-analyzer` huh :P

Issue #965, opened 2023-02-04
Better errors when parsing zls.json config file


Ultra-Code:
### Zig Version

0.11.0-dev.1568+c9b957c93

### Zig Language Server Version

0.11.0-dev.160+384f227

### Steps to Reproduce

Using this configuration
```json
{
    "enable_snippets": true,
    "enable_ast_check_diagnostics": true,
    "enable_autofix": true,
    "enable_import_embedfile_argument_completions": true,
    "warn_style": true,
    "use_comptime_interpreter": true,
    "enable_semantic_tokens": false,
    "enable_inlay_hints": true,
    "operator_completions": true,
    "include_at_in_builtins": false,
}

```

open neovim with lsp-config installed.
Run :LspLog
you would have this output
```
[START][2023-02-04 16:51:49] LSP logging initiated
[ERROR][2023-02-04 16:51:49] .../vim/lsp/rpc.lua:734 "rpc" "zls" "stderr"
"warning: (config): Error while parsing configuration file: error.InvalidValueBegin
info : ( main ): No config file zls.json found.
info : (config): Using zig executable /usr/bin/zig
info : (config): Using zig lib path '/firefox/home/.local/zig/zig-linux-x86_64-0.11.0-dev.1568+c9b957c93/lib'
info : (server): client is 'Neovim-0.9.0'
info : (server): zls initializing
info : (server): Server.ClientCapabilities{ .supports_snippets = true, .supports_apply_edits = true, .supports_will_save = true, .supports_will_save_wait_until = true, .supports_publish_diagnostics = true, .supports_code_action_fixall = false, .hover_supports_md = true, .completion_doc_supports_md = true, .label_details_support = true, .supports_configuration = true }
info : (server): Using offset encoding: utf-16"
[ERROR][2023-02-04 16:52:01] .../vim/lsp/rpc.lua:734 "rpc" "zls" "stderr"
"info : (server): Setting configuration...
info : (config): Using zig executable /usr/bin/zig"
```

This behaviour is also experienced with 
`zls --show-config-path`
```
warning: (config): Error while parsing configuration file: error.InvalidValueBegin
/firefox/home/.config/zls.json
```

I tested this with both the nightly version of `Neovim-0.9.0` and `Neovim-0.8.3` on `archlinux`

### Expected Behavior

The `zls.json` configuration file is read by zls successfully and used. That is, there should be no error while parsing the configuration file and the configuration should be respected and override the default configurations.

### Actual Behavior

[...]

PR #1002, opened 2023-02-14
improve error message when parsing zls.json

Techatrix:
this could still be improved but still more helpful than the status quo.
#965 could be considered fixed


SuperAuguste (approved):
Huh I didn't even know `findLineColumn` was a thing. Good stuff!

PR #1005, opened 2023-02-15
Fix build runner for latest changes

leecannon:
Now caching is in the std lib.

### DIFF
--- a/src/special/build_runner.zig
+++ b/src/special/build_runner.zig
@@ -2,10 +2,22 @@ const root = @import("@build@");
 const std = @import("std");
 const log = std.log;
 const process = std.process;
-const Builder = std.build.Builder;
-const InstallArtifactStep = std.build.InstallArtifactStep;
-const LibExeObjStep = std.build.LibExeObjStep;
-const OptionsStep = std.build.OptionsStep;
+
+// Zig 0.11.0-dev.1524+
+const Build = if (@hasDecl(std, "Build")) std.Build else std.build;
+
+// Zig 0.11.0-dev.1524+
+const Builder = if (@hasDecl(std, "Build")) std.Build else std.build.Builder;
+
+// Zig 0.11.0-dev.1637+
+const Cache = if (@hasDecl(Build, "Cache")) std.Build.Cache else void;
+const has_cache = Cache != void;
+
+// Zig 0.11.0-dev.1524+
+const CompileStep = if (@hasDecl(Build, "CompileStep")) Build.CompileStep else Build.LibExeObjStep;
+
+const InstallArtifactStep = Build.InstallArtifactStep;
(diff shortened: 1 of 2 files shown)


src/special/build_runner.zig
```diff
@@ -2,10 +2,22 @@ const root = @import("@build@");
...
-const Builder = std.build.Builder;
-const InstallArtifactStep = std.build.InstallArtifactStep;
-const LibExeObjStep = std.build.LibExeObjStep;
-const OptionsStep = std.build.OptionsStep;
+
+// Zig 0.11.0-dev.1524+
+const Build = if (@hasDecl(std, "Build")) std.Build else std.build;
+
+// Zig 0.11.0-dev.1524+
+const Builder = if (@hasDecl(std, "Build")) std.Build else std.build.Builder;
```
InKryption, line 10: If compatibility is the goal, wouldn't it be better to just continue to use `std.build` until it's deprecated? Otherwise, when it is finally removed, this will compile and possibly be left unchanged for a while longer than it should.

InKryption: I don't think trying to maintain that type of backwards compatibility at the source level here is sustainable; there's been a lot of breaking changes, and there's going to be a lot more breaking changes coming very soon (e.g. parallel builds & metadata changes).

A more practical solution may be to have build_runner.zig files for multiple versions, and use one based on a query to the zig compiler.

InKryption: That, or have it as a config option, to avoid having to invoke the executable.

leecannon: I have two issues with having multiple build_runner files for different versions:

- Adding new features that required runner changes (for example #949 #686 #675)  would require a lot of duplicated changes
- With the current method _any_ zig commit after 0.9.0 works using a single file however if using individual files and **wanting the same compatibility** every commit that breaks the build runner would need a new file. For example their would be one file with the `std.build` -> `std.Build` changes then another file with the `Cache` change, etc. So for the same end user visible compatibility we would already have a minimum of 6 (off the top of my head) files.

SuperAuguste: Thanks for supporting multiple versions Lee and for bringing this up InK. As Zig evolves the build runner is only going to get more convoluted. We should probably setup a policy for when to stop supporting Zig versions in the near to far future :)

leecannon: While making the change in the CI of 0.10.0 to 0.10.1 I did think about dropping 0.9.1 but atleast at the moment the burden isn't that big.
However with the large number of changes coming through leading up to 0.11, it might be worth dropping support for old versions once 0.11.1 is released.

SuperAuguste (approved):
LGTM

PR #1006, opened 2023-02-15
Fix handling of invalid tokens in getPositionContext

nullptrdevs:
Fuzz this a few times for good measure .. or/and test with the known principal/`pub const a = @import(""\<request completion here>);`

Cheers,


SuperAuguste (approved):
Fuzzing seems to have passed so this LGTM! Thanks! :)

PR #1011, opened 2023-02-20
add multi object for loop support

Techatrix:
faster than [ziglang.org](https://www.youtube.com/watch?v=dQw4w9WgXcQ)


SuperAuguste:
Weird error with inlines occurring now but that's pre-existing so this seems to be good to go! tysm Techatrix!!

SuperAuguste (approved):

PR #1013, opened 2023-02-20
build: add PIE (Position Independent Executable) option

orhun:
This PR adds a `-Dpie=bool` option to `build.zig` for enabling/disabling [PIE](https://en.wikipedia.org/wiki/Position-independent_code) for `zig build` command.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -26,6 +26,7 @@ pub fn build(b: *std.build.Builder) !void {
     const exe_options = b.addOptions();
     exe.addOptions("build_options", exe_options);
 
+    const pie = b.option(bool, "pie", "Build a Position Independent Executable") orelse false;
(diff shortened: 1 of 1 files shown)


build.zig
```diff
@@ -42,6 +43,12 @@ pub fn build(b: *std.build.Builder) !void {
         b.option(std.log.Level, "log_level", "The Log Level to be used.") orelse .info,
     );
 
+    exe_options.addOption(
+        bool,
+        "pie",
+        pie,
+    );
+
```
nektro, line 51: no reason for this part to be here but otherwise good

orhun: you're right! PR is updated.

SuperAuguste (approved):

PR #1019, opened 2023-02-22
support new module cli arguments

leecannon:
Changes the build runner arguments to match new module cli arguments: https://github.com/ziglang/zig/pull/14664

### DIFF
--- a/.github/workflows/build_runner.yml
+++ b/.github/workflows/build_runner.yml
@@ -3,9 +3,11 @@ name: BuildRunner
 on:
   push:
     paths:
+      - ".github/workflows/build_runner.yml"
       - "src/special/build_runner.zig"
   pull_request:
     paths:
+      - ".github/workflows/build_runner.yml"
       - "src/special/build_runner.zig"
   schedule:
     - cron: '0 0 * * *'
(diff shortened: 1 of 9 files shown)


leecannon:
oh yea :(

leecannon:
So this bring up a similar question as [here](https://github.com/zigtools/zls/pull/1000#issuecomment-1436794639), how seriously do we take ensuring that zls master can be used with (not compiled with) previous versions of zig.

Do we just recommend people to use tagged commits/releases for the matching zig release or do we keep bending over backward?

In this case its not that bad as we already run `zig env` so we can just capture the version and do some semantic comparison to use the right module arguments but im thinking its time we start thinking about what we want to do going forward as features like #1000 are going to have to be selectively disabled based on zig version and the version we check for is going to change every time there is a breaking language change that affects ast-gen.

leecannon:
The latest commit grabs the runtime zig version from `zig env` and holds onto it, this is then used to send the correct module cli arguments to the buildrunner.

thepheer:
Seems like `0.11.0-dev.1782+b52be973d` is not the exact version where `--pkg-*` paremeters were removed?
The PR doesn't work on `0.11.0-dev.1718+2737dce84`.

Those who want to use the PR already, update your Zig again just in case.

leecannon:
@thepheer fixed

SuperAuguste (approved):
#1020 is still pending but this LGTM.

PR #1026, opened 2023-02-25
make `diff.edits` memory safe

Techatrix:
`diff.edits` would leak if you haven't used a `ArenaAllocator`.
I've also adjusted the test to be slightly faster by changing the iteration count and file length.

### DIFF
--- a/tests/utility/diff.zig
+++ b/tests/utility/diff.zig
@@ -1,30 +1,37 @@
 const std = @import("std");
 const zls = @import("zls");
 
-const allocator = std.testing.allocator;
-
 fn gen(alloc: std.mem.Allocator, rand: std.rand.Random) ![]const u8 {
-    var buffer = try alloc.alloc(u8, rand.intRangeAtMost(usize, 16, 1024));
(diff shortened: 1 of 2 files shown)


tests/utility/diff.zig
```diff
@@ -1,30 +1,37 @@
...
 
 test "diff - random" {
-    var arena = std.heap.ArenaAllocator.init(allocator);
-    defer arena.deinit();
-
-    var rand = std.rand.DefaultPrng.init(0);
-
-    var index: usize = 0;
+    const allocator = std.testing.allocator;
+    try std.testing.checkAllAllocationFailures(allocator, testDiff, .{ 0, .@"utf-8" });
```
Techatrix, line 12: Mom, can we have borrow checker?
No, we already have borrow checker at home.
at home:
`std.testing.checkAllAllocationFailures`

SuperAuguste (approved):
LGTM

We might have to take https://github.com/travisstaloch/diff-zimilar for a spin in the future, but this is a great solution for now :)

PR #1031, opened 2023-03-04
Branching type resolution

SuperAuguste:
Finally, our mortal enemy has been defeated over the course of a stream. [(shameless plug)](https://twitch.tv/SuperAuguste)

image

- [x] Resolve branching types in completions
- [x] Resolve in goto def (multi Location result)
- [x] Resolve in hover
- ~~Resolve in signatureHelp~~ (it's so hard 😭)

### DIFF
--- a/src/Server.zig
+++ b/src/Server.zig
@@ -545,6 +545,7 @@ pub fn typeToCompletion(
     list: *std.ArrayListUnmanaged(types.CompletionItem),
     field_access: analysis.FieldAccessReturn,
     orig_handle: *const DocumentStore.Handle,
+    either_descriptor: ?[]const u8,
 ) error{OutOfMemory}!void {
     var allocator = server.arena.allocator();
 
@@ -587,6 +588,7 @@ pub fn typeToCompletion(
                 orig_handle,
                 type_handle.type.is_type_val,
                 null,
+                either_descriptor,
             );
         },
         .other => |n| try server.nodeToCompletion(
@@ -596,6 +598,7 @@ pub fn typeToCompletion(
             orig_handle,
             type_handle.type.is_type_val,
             null,
+            either_descriptor,
         ),
         .primitive, .array_index => {},
         .@"comptime" => |co| try analyser.completions.dotCompletions(
@@ -608,7 +611,7 @@ pub fn typeToCompletion(
         ),
         .either => |bruh| {
             for (bruh) |a|
-                try server.typeToCompletion(list, .{ .original = a }, orig_handle);
+                try server.typeToCompletion(list, .{ .original = a.type_with_handle }, orig_handle, a.descriptor);
         },
     }
 }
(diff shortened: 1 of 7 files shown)


leecannon:
```zig
const std = @import("std");

pub fn main() !void {
    const t = std.os.system.dup(1);
    _ = t;
}
```
Hover over `dup`:
```
error: (server): got error. error while handling textDocument/hover
/home/lee/src/zls/src/analysis.zig:1142:9: 0x45eca9 in getAllTypesWithHandlesArrayList (zls)
        return switch (ty.type.data) {
        ^
/home/lee/src/zls/src/analysis.zig:1137:9: 0x45edea in getAllTypesWithHandles (zls)
        try ty.getAllTypesWithHandlesArrayList(arena, &all_types);
        ^
/home/lee/src/zls/src/Server.zig:1190:40: 0x4f71d7 in getSymbolFieldAccesses (zls)
        const container_handle_nodes = try container_handle.getAllTypesWithHandles(server.arena.allocator());
                                       ^
/home/lee/src/zls/src/Server.zig:1244:20: 0x473b3a in hoverDefinitionFieldAccess (zls)
    const decls = (try server.getSymbolFieldAccesses(handle, source_index, loc)) orelse return null;
                   ^
/home/lee/src/zls/src/Server.zig:2469:32: 0x3eff9c in hoverHandler (zls)
        .field_access => |loc| try server.hoverDefinitionFieldAccess(handle, source_index, loc),
                               ^
debug: (server): Took 232ms to process method textDocument/hover
[Error - 18:07:20] Request textDocument/hover failed.
  Message: InternalError
  Code: 108
```

SuperAuguste:
Hmm I can't seem to reproduce this...

leecannon:
Looks like this might be a miscompilation, its fixed by removing return from `getAllTypesWithHandlesArrayList`:
```zig
pub fn getAllTypesWithHandlesArrayList(ty: TypeWithHandle, arena: std.mem.Allocator, all_types: *std.ArrayListUnmanaged(TypeWithHandle)) !void {
    switch (ty.type.data) {
        .pointer, .slice, .error_union, .other, .primitive => try all_types.append(arena, ty),
        .either => |e| for (e) |i| try i.type_with_handle.getAllTypesWithHandlesArrayList(arena, all_types),
        .array_index, .@"comptime" => {
            // TODO
        },
    }
}
```

SuperAuguste:
wtf... nice find Lee! Also `getAllTypesWithHandlesArrayList` should just have an `else` instead of that first condition :P I'll fix it now!

leecannon:
That works for me now, I absolutely love this feature!

Something not required but would be nice, is showing the source of each declaration as while messing with this I immediately hit one where the types are different (although *mostly* compatible):
image

Jarred-Sumner:
very exciting

SuperAuguste:
> That works for me now, I absolutely love this feature!
> 
> Something not required but would be nice, is showing the source of each declaration as while messing with this I immediately hit one where the types are different (although _mostly_ compatible): image

Tried implementing this but it's a bit of a hackfest, I think merging this now and then adding this later makes sense. I might take a jab at rewriting some analysis code so it's a bit less of a rats nest and try again. :)

leecannon (approved):

Issue #1024, opened 2023-02-23
Builtin dependency system will break nix flake.


acristoffers:
### Zig Version

0.11.0-dev.1782+b52be973d

### Zig Language Server Version

1a7b0a7a70e19070445db98be1393b6cef3b0052

### Steps to Reproduce

When #1017 gets accepted, it will break nix flakes as nix does not allow the build tool to download files during the build phase (or any other phase). To fix it, we can change the flake.nix file to:

```nix
{
  inputs = {
    nixpkgs.url = "github:NixOS/nixpkgs/nixpkgs-unstable";

    zig-overlay.url = "github:mitchellh/zig-overlay";
    zig-overlay.inputs.nixpkgs.follows = "nixpkgs";

    gitignore.url = "github:hercules-ci/gitignore.nix";
    gitignore.inputs.nixpkgs.follows = "nixpkgs";

    flake-utils.url = "github:numtide/flake-utils";

    known-folders.url = "https://github.com/ziglibs/known-folders/archive/53fe3b676f32e59d46f4fd201d7ab200e5f6cb98.tar.gz";
    known-folders.flake = false;

    tres.url = "https://github.com/ziglibs/tres/archive/d8b0c24a945da02fffdae731edd1903c6889e73c.tar.gz";
    tres.flake = false;

    diffz.url = "https://github.com/ziglibs/diffz/archive/efc91679b000a2d7f86fb40930f0a95a0d349bff.tar.gz";
    diffz.flake = false;
  };

  outputs = { self, nixpkgs, zig-overlay, gitignore, flake-utils, known-folders, tres, diffz }:
    let
      systems = [ "x86_64-linux" "aarch64-linux" "x86_64-darwin" "aarch64-darwin" ];
      inherit (gitignore.lib) gitignoreSource;
    in
    flake-utils.lib.eachSystem systems (system:
      let
        pkgs = nixpkgs.legacyPackages.${system};
        zig = zig-overlay.packages.${system}.master;
      in
      rec {
        packages.default = packages.zls;
        packages.zls = pkgs.stdenvNoCC.mkDerivation {
          name = "zls";
          version = "master";
          src = gitignoreSource ./.;
          nativeBuildInputs = [ zig ];
          dontConfigure = true;
          dontInstall = true;
          buildPhase = ''
            mkdir -p $out
            mkdir -p .cache/{p,z,tmp}
[...]

PR #1017, opened 2023-02-22
update zig submodules to new builtin dependency system

nektro:
package manager was added in `0.11.0-dev.1570+693b12f8e` so updating the minimum version was not necessary

PR #1032, opened 2023-03-04
Fixes nix build.

acristoffers:
As I said in #1024, #1017 broke nix. This fixes it, but every time the zon file gets updated we have to manually copy the URLs from it into the nix file.


acristoffers:
Merged and updated the urls.

SuperAuguste (approved):

PR #1037, opened 2023-03-06
send response on shutdown request

Techatrix:
the shutdown request still needs a response even though it is meaningless


nullptrdevs:
My bad.

PR #1042, opened 2023-03-08
Fixes nix standard derivation.

acristoffers:
The standard derivation has been broken for quite some time now, but I didn't use it, so I didn't care to fix it. Today I cared. It now builds and is updated for the new packaging system.

I have a feeling this one will require less maintenance than the flake, as we only have to manually update the revision and hashes for zig and gitignore, and both should brake zls less often than its direct dependencies.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -4,19 +4,38 @@
 
 let
   zig-overlay = pkgs.fetchFromGitHub {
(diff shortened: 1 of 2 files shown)


SuperAuguste (approved):
this is pretty cool, tysm!

PR #1055, opened 2023-03-12
Add some more basic overflow checks

SuperAuguste:
https://fuzzing.zigtools.org/group/c00b156c6e01256103efebfed586e30d and then some


Techatrix:
Why use saturating subtraction when you check for zero?

SuperAuguste:
Oh my god you’re totally right I missed that completely 🤦‍♂️

PR #1059, opened 2023-03-13
exclude default cimport declarations from completions

Techatrix:
Previously completions on cImport would be filled with symbols like `__GNUC__` `__builtin_exp` `_fastcall`. I doubt many users are intrested in them so I thinks its best if we ignore them.

### DIFF
--- a/src/features/completions.zig
+++ b/src/features/completions.zig
@@ -343,7 +343,7 @@ fn declToCompletion(context: DeclToCompletionContext, decl_handle: Analyser.Decl
             .{ "WINNT", {} },
             .{ "WIN64", {} },
         });
-        if (exclusions.has("name")) return;
+        if (exclusions.has(name)) return;
     }
 
     switch (decl_handle.decl.*) {
--- a/src/features/completions.zig
+++ b/src/features/completions.zig
@@ -330,7 +330,23 @@ fn declToCompletion(context: DeclToCompletionContext, decl_handle: Analyser.Decl
     const tree = decl_handle.handle.tree;
     const decl = decl_handle.decl.*;
 
-    switch (decl) {
+    const is_cimport = std.mem.eql(u8, std.fs.path.basename(decl_handle.handle.uri), "cimport.zig");
+    if (is_cimport) {
+        const name = tree.tokenSlice(decl_handle.nameToken());
+        if (std.mem.startsWith(u8, name, "_")) return;
+        // TODO figuring out which declarations should be excluded could be made more complete and accurate
+        // by translating an empty file to acquire all exclusions
+        const exclusions = std.ComptimeStringMap(void, .{
+            .{ "linux", {} },
+            .{ "unix", {} },
+            .{ "WIN32", {} },
+            .{ "WINNT", {} },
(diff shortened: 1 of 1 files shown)


leecannon:
This looks really nice!

But I think this should be have a config option (default to hiding) as it is possible that someone wants these symbols.

src/Server.zig
```diff
@@ -1327,6 +1327,15 @@ pub fn declToCompletion(context: DeclToCompletionContext, decl_handle: analysis.
     var allocator = context.server.arena.allocator();
 
     const tree = decl_handle.handle.tree;
+    const name = tree.tokenSlice(decl_handle.nameToken());
+
+    const is_cimport = std.mem.eql(u8, std.fs.path.basename(decl_handle.handle.uri), "cimport.zig");
+    if (is_cimport) {
+        if (std.mem.startsWith(u8, name, "_")) return;
+        if (std.mem.eql(u8, name, "linux")) return;
+        if (std.mem.eql(u8, name, "unix")) return;
```
leecannon, line 1336: There are other symbols like these for windows:
```zig
pub const WIN32 = @as(c_int, 1);
pub const WINNT = @as(c_int, 1);
pub const WIN64 = @as(c_int, 1);
```

Looks like there are none without an underscore at the beginning on macos, but there might be more symbols like this on other targets.

InKryption: One way to solve this could be to translate-c an empty file for the target, and extract the identifiers that are present by default that way.

Techatrix: > One way to solve this could be to translate-c an empty file for the target, and extract the identifiers that are present by default that way.

Definitely doable, but I don't really think its worth the effort right now for such a small quality of life improvement. I will add a TODO on this idea.

Techatrix:
> But I think this should be have a config option (default to hiding) as it is possible that someone wants these symbols.

Do you think there are enough users of these symbols that would make it worth adding a new config option? Trying to keep the number of config option as low as possible would be nice in my opinion. And this seems like a very rare use-case. Just a suggestion, If you want a config option, I will happily add one.

src/features/completions.zig
```diff
@@ -330,7 +330,23 @@ fn declToCompletion(context: DeclToCompletionContext, decl_handle: Analyser.Decl
...
+        // TODO figuring out which declarations should be excluded could be made more complete and accurate
+        // by translating an empty file to acquire all exclusions
+        const exclusions = std.ComptimeStringMap(void, .{
+            .{ "linux", {} },
+            .{ "unix", {} },
+            .{ "WIN32", {} },
+            .{ "WINNT", {} },
+            .{ "WIN64", {} },
+        });
+        if (exclusions.has("name")) return;
```
leecannon, line 346:
```suggestion
        if (exclusions.has(name)) return;
```

leecannon:
I'm fine with not having a config option.

src/features/completions.zig
Techatrix: how did I manage to get this wrong just by copy pasting code...

leecannon (approved):

Issue #338, opened 2021-06-01
Allow limiting semantic token highlighting to identifiers


Vexu:
Semantic token highlighting is nice and all but using a syntax file like the one provided by [vscode-zig](https://github.com/ziglang/vscode-zig) gives much more control. Limiting semantic token highlighting to only identifiers gives the best of both worlds. I think the C/C++ language server does something similar since the result looks much more like my what I was able to achieve with a quick patch.

<details>
<summary>Zls master version</summary>
<img src="https://user-images.githubusercontent.com/15308111/120354046-d5439f00-c30a-11eb-9a24-2bcf27821618.png"></img>
</details>

<details>
<summary>With patch applied</summary>
<img src="https://user-images.githubusercontent.com/15308111/120354053-d674cc00-c30a-11eb-87fb-db5ab16b6d59.png"></img>
</details>

<details>
<summary>Quick patch to enable this behavior, not sure how correct this is</summary>

```diff
diff --git a/src/semantic_tokens.zig b/src/semantic_tokens.zig
index eae2a54..8c813f3 100644
--- a/src/semantic_tokens.zig
+++ b/src/semantic_tokens.zig
@@ -66,6 +66,18 @@ const Builder = struct {
     }
 
     fn add(self: *Builder, token: ast.TokenIndex, token_type: TokenType, token_modifiers: TokenModifiers) !void {
+        switch (token_type) {
+            .type,
+            .parameter,
+            .variable,
+            .enumMember,
+            .field,
+            .errorTag,
+            .function,
+            .label, => {},
+            else => return,
+        }
+
         const tree = self.handle.tree;
         const starts = tree.tokens.items(.start);
         const tags = tree.tokens.items(.tag);
@@ -159,6 +171,18 @@ const Builder = struct {
     }
 
     fn addDirect(self: *Builder, tok_type: TokenType, tok_mod: TokenModifiers, start: usize, length: usize) !void {
+        switch (tok_type) {
+            .type,
+            .parameter,
+            .variable,
+            .enumMember,
+            .field,
+            .errorTag,
+            .function,
+            .label, => {},
+            else => return,
+        }
+
         const delta = offsets.tokenRelativeLocation(
             self.handle.tree,
             self.previous_position,
```
</details>

SuperAuguste:
Quite the interesting idea, perhaps worthy of a configuration option :)

Techatrix:
implemented by #1064

PR #1064, opened 2023-03-14
make semantic token configuration an enum

Vexu:
Implements #338

I'm not sure how the config gen stuff works so any help on that would be appreciated.

Vscode companion pr https://github.com/ziglang/vscode-zig/pull/90

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ The following options are currently available.
 | `enable_ast_check_diagnostics` | `bool` | `true` | Whether to enable ast-check diagnostics |
 | `enable_autofix` | `bool` | `true` | Whether to automatically fix errors on save. Currently supports adding and removing discards. |
 | `enable_import_embedfile_argument_completions` | `bool` | `true` | Whether to enable import/embedFile argument completions |
-| `enable_semantic_tokens` | `bool` | `true` | Enables semantic token support when the client also supports it |
+| `semantic_tokens` | `enum` | `.full` | Set level of semantic tokens. Partial only includes information that requires semantic analysis. |
 | `enable_inlay_hints` | `bool` | `true` | Enables inlay hint support when the client also supports it |
 | `inlay_hints_show_builtin` | `bool` | `true` | Enable inlay hints for builtin functions |
 | `inlay_hints_exclude_single_argument` | `bool` | `true` | Don't show inlay hints for single argument calls |
--- a/schema.json
+++ b/schema.json
@@ -24,10 +24,15 @@
             "type": "boolean",
             "default": "true"
         },
-        "enable_semantic_tokens": {
-            "description": "Enables semantic token support when the client also supports it",
-            "type": "boolean",
-            "default": "true"
+        "semantic_tokens": {
+            "description": "Set level of semantic tokens. Partial only includes information that requires semantic analysis.",
+            "type": "string",
+            "enum": [
+                "none",
+                "partial",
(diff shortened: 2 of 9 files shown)


Techatrix:
> I'm not sure how the config gen stuff works so any help on that would be appreciated.

You can use config_gen by only editing `src/config_gen/config.json` and then running `zig build gen` which should update the readme, schema.json and Config.zig.
You can also run `zig build gen -- --vscode-config-path /some/output/path.json` which will also generate the json configuration options that you can then copy into `package.json` in vscode-zig.

I've never added support for enums in config_gen so until I take care of that, it can't be used. I will do a follow up PR to fix that.

And also, is it intentional that the default value is different between the ZLS config and the vscode-zig config?

Vexu:
> And also, is it intentional that the default value is different between the ZLS config and the vscode-zig config?

Looks like I forgot to edit the default value for the zls config, it should probably be `"full"` as you suggested. The vscode extension already provides grammar based highlighting so for it the default should be partial.

Techatrix:
This patch with #1072 should update allow files accordingly after running `zig build gen`
```patch
diff --git a/src/config_gen/config.json b/src/config_gen/config.json
index 1ca86d9..3ac4b16 100644
--- a/src/config_gen/config.json
+++ b/src/config_gen/config.json
@@ -25,10 +25,15 @@
             "default": "true"
         },
         {
-            "name": "enable_semantic_tokens",
-            "description": "Enables semantic token support when the client also supports it",
-            "type": "bool",
-            "default": "true"
+            "name": "semantic_tokens",
+            "description": "Set level of semantic tokens. Partial only includes information that requires semantic analysis.",
+            "type": "enum",
+            "enum": [
+                "none",
+                "partial",
+                "full"
+            ],
+            "default": "full"
         },
         {
             "name": "enable_inlay_hints",
```

Techatrix (approved):
:+1: 
If you didn't know, [`tres`](https://github.com/ziglibs/tres) can be used to avoid having to do the `jsonStringify` manually.
```patch
diff --git a/src/main.zig b/src/main.zig
index 19cc527..b3e6251 100644
--- a/src/main.zig
+++ b/src/main.zig
@@ -3,6 +3,7 @@ const zig_builtin = @import("builtin");
 const build_options = @import("build_options");
 const tracy = @import("tracy.zig");
 const known_folders = @import("known-folders");
+const tres = @import("tres");
 const Config = @import("Config.zig");
 const configuration = @import("configuration.zig");
 const Server = @import("Server.zig");
@@ -129,7 +130,7 @@ fn updateConfig(
         var buffer = std.ArrayListUnmanaged(u8){};
         defer buffer.deinit(allocator);
 
-        try std.json.stringify(cfg, .{}, buffer.writer(allocator));
+        try tres.stringify(cfg, .{}, buffer.writer(allocator));
         const header = Header{ .content_length = buffer.items.len };
         try header.write(false, file.writer());
         try file.writeAll(buffer.items);
```
Feels weird to review your code, instead of the other ways round...

Vexu:
> If you didn't know, [`tres`](https://github.com/ziglibs/tres) can be used to avoid having to do the `jsonStringify` manually.

I did not know, it's a bit nicer so thanks for the suggestion.

> Feels weird to review your code, instead of the out ways round...

Nice to swap roles occasionally.

Vexu:
Does this need anything else?

PR #1075, opened 2023-03-17
Struct init fields completion.

nullptrdevs:
Up for logic review and a fuzz (suggestions still welcome).


leecannon:
This works well for the example case but there are some very common cases that it does not support:

- types returned by functions (any generic)
```zig
fn NonGenericGeneric() type {
    return struct {
        a: bool,
        b: bool,
    };
}

fn nonGenericGeneric() void {
    var r = NonGenericGeneric(){}; // no completion
}
```

- completion within the container itself
```zig
const MyStruct = struct {
    a: bool,
    b: bool,
    
    fn inside() void {
        var s = MyStruct{}; // no completion 
    }
};

fn outside() void {
    var s = MyStruct{}; // completion
}
```

- file level structs
```zig
// inside "MyStruct.zig"
a: bool,
b: bool,

// inside "main.zig"
const MyStruct = @import("MyStruct.zig");

fn topLevelStruct() void {
    var s = MyStruct{}; // no completion 
}
```

- aliases
```zig
const MyStruct = struct {
    a: bool,
    b: bool,
};

const Alias = MyStruct;

fn alias() void {
    var s = Alias{}; // no completion
}
```

leecannon:
I'm okay merging this as is, as it does increases ZLS capabilities.
I'm just making note of things that can still be improved in this area.

nullptrdevs:
Thank you for taking the time to review, I will look into all of these, but some initial thoughts:

Case 1: No idea where to start. Something something ComptimeInterpreter.
Case 2: The parser skips creating a node for MyStruct,
```zig
const MyStruct = struct {
    a: bool,
    b: bool,
    
    fn inside() void {
        var s = MyStruct{.}; // no completion 
    }
};

fn outside() void {
    var s = MyStruct{}; // completion
    MyStruct. // no completion here either
}
```
(Custom parser when?) Custom parse tokens looking for same .identifier, `MyStruct`, then parse .identifiers that might be fields(look for a .colon).
Case 3: This would be priority No.1.
Case 4: How common is this (worth the cpu cycles 😀 )?

Vaxeral:
There is a bug, struct member completions fail to appear if the struct is defined after it is used.

Screenshot from 2023-04-04 14-06-56

Screenshot from 2023-04-04 14-07-08

Vaxeral:
I am curious to know why this happens.

PR #1077, opened 2023-03-18
get build runner functional

leecannon:
This PR does the bare minimum to get the build runner working with zig master (also removed a few backwards compatibility checks while I was there).

To get around the pkg-config stuff being made private I just copied the implementation into the build runner.


leecannon:
I know this is not the long term solution as intergrating with zig as a server would be 100% better.

SuperAuguste (approved):
If it works it works 🤷‍♂️

Issue #1076, opened 2023-03-17
Explore compile server integration


SuperAuguste:
Semi-reviving #870.

Additionally this might allow us to do some:
* Build on save
* Hot reload on save (🙀)

Here's a nice resource: https://github.com/kubkon/zig-hcs-client

~~Please don't nerd snipe me @Techatrix :P~~

SuperAuguste:
Like a fine wine, I think we should let this compiler feature mature a bit as it's a tad too unstable for use in zls right now (and that's saying something because we're talking about zls here :P). I'll leave this here for the near to far future when it stabilizes a bit. :)

nullptrdevs:
https://github.com/ziglang/zig/compare/hcs-win-poc

techatrix:
We have Build on save in ZLS. So I guess this is fixed?

PR #1079, opened 2023-03-20
Fix remaining popScope issues

SuperAuguste:
You foolish bugs; you cannot escape me for I have run into you while messing around with #1076.

### DIFF
--- a/src/analysis.zig
+++ b/src/analysis.zig
@@ -2712,7 +2712,6 @@ fn makeScopeInternal(context: ScopeContext, node_idx: Ast.Node.Index) error{OutO
                     },
                     .other,
                 );
-                defer context.popScope();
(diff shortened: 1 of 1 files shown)


Techatrix (approved):
looks good. Do you have a code example that could be converted into a test?
I've also worked on document scope creation which should also fix this. I will PR that, It will probably undo your changes if accepted because I did a more aggressive rework.

SuperAuguste:
Thanks for the quick review! Sounds good to me; I'm down to merge this to quickly fix this until your PR lands if that works for you. No tests right now but I'll gladly PR them in the future so we can make sure we don't regress. :)

SuperAuguste:
Superseded by #1084.

PR #1083, opened 2023-03-20
update build.zig

Techatrix:
looks like the build script stopped running the test but instead only compiled them since zig/14647.


leecannon (approved):
Looks good.

Bit weird that running tests outputs what looks like an error but all tests pass (or skip).
```
$ zig build test
run test: error: const bar = foo;
const foo = 3;
0:12: use of undeclared identifier 'foo'
```

PR #1101, opened 2023-03-29
Make Analyser use its own arena

SuperAuguste:
Thanks for the advice @Techatrix! Let's see if this works...


SuperAuguste:
Huh wtf is going on with CI...??

Techatrix (approved):

PR #445, opened 2021-12-30
add a default option for `@import(builtin)`

nektro:
fixes #401

image

please sqush commit this

PR #1103, opened 2023-03-31
provide more semantic token information + refactor and fixes

### LINKED ISSUE #843: semantic_tokens: Normal comments don't get evaluated if they are alone

Techatrix:
these changes had a visible effect depending on the theme used:
- whether or not something is a namespace, struct, union, etc. is now set with a `TokenType` instead of `TokenModifiers`
- container fields that are named after a keyword were previously highlighted like an identifier

fixes #843

Here is how the above changed have affected highlighting.

Before |After
|:--------:|:--------:|
VS-Code "Default Dark+" | VS-Code "Default Dark+"
  |  
VS-Code "Monokai" | VS-Code "Monokai"
 | 
Sublime Text 4 "Default Dark" | Sublime Text 4 "Default Dark"
 |

### DIFF
--- a/src/Server.zig
+++ b/src/Server.zig
@@ -562,22 +562,8 @@ fn initializeHandler(server: *Server, request: types.InitializeParams) Error!typ
                     .full = .{ .bool = true },
                     .range = .{ .bool = true },
                     .legend = .{
-                        .tokenTypes = comptime block: {
-                            const tokTypeFields = std.meta.fields(semantic_tokens.TokenType);
-                            var names: [tokTypeFields.len][]const u8 = undefined;
-                            for (tokTypeFields, &names) |field, *name| {
-                                name.* = field.name;
-                            }
-                            break :block &names;
-                        },
-                        .tokenModifiers = comptime block: {
-                            const tokModFields = std.meta.fields(semantic_tokens.TokenModifiers);
-                            var names: [tokModFields.len][]const u8 = undefined;
-                            for (tokModFields, &names) |field, *name| {
-                                name.* = field.name;
-                            }
-                            break :block &names;
-                        },
+                        .tokenTypes = std.meta.fieldNames(semantic_tokens.TokenType),
+                        .tokenModifiers = std.meta.fieldNames(semantic_tokens.TokenModifiers),
                     },
                 },
             },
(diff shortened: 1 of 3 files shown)


leecannon:
Few things that don't seem quite right.

These four should all have the same color (also happened to notice that https://github.com/zigtools/zls/pull/445 doesn't work any more)
image

Its weird that both usages of `var_stdout` have the same color but the same is not true of `const_stdout`
image

Techatrix:
I've reverted the change about marking const declarations readonly and it now looks more consistent even though a bit less colorful.

leecannon (approved):

PR #1108, opened 2023-04-03
remove zls.json

Techatrix:
You may think the commit message is a bit misleading.
But that's exactly why I chose it...


Techatrix:
got 'em

PR #1114, opened 2023-04-04
disable assertions in smallestEnclosingSubrange

Techatrix:
This code produces ast nodes that have intersecting source locations because calling `firstToken` on the `bar` var decl node will yield the `""` token and not the `var` token. 
```zig
foo: ""
var bar = 0;
```

### DIFF
--- a/src/ast.zig
+++ b/src/ast.zig
@@ -1642,10 +1642,11 @@ pub fn smallestEnclosingSubrange(children: []const offsets.Loc, loc: offsets.Loc
         0 => return null,
         1 => return if (offsets.locInside(loc, children[0])) .{ .start = 0, .len = 1 } else null,
         else => {
-            for (children[0 .. children.len - 1], children[1..]) |previous_loc, current_loc| {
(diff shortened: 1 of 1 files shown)


SuperAuguste:
huh nice catch! what an odd parser behavior 😅

Techatrix:
its actually not that surprising once you look at how firstToken is implemented on var decls:
```zig
.global_var_decl,
.local_var_decl,
.simple_var_decl,
.aligned_var_decl,
=> {
    var i = main_tokens[n]; // mut token
    while (i > 0) {
        i -= 1;
        switch (token_tags[i]) {
            .keyword_extern,
            .keyword_export,
            .keyword_comptime,
            .keyword_pub,
            .keyword_threadlocal,
            .string_literal,
            => continue,

            else => return i + 1 - end_offset,
        }
    }
    return i - end_offset;
},
```
You may do something like `extern "a" var foo = 3;` which is why string literals are included.

leecannon (approved):

PR #1115, opened 2023-04-05
Avoid accessing inactive union fields in `completeDot`

nullptrdevs:
As reported on Discord by CodotakuYT:
```zig
fn populate(T: type, index: i32) T {
    _ = index;
    @typeInfo(T).struct.
}
```
```
thread 17384 panic: access of union field 'ast_node' while field 'param_payload' is active
C:\Users\taoua\zls\src\features\completions.zig:875:61: 0x7ff68070caac in completeDot (zls.exe.obj)
                const node_data = nodes_data[local_decl.decl.ast_node];
```


SuperAuguste (approved):
LGTM

PR #1116, opened 2023-04-05
Nested struct init fields completion

nullptrdevs:
Attempt at making nested struct init fields completion work

Needs testing.. more so that it doesn't break existing stuff.

Examples (ignore the obvious field types mismatch, didn't need to be correct for this autocomplete example):
nested_struct_init_example1
nested_struct_init_example2
nested_struct_init_example3

### DIFF
--- a/src/features/completions.zig
+++ b/src/features/completions.zig
@@ -821,7 +821,7 @@ fn completeDot(server: *Server, handle: *const DocumentStore.Handle, source_inde
         }
 
         // iterate until we find current token loc (should be a .period)
-        while (upper_index > 1) : (upper_index -= 1) {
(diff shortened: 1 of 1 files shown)


leecannon (approved):

leecannon:
Looks like github got confused with this one, I merged this PR in 7978485aaf5023dab76957741fa3d36da2da4acb but the page froze and after refresh the PR is still open and is complaining about conflicts before merging (and it looks like it is conflicting with itself).

This PR has definitely been merged into master (https://github.com/zigtools/zls/commit/735f884394ca23fbb529c9389dadfcd9e274e591) so i'm going to close this PR.

PR #1122, opened 2023-04-13
Use plain text if the client tells zls it's preferred

arp242:
LSP clients send a list of supported formats for definition and completion; the specification says "the order describes the preferred format of the client."[1]

My client sends:

	'hover': {'contentFormat': ['plaintext', 'markdown']},

So it should prefer plaintext, using markdown if that's not supported by the server. zls behaved slightly different: it would use Markdown if it appears at all in the list of supported formats.

This fixes it so that it will use plain text if that appears before 'markdown' in the list.

[1]: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_hover

### DIFF
--- a/src/Server.zig
+++ b/src/Server.zig
@@ -399,6 +399,9 @@ fn initializeHandler(server: *Server, request: types.InitializeParams) Error!typ
         if (textDocument.hover) |hover| {
             if (hover.contentFormat) |content_format| {
                 for (content_format) |format| {
+                    if (format == .plaintext) {
+                        break;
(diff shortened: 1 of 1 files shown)


arp242:
I'd add a test, but it doesn't look like there are any tests for this yet and this is the first Zig code I'm writing; I don't really know enough (yet) to write decent tests. I tested it manually by frobbing with the contentFormat in my LSP client.

leecannon (approved):

PR #1124, opened 2023-04-13
update to latest zig: add c_char type

kubkon:
I don't know whether the stage2 sources are updated manually or if there exists an automatic way of doing this, but here's a diff that updates to latest Zig master with breaking AST changes. Feel free to close this PR and apply yourself if it's not the correct way to do it.


nullptrdevs:
I don't know if this is correct either, but it works :smile: 

Anyhow -- it'd be wise to bump up min_zig to prevent confusion given how cryptic of an error
```
src/stage2/AstGen.zig:9069:13: error: std.zig.isPrimitive() is not aware of Zir instr 'c_char_type'
```
is.

SuperAuguste (approved):

SuperAuguste:
Thanks @kubkon!! :)

Issue #1147, opened 2023-04-20
Can't build with Zig master


yerlaser:
### Zig Version

0.11.0-dev.2680+a1aa55ebe

### Zig Language Server Version

825cbd2e074488259ca1123a1d9288b2bb244a62

### Steps to Reproduce

Run ```zig build -Doptimize=ReleaseSafe``` or ```zig build```

### Expected Behavior

ZLS should build.

### Actual Behavior

Error message: ```error: EndOfStream```

kassane:
Same error!!

leecannon:
Same here but works with zig 0.11.0-dev.2661+d2a799c65

benbot:
A rough `git bisect` tells me that the first commit with this issue is `a78950ce82b92d1e107fec5e6529066b5e8eb18d`

SuperAuguste:
@benbot this is not an issue with zls but rather with Zig.

leecannon:
The upstream issue (https://github.com/ziglang/zig/issues/15376) has been resolved.

I am not at a computer can someone see if this issue is resolved. @yerlaser @kassane 

_note: at time of writing the master zig tarball will not contain the fix_

leecannon:
Thanks @BratishkaErik

yerlaser:
> 

Yep. Should I build from sources instead?

PR #1148, opened 2023-04-22
Mem fixes

nullptrdevs:
* `gpa.deinit()` now returns an enum (Zig `0.11.0-dev.2730+`)
* avoid double free when `--config-path` is specified
* avoid leaking memory in `Server.create` if `configChanged` returns err

These fixes reveal a discrepancy — `main.getConfig` expects `--config-path` to be a path to file, a `zls.json`, while
`Server.configChanged` expects it to be a `/path/to/dir` (`builtin_creation_dir`).


leecannon (approved):

leecannon:
CI failure is due to https://github.com/zigtools/zls/issues/1147, not the changes in this PR.

PR #1152, opened 2023-04-24
Fix to make zls compile for 0.11.0-dev.2775+21aa55d34.

artix-boi:


leecannon:
CI failure is due to https://github.com/zigtools/zls/issues/1147 not the change in this PR.

PR #1157, opened 2023-04-27
diff: replace diffz with diff-zimilar

travisstaloch:


travisstaloch:
I'm not confident that this is complete. It seemed a little too easy.  For some reason, I thought the integration was much more involved.  

Also, I'm not sure how to benchmark whether its an improvement over diffz yet.  So far, I have only tried opening ziglang/zig/src/Sema.zig with a ReleaseFast zls build.  I tried adding one newline close to the end of the file and saving.  
With --log-enabled, i'm seeing around 65ms both before AND after my changes.  But I have no idea how to really measure this.  

Please let me know what I've missed and if I can somehow better benchmark the diffing time difference.

travisstaloch:
In case anyone is interested, I ported [diff-zimilar](https://github.com/travisstaloch/diff-zimilar/) because it seemed like a very memory-efficient diff match patch implementation.  I did it after helping Auguste with the diffz port and thinking the google diff match patch algo's memory usage might not be very efficient.  Unfortunately I haven't been able to measure this yet. Any help to that end would be appreciated.

travisstaloch:
discovered that diff-zimilar has some memory issues. will update here if i can improve them.

travisstaloch:
will maybe re-open if i can solve travisstaloch/diff-zimilar#1.

PR #1160, opened 2023-04-29
Updates zig-overlay referenced commit.

acristoffers:
Also updates default.nix to use a newer version of Zig.

Should we keep this file around (and updated) or remove it and keep only the flake? I've never seen anyone update it, whereas the flake gets fixed as soon as it breaks...


erikarvstedt:
Yes, `default.nix` should be removed. Non-flake users can always use [flake-compat](https://github.com/edolstra/flake-compat) instead.

leecannon (approved):

PR #1165, opened 2023-05-02
print zls version and path during startup

leecannon:
It will be much easier to debug issues if the zls version is output, this will be particularly handy when zls is managed by the extension.


leecannon:
@nullptrdevs good idea, done

SuperAuguste (approved):
this PR makes so much sense, not sure why we didn't implement this earlier 🤦‍♂️

Issue #1170, opened 2023-05-05
zls crashes when providing a zls.json


Ekopalypse:
### Zig Version

0.11.0-dev.2983+f05cd008d

### Zig Language Server Version

0.11.0-dev.408+28863f4

### Steps to Reproduce

Running zls --config-path C:\temp\zls.json

```
D:\ProgramData\Compiler\zig\zls\bin>zls --config-path C:\temp\zls.json
info : ( main ): Starting ZLS 0.11.0-dev.408+28863f4 @ 'zls'
info : (config): Using zig executable 'D:\ProgramData\Compiler\zig\zig.exe'
info : (config): Using zig lib path 'D:\ProgramData\Compiler\zig\lib'
info : (config): Using build runner global cache path 'C:\Users\eko\AppData\Local\zig'
info : (config): Detected runtime zig version: '0.11.0-dev.2983+f05cd008d'
error: (default): NotDir
???:?:?: 0x7ff72369512d in ??? (zls.exe)
???:?:?: 0x7ff723695135 in ??? (zls.exe)
???:?:?: 0x7ff72367cbbe in ??? (zls.exe)
???:?:?: 0x7ff72367d039 in ??? (zls.exe)
???:?:?: 0x7ff72368681a in ??? (zls.exe)
???:?:?: 0x7ff723684340 in ??? (zls.exe)
???:?:?: 0x7ff72368681a in ??? (zls.exe)
???:?:?: 0x7ff723684340 in ??? (zls.exe)
```

Procmon reports

```
10:29:48,6037407	zls.exe	9536	CreateFile	C:\temp\zls.json	SUCCESS	Desired Access: Generic Read, Disposition: Open, Options: Synchronous IO Non-Alert, Non-Directory File, Attributes: N, ShareMode: Read, Write, Delete, AllocationSize: n/a, OpenResult: Opened
10:29:48,6038658	zls.exe	9536	ReadFile	C:\temp\zls.json	SUCCESS	Offset: 0, Length: 282, Priority: Normal
10:29:48,6039368	zls.exe	9536	ReadFile	C:\temp\zls.json	END OF FILE	Offset: 282, Length: 4.967
10:29:48,6040461	zls.exe	9536	CloseFile	C:\temp\zls.json	SUCCESS	
[...]

PR #1177, opened 2023-05-08
 Fix panic caused by custom zls.json path

Hanaasagi:
Resolve https://github.com/zigtools/zls/issues/1170. 

If `builtin_path` is set to `null` in zls.json, then the same directory as zls.json will be used.

Support:

- relative path, e.g. `./zls.json` 
- absolute path, e.g. `/tmp/zls.json`

### DIFF
--- a/src/configuration.zig
+++ b/src/configuration.zig
@@ -130,7 +130,11 @@ pub fn configChanged(config: *Config, runtime_zig_version: *?ZigVersionWrapper, 
         defer allocator.free(result.stdout);
         defer allocator.free(result.stderr);
 
-        var d = try std.fs.cwd().openDir(builtin_creation_dir.?, .{});
+        var d = if (std.fs.path.isAbsolute(builtin_creation_dir.?))
+            try std.fs.openDirAbsolute(builtin_creation_dir.?, .{})
+         else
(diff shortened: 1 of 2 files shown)


src/configuration.zig
```diff
@@ -130,7 +130,11 @@ pub fn configChanged(config: *Config, runtime_zig_version: *?ZigVersionWrapper,
         defer allocator.free(result.stdout);
         defer allocator.free(result.stderr);
 
-        var d = try std.fs.cwd().openDir(builtin_creation_dir.?, .{});
+        var d = if (std.fs.path.isAbsolute(builtin_creation_dir.?))
```
SuperAuguste, line 133: I don't believe that this needs to be changed as openDir from cwd should handle absolute paths properly without this check :)

SuperAuguste (changes requested):

leecannon (approved):

leecannon (approved):

SuperAuguste (approved):
Sorry for the slow review, this looks great now, thanks! :))

SuperAuguste (approved):
Sorry for the slow review, this looks great now, thanks! :))

PR #1180, opened 2023-05-09
Added std.zig.number_literal.Error.invalid_hex_exponent rule

KoltPenny:
Added simple support for a new field in `std.zig.number_literal.Error` for zig version `0.11.0-dev`.

This missing field in the switch breaks compilation.


Techatrix:
this has already been fixed in #1178.

KoltPenny:
Thanks. It was such a quick fix that I forgot to check.

PR #1194, opened 2023-05-19
Update Nix Flakes file

sagehane:
Fixes the breakage for Nix flakes introduced by https://github.com/zigtools/zls/pull/1191.

PR #1195, opened 2023-05-20
chore: update flake.lock to fix nix-build

willruggiano:


willruggiano:
#1194

sagehane:
I've also been wondering if adding `result` to the `.gitignore` file is worth it.

PR #1198, opened 2023-05-24
Replace nonexistent `std.sort.sort` with `std.mem.sort`

Tyrubias:
The pull request ziglang/zig#15412 changes the sorting functions built into Zig and as a result `std.sort.sort` doesn't exist anymore. This pull request changes usages of `std.sort.sort` to `std.mem.sort` which is consistent with the changes made in the standard library to change `std.sort.sort` to `std.mem.sort`. Currently, `std.mem.sort` calls `std.sort.block`, which is an stable in-place sort with $`O(n)`$ best case, $`O(n\log(n))`$ worst case, and $`O(n\log(n))`$ average case time complexity. The block sort also has $`O(1)`$ memory usage.

### DIFF
--- a/src/Server.zig
+++ b/src/Server.zig
@@ -1142,7 +1142,7 @@ fn inlayHintHandler(server: *Server, request: types.InlayHintParams) Error!?[]ty
         }
     };
 
-    std.sort.sort(inlay_hints.InlayHint, hints, {}, helper.lessThan);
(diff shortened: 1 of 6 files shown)


leecannon:
Just waiting for the zig tarballs to update with this change.

leecannon:
CI is failing due to https://github.com/zigtools/zls/issues/1199
But local tests are passing.

leecannon (approved):

Issue #689, opened 2022-09-29
Code Actions & Autofix


Techatrix:
This issue contains a collection of possible code actions that can be integrated into zls.
Some of them may not be easily implementable without #551 or #552.
Feel free to make suggestions if you have some ideas.

### Unused Parameter
```zig
fn foo(parameter: u32) void {}
```
- [x] remove parameter
```zig
fn foo() void {}
```
- [x] discard parameter (part of autofix)
```zig
fn foo(parameter: u32) void {
    _ = parameter;
}
```

### Unused Variable/Constant
```zig
fn foo() void {
    var variable = 0;
    const constant = 0;
}
```

- [ ] remove variable/constant
```zig
fn foo() void {}
```

- [x] discard variable/constant (part of autofix)
```zig
fn foo() void {
    var variable = 0;
    _ = variable;
    const constant = 0;
    _ = constant;
}
```

### Capture
```zig
for("") |unused_capture, unused_index_capture| {}
switch(tagged_union) {
    .tag1 => |capture| expression,
    .tag2 => |capture| {},
}
```
- [x] remove capture
```zig
for ("") |_| {}
switch (tagged_union) {
    .tag1 => expression,
    .tag2 => {},
}
```
- [ ] discard capture (part of autofix)
```zig
for ("") |unused_capture, unused_index_capture| {
    _ = unused_capture;
    _ = unused_index_capture;
}
switch (tagged_union) {
    .tag1 => |capture| blk: {
        _ = capture;
        break :blk expression;
    },
    .tag2 => |capture| {
        _ = capture;
    },
}
```

### Unreachable code

```zig
fn foo() u32 {
    return 0;
    return 1;
}
```

- [ ] remove unreachable code

```zig
fn foo() u32 {
    return 0;
}
```

- [ ] comment out unreachable code
This can be part of autofix if this step is also reversible.
```zig
fn foo() u32 {
    return 0;
    // return 1;
}
```

### A ./ is not needed in imports
```zig
const file = @import("./file.zig");
```
- [ ] remove unnecessery ./
```zig
const file = @import("file.zig");
```

### Functions should be camelCase
```zig
fn foo_bar_baz() void {}
```
- [ ] make function name camelCase
```zig
fn fooBarBaz() void {}
```

### Type functions should be PascalCase
```zig
fn foo_bar_baz(comptime T: type) type { return T; }
```
- [ ] make function name PascalCase
```zig
fn FooBarBaz(comptime T: type) type { return T; }
```

[...]

PR #1202, opened 2023-05-27
autofix: discard captures + some multi for loop support

travisstaloch:
this patch makes autofix add discards for unused loop and switch case captures which have curlies.  it prevents adding duplicate 'remove capture' actions by checking previous action ranges.

it removes special casing of index captures now that multi for loops have arrived.

* make getCaptureLoc() return only single Loc
* remove CaptureLocs struct which is no longer used
* add DiagnosticKind.@"switch tag capture" to handle 'inline else => |x, tag|' discards
* add test "getCaptureLoc"

### DIFF
--- a/src/features/code_actions.zig
+++ b/src/features/code_actions.zig
@@ -478,13 +478,17 @@ fn getCaptureLoc(text: []const u8, loc: offsets.Loc) ?offsets.Loc {
     const start_pipe_position = blk: {
         var i = loc.start;
         while (true) : (i -= 1) {
-            if (text[i] == '|' or i == 0) break;
+            if (text[i] == '|') break;
+            if (i == 0) return null;
         }
         break :blk i;
     };
 
     const end_pipe_position = (std.mem.indexOfScalarPos(u8, text, start_pipe_position + 1, '|') orelse
         return null) + 1;
+
+    const trimmed = std.mem.trim(u8, text[start_pipe_position + 1 .. end_pipe_position - 1], &std.ascii.whitespace);
+    if (trimmed.len == 0) return null;
 
     return .{ .start = start_pipe_position, .end = end_pipe_position };
 }
@@ -504,10 +508,13 @@ test "getCaptureLoc" {
         const captext = text[caploc.start..caploc.end];
         try std.testing.expectEqualStrings(text, captext);
     }
-    {
-        const caploc = getCaptureLoc("||", .{ .start = 1, .end = 2 });
-        try std.testing.expect(caploc == null);
-    }
+
+    try std.testing.expect(getCaptureLoc("||", .{ .start = 1, .end = 2 }) == null);
+    try std.testing.expect(getCaptureLoc(" |", .{ .start = 1, .end = 2 }) == null);
+    try std.testing.expect(getCaptureLoc("| ", .{ .start = 1, .end = 2 }) == null);
+    try std.testing.expect(getCaptureLoc("||", .{ .start = 1, .end = 1 }) == null);
+    try std.testing.expect(getCaptureLoc("| |", .{ .start = 1, .end = 3 }) == null);
+    try std.testing.expect(getCaptureLoc("|    |", .{ .start = 1, .end = 6 }) == null);
 }
 
 fn isSymbolChar(char: u8) bool {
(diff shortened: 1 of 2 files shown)


src/features/code_actions.zig
```diff
@@ -166,64 +164,67 @@ fn handleUnusedVariableOrConstant(builder: *Builder, actions: *std.ArrayListUnma
...
+    // search for a matching action by comparing ranges.
+    const remove_cap_loc = builder.createTextEditLoc(capture_loc, "");
+    const found_remove_capture_action = outer: for (actions.items) |item| {
+        const edit = item.edit orelse continue;
+        const changes = edit.changes orelse continue;
+        var iter = changes.iterator();
+        while (iter.next()) |entry| {
+            const t_edits = entry.value_ptr.*;
+            for (t_edits) |t_edit| {
+                if (remove_cap_loc.range.start.line == t_edit.range.start.line and
```
SuperAuguste, line 209: At a glance this seems HashMapable, I like the idea you mentioned where we'd pass around the map with actions :)

src/features/code_actions.zig
```diff
@@ -166,64 +164,67 @@ fn handleUnusedVariableOrConstant(builder: *Builder, actions: *std.ArrayListUnma
...
-                "",
-            )}),
-        });
+    // look for next non-whitespace after last '|'. if its a '{' we can insert discards.
+    // this means bare loop/switch captures (w/out curlies) aren't supported.
+    var block_start = capture_loc.end + 1;
+    while (block_start < builder.handle.text.len and
+        std.ascii.isWhitespace(builder.handle.text[block_start])) : (block_start += 1)
+    {}
+    if (builder.handle.text[block_start] != '{') {
```
SuperAuguste, line 179: This breaks for a case like
```
    for (0..10, 0..10, 0..10) |i, j, k| i = 1;
```
but so does every other capture, definitely doesn't have to be fixed in this PR; perhaps we could open an issue for this :)

src/features/code_actions.zig
```diff
@@ -166,64 +164,67 @@ fn handleUnusedVariableOrConstant(builder: *Builder, actions: *std.ArrayListUnma
...
-            .edit = try builder.createWorkspaceEdit(&.{builder.createTextEditLoc(
-                .{ .start = capture_locs.value.end, .end = capture_locs.loc.end - 1 },
-                "",
-            )}),
-        });
+    // look for next non-whitespace after last '|'. if its a '{' we can insert discards.
+    // this means bare loop/switch captures (w/out curlies) aren't supported.
+    var block_start = capture_loc.end + 1;
+    while (block_start < builder.handle.text.len and
+        std.ascii.isWhitespace(builder.handle.text[block_start])) : (block_start += 1)
```
SuperAuguste, line 177: Another edge case that we can open an issue for:
```
if (1 == 1) |a|
    //a
    {}
```
won't work as the comment isn't whitespace - I've literally never seen zig code that does this in the wild but you never know - perhaps we could even tag this issue contributor friendly and bring in some new peeps :)

src/features/code_actions.zig
travisstaloch: created https://github.com/zigtools/zls/issues/1203

src/features/code_actions.zig
travisstaloch: Created https://github.com/zigtools/zls/issues/1204

travisstaloch:
addresses #689

tested with
```zig
test {
    for (0..10, 0..10, 0..10) |i, j, k| {}
    switch (union(enum) {}) {
        inline .a => |cap, tag| {},
    }
    if (null) |x| {}
    if (null) |v| {} else |e| {}
}
```
this becomes:
```zig
test {
    for (0..10, 0..10, 0..10) |i, j, k| {
        _ = k;
        _ = j;
        _ = i;
    }
    switch (union(enum) {}{}) {
        inline .a => |cap, tag| {
            _ = tag;
            _ = cap;
        },
    }
    if (null) |x| {
        _ = x;
    }
    if (null) |v| {
        _ = v;
    } else |e| {
        _ = e;
    }
}
```

leecannon:
This looks good :+1:

leecannon (approved):

travisstaloch:
interesting note: just tested a release-fast build and the discard order is reversed back to source order. 

```zig
test {
    for (0..10, 0..10, 0..10) |i, j, k| {
        _ = i;
        _ = j;
        _ = k;
    }
...
```

glad to see they're in source order. not sure why they were reversed in the debug build.

PR #1236, opened 2023-06-14
autofix now handles comments in the event of a newline w/ comments

jowi-dev:
This expands slightly on the implementation which currently checks to see if there was a non whitespace character by:
- checking to see if there is a comment
- if there is a comment, wait until there is a newline before caring about whether or not there is a non-whitespace character

This should cover the basis of needing to handle comments, and if there are comments from line to line that should also be fine

### DIFF
--- a/src/features/code_actions.zig
+++ b/src/features/code_actions.zig
@@ -183,9 +183,16 @@ fn handleUnusedCapture(
     // look for next non-whitespace after last '|'. if its a '{' we can insert discards.
     // this means bare loop/switch captures (w/out curlies) aren't supported.
     var block_start = capture_loc.end + 1;
-    while (block_start < builder.handle.text.len and
-        std.ascii.isWhitespace(builder.handle.text[block_start])) : (block_start += 1)
-    {}
+    var is_comment = false;
+    while (block_start < builder.handle.text.len) : (block_start += 1)
+    {
+        //If the next two characters are // then its a comment
+        if(builder.handle.text[block_start] == '/' and builder.handle.text[block_start + 1] == '/') is_comment = true;
+        // if we go to a new line, drop the is_comment boolean
+        if(builder.handle.text[block_start] == '\n' and is_comment) is_comment = false;
+        //If the character is not a whitespace, and we're not in a comment then break out of the loop
+        if(!std.ascii.isWhitespace(builder.handle.text[block_start]) and !is_comment) break;
+    }
     if (builder.handle.text[block_start] != '{') {
         return;
     }
--- a/tests/lsp_features/code_actions.zig
+++ b/tests/lsp_features/code_actions.zig
@@ -81,6 +81,25 @@ test "code actions - discard captures" {
     );
 }
 
+test "code actions - discard capture with comment" {
+    try testAutofix(
+        \\test {
+        \\  if (1 == 1) |a|
+        \\      //a
+        \\      {}
+        \\}
+    ,
+        \\test {
+        \\    if (1 == 1) |a|
+        \\    //a
+        \\    {
+        \\        _ = a;
+        \\    }
+        \\}
+        \\
+    );
+}
+
 test "code actions - remove pointless discard" {
     try testAutofix(
         \\fn foo(a: u32) u32 {


src/features/code_actions.zig
```diff
@@ -183,9 +183,16 @@ fn handleUnusedCapture(
...
     // this means bare loop/switch captures (w/out curlies) aren't supported.
     var block_start = capture_loc.end + 1;
-    while (block_start < builder.handle.text.len and
-        std.ascii.isWhitespace(builder.handle.text[block_start])) : (block_start += 1)
-    {}
+    var is_comment = false;
+    while (block_start < builder.handle.text.len) : (block_start += 1)
+    {
+        //If the next two characters are // then its a comment
+        if(builder.handle.text[block_start] == '/' and builder.handle.text[block_start + 1] == '/') is_comment = true;
```
travisstaloch, line 190: you'll need to prevent reading oob here.

travisstaloch (changes requested):

src/features/code_actions.zig
nullptrdevs: A. Travis has a point

B. Somehow, in my mind, this implied that the check won't be done every iteration;
> - if there is a comment, wait until there is a newline before caring about whether or not there is a non-whitespace character

to illustrate:
```zig
    var in_a_comment = false;
    while (block_start < builder.handle.text.len) : (block_start += 1) {
        // if in_a_comment, only care if builder.handle.text[block_start] is == '\n'
        if (in_a_comment) {
            if (builder.handle.text[block_start] != '\n') continue; // loop until then
            in_a_comment = false;
            continue; // we already know current char is '\n'
        }
        // Fun fact: code_action are only provided if handle.tree.errors.len == 0
        // and a '/' after a capture is an error, ergo we're looking at a comment
        if (builder.handle.text[block_start] == '/') {
            in_a_comment = true;
            continue; // we know what current char is, go next
        }
        if (!std.ascii.isWhitespace(builder.handle.text[block_start])) break;
    }
```

jowi-dev: @nullptrdevs - unless I'm misunderstanding - does the check need to be done every iteration? Since there are no multiline comments in zig, my understanding of the problem here was that if there was a newline, we could reset the comment flag regardless of if we were already in one or not.

To your fun fact block - that is a clever trick! Would that work in the case of division? e.g.
```
if ( 1  == 1) |a|
//a
{
var b = 1/1;
}
```

it seems like technically we're not going to run into issues with the above, but searching on a single `/` seems riskier than just checking OOB + the next character. What do you think?

jowi-dev: OOB check added

jowi-dev: Added an additional iteration inside of the `is_comment = true` block so that we don't waste a loop on a character we've already checked

travisstaloch (approved):

src/features/code_actions.zig
```diff
@@ -183,9 +183,24 @@ fn handleUnusedCapture(
...
+            block_start += 1;
+            continue;
+        }
+        // if we go to a new line, drop the is_comment boolean
+        if(builder.handle.text[block_start] == '\n' and is_comment) {
+            is_comment = false;
+            continue;
+        }
+        //If the character is not a whitespace, and we're not in a comment then break out of the loop
+        if(!std.ascii.isWhitespace(builder.handle.text[block_start]) and !is_comment) break;
```
travisstaloch, line 202: this looks good. minor cleanup: seems like it would be easier to read if it was a `switch`. maybe something like:
```zig
        switch (builder.handle.text[block_start]) {
         '/' => if (block_start + 1 < builder.handle.text.len and builder.handle.text[block_start + 1] == '/') {
            is_comment = true;
            // we already know the next character is a `/` so lets skip that iteration
            block_start += 1;
        },
        // if we go to a new line, drop the is_comment boolean
        '\n' => if (is_comment) {
            is_comment = false;
        },
        //If the character is not a whitespace, and we're not in a comment then break out of the loop
        else => |c| if (!std.ascii.isWhitespace(c) and !is_comment) break,
        }
```

jowi-dev: Love this suggestion. updated

travisstaloch:
looks good. :+1:  thanks!

PR #1237, opened 2023-06-15
ci: add action to update the flake lock

leecannon:
I seriously don't understand nix or flakes so I don't know if this is a good idea or not :)


leecannon:
Someone who knows more about nix than "its some package manager thing" has said this is a bad idea or at least not a good one.

PR #1239, opened 2023-06-17
Fixed compatibility with latest builds of Zig

FalsePattern:
# Problem
`std.math.min`,
`std.math.max`,
`std.math.inf_f32`
Have been removed from the zig standard library and replaced with `@compileError`

# Solution
Replaced them with `@min`, `@max`, and `std.math.inf(f32)` respectively.

# Testing
This fix compiles and all tests with `zig build test` run properly under both `0.11.0-dev.3658+5d9e8f27d` and `0.11.0-dev.3678+f043071cd` when both this fix and the fix in diffz are applied. (latest versions on the website and the zig git repo as of writing this PR)

## Notes
Related to: https://github.com/ziglibs/diffz/pull/13

### DIFF
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -12,8 +12,8 @@
             .hash = "12204d32c44b494c0cbe553811dedfb7a73da37b13b492872dd4edc0340983072697",
         },
         .diffz = .{
-            .url = "https://github.com/ziglibs/diffz/archive/b966296b4489eb082b0831ec9a37d6f5e1906040.tar.gz",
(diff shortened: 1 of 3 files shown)


FalsePattern:
The aforementioned blocking diffz PR has been merged, `zig build test` now runs correctly on both compiler versions.

Paging @acristoffers, the nix flake will likely need to be updated because of the diffz dependency update.

leecannon (approved):

acristoffers:
The change in the Nix files will follow soon.

PR #1262, opened 2023-06-24
Resolve type of simple labeled block

FnControlOption:
For example, the following snippet is found in [`std.testing`](https://github.com/ziglang/zig/blob/a5e15ec/lib/std/testing.zig):

```zig
pub var allocator_instance = b: {
    if (!builtin.is_test)
        @compileError("Cannot use testing allocator outside of test block");
    break :b std.heap.GeneralPurposeAllocator(.{}){};
};
```

### DIFF
--- a/src/analysis.zig
+++ b/src/analysis.zig
@@ -1140,6 +1140,33 @@ fn resolveTypeOfNodeUncached(analyser: *Analyser, node_handle: NodeWithHandle) e
                 .handle = handle,
             };
         },
+        .block,
+        .block_semicolon,
+        .block_two,
+        .block_two_semicolon,
+        => {
+            const first_token = tree.firstToken(node);
+            if (token_tags[first_token] != .identifier) return null;
+
+            const block_label = tree.tokenSlice(first_token);
+
+            var buffer: [2]Ast.Node.Index = undefined;
+            const statements = ast.blockStatements(tree, node, &buffer).?;
+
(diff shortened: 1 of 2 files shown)


src/analysis.zig
```diff
@@ -1140,6 +1140,33 @@ fn resolveTypeOfNodeUncached(analyser: *Analyser, node_handle: NodeWithHandle) e
...
+            const first_token = tree.firstToken(node);
+            if (token_tags[first_token] != .identifier) return null;
+
+            const block_label = tree.tokenSlice(first_token);
+
+            var buffer: [2]Ast.Node.Index = undefined;
+            const statements = ast.blockStatements(tree, node, &buffer).?;
+
+            for (statements) |child_idx| {
+                // TODO: Recursively find matching `break :label` (e.g. inside `if`)
```
Techatrix, line 1157: If you wish to follow up and resolve this TODO, helper functions like [`iterateChildrenRecursive`](https://github.com/zigtools/zls/blob/a16fb19797d627a6cd500065aacee098c624a66c/src/ast.zig#L1545) may help you traverse the ast while searching for break statements.

Techatrix (changes requested):
Can you add a test case in [completion.zig](https://github.com/zigtools/zls/blob/master/tests/lsp_features/completion.zig).

Techatrix (approved):

PR #1269, opened 2023-06-25
Hover: Show name of struct type defined at file scope

FnControlOption:
For example, [`std.mem.Allocator`](https://github.com/ziglang/zig/blob/master/lib/std/mem/Allocator.zig) and [`std.http.Client`](https://github.com/ziglang/zig/blob/master/lib/std/http/Client.zig)


nektro:
screenshot?

nullptrdevs:
It seems to depend on the `pub const Name =` and ` @import("Name.zig");` matching, but that's probably true in 99% of the cases. :+1: :grinning:

leecannon (approved):

PR #1279, opened 2023-06-28
enable inlay hints for single-argument builtins

### LINKED ISSUE #1276: No inlays for single argument builtins, even if they're configured to appear
### Zig Version

0.11.0-dev.3879+5804f3f75

0.11.0-dev.515+e7af795

### Steps to Reproduce

Have `enable_inlay_hints` and `inlay_hints_show_builtin` as `true`, their default value.

Set `inlay_hints_exclude_single_argument` to `false`.

Write code that uses single argument builtins, e.g.

`@intFromFloat(value)`

### Expected Behavior

An inlay hint appears, e.g.

`@intFromFloat(`float: `value)`

### Actual Behavior

No inlay hint appears.

bertie-wheen:
This should resolve #1276, and as a side-effect might make `inlay_hints_exclude_builtins` less rare (in particular, people might like to exclude `@import`), and perhaps motivate increasing its configurability.

### DIFF
--- a/src/features/inlay_hints.zig
+++ b/src/features/inlay_hints.zig
@@ -14,7 +14,6 @@ const Config = @import("../Config.zig");
 const data = @import("../data/data.zig");
 
 /// don't show inlay hints for the given builtin functions
-/// builtins with one parameter are skipped automatically
 /// this option is rare and is therefore build-only and
 /// non-configurable at runtime
 pub const inlay_hints_exclude_builtins: []const u8 = &.{};
@@ -275,7 +274,7 @@ fn writeNodeInlayHint(
             var buffer: [2]Ast.Node.Index = undefined;
             const params = ast.builtinCallParams(tree, node, &buffer).?;
 
-            if (!builder.config.inlay_hints_show_builtin or params.len <= 1) break :blk;
+            if (!builder.config.inlay_hints_show_builtin or params.len == 0) break :blk;
 
             const name = tree.tokenSlice(main_tokens[node]);
 


SuperAuguste:
I disagree with this change; in my mind single-argument builtins are clear enough that inlay hints are not necessary.

bertie-wheen:
That's 100% a valid opinion, but surely it should be configurable? Perhaps the `inlay_hints_exclude_single_argument` should be split into two, e.g. `inlay_hints_exclude_single_argument` and `inlay_hints_exclude_single_argument_builtins` - some people might want single-argument inlays only for non-builtins, and other people might want inlays for everything.

bertie-wheen:
I can imagine that yours might be the majority view, and therefore the defaults should reflect it, but there's no harm in letting people choose differently, right?

bertie-wheen:
Btw, a counter-example to there being no benefit due to obviousness might be `@intFromEnum(enum_or_tagged_union: anytype) anytype`, as the parameter name gives information not obvious from the builtin name alone (i.e. that you can pass a tagged union instead of an enum).

leecannon:
I personally agree with @SuperAuguste's opinion on the usefulness of inlay hints for single argument builtins.

However it is weird that `inlay_hints_exclude_single_argument` applies to non-builtins but not builtins even when `inlay_hints_show_builtin` is true.

Personally I am in favour of merging this to at least provide the configuration option to users.

leecannon (approved):

PR #1280, opened 2023-06-28
Analyser: Set scope start at payload token if present

FnControlOption:
This enables hover for payload declaration:

payload

<sub><sub>(Yes, I changed to dark theme just for this screenshot)</sub></sub>


FnControlOption:
This should use `ast.lastToken` instead of `Ast.lastToken`

leecannon (approved):

PR #1281, opened 2023-06-28
Hover: Add links to referenced types when possible

FnControlOption:
Inspired by rust-analyzer, this PR adds "Go to &lt;type&gt;" in hover pop-ups:

pointer

function

chain

### DIFF
--- a/src/analysis.zig
+++ b/src/analysis.zig
@@ -3107,3 +3107,159 @@ fn makeScopeInternal(context: ScopeContext, tree: Ast, node_idx: Ast.Node.Index)
         },
     }
 }
+
+pub const ReferencedType = struct {
+    str: []const u8,
+    uri: []const u8,
+    loc: types.Position,
+
+    fn init(str: []const u8, handle: *const DocumentStore.Handle, token: Ast.TokenIndex) ReferencedType {
+        const loc = offsets.tokenToPosition(handle.tree, token, .@"utf-8");
+        return .{
+            .str = str,
+            .uri = handle.uri,
+            .loc = loc,
+        };
+    }
+};
+
+pub fn referencedTypes(
+    analyser: *Analyser,
+    allocator: std.mem.Allocator,
+    resolved_type: TypeWithHandle,
+    resolved_type_str: *[]const u8,
+    referenced_types: *[]const ReferencedType,
+) error{OutOfMemory}!void {
+    var list = std.ArrayList(ReferencedType).init(allocator);
+    if (resolved_type.type.is_type_val) {
+        _ = try analyser.addReferencedTypes(resolved_type, &list);
+        resolved_type_str.* = switch (resolved_type.type.data) {
+            .@"comptime" => |co| try std.fmt.allocPrint(allocator, "{}", .{co.value.index.fmt(co.interpreter.ip.*)}),
+            else => "type",
+        };
+    } else if (try analyser.addReferencedTypes(resolved_type, &list)) |str| {
+        resolved_type_str.* = str;
+    }
+    referenced_types.* = list.items;
+}
+
+fn addReferencedTypesFromNode(
+    analyser: *Analyser,
+    node_handle: NodeWithHandle,
+    referenced_types: *std.ArrayList(ReferencedType),
+) error{OutOfMemory}!void {
+    const type_handle = try analyser.resolveTypeOfNode(node_handle) orelse return;
+    _ = try analyser.addReferencedTypes(type_handle, referenced_types);
+}
+
+fn addReferencedTypes(
+    analyser: *Analyser,
+    type_handle: TypeWithHandle,
+    referenced_types: *std.ArrayList(ReferencedType),
+) error{OutOfMemory}!?[]const u8 {
+    const allocator = referenced_types.allocator;
+
+    const handle = type_handle.handle;
+    const tree = handle.tree;
+
+    const node_tags = tree.nodes.items(.tag);
+    const token_tags = tree.tokens.items(.tag);
+    const main_tokens = tree.nodes.items(.main_token);
+
+    switch (type_handle.type.data) {
+        .pointer,
+        .slice,
+        .error_union,
+        .primitive,
+        => |p| return offsets.nodeToSlice(tree, p),
+
+        .other => |p| switch (node_tags[p]) {
+            .root => {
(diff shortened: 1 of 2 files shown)


src/analysis.zig
```diff
@@ -3107,3 +3107,159 @@ fn makeScopeInternal(context: ScopeContext, tree: Ast, node_idx: Ast.Node.Index)
...
     }
 }
+
+pub const ReferencedType = struct {
+    str: []const u8,
+    uri: []const u8,
+    loc: types.Position,
+
+    fn init(str: []const u8, handle: *const DocumentStore.Handle, token: Ast.TokenIndex) ReferencedType {
+        const loc = offsets.tokenToPosition(handle.tree, token, .@"utf-8");
```
Techatrix, line 3117: use `server.offset_encoding` instead of hard coding utf-8

src/features/hover.zig
```diff
@@ -18,6 +18,9 @@ pub fn hoverSymbol(server: *Server, decl_handle: Analyser.DeclWithHandle, markup
     const tracy_zone = tracy.trace(@src());
     defer tracy_zone.end();
 
+    var temp_arena = std.heap.ArenaAllocator.init(server.allocator);
```
Techatrix, line 21: why create a new arena allocator here instead of using `server.arena`?

src/analysis.zig
```diff
@@ -3107,3 +3107,159 @@ fn makeScopeInternal(context: ScopeContext, tree: Ast, node_idx: Ast.Node.Index)
...
+        };
+    }
+};
+
+pub fn referencedTypes(
+    analyser: *Analyser,
+    allocator: std.mem.Allocator,
+    resolved_type: TypeWithHandle,
+    resolved_type_str: *[]const u8,
+    referenced_types: *[]const ReferencedType,
```
Techatrix, line 3131: nitpick: you could also return a struct instead of doing out parameters.

Techatrix (changes requested):
Great work, but I found some weird behavior when hovering over type declarations:
```zig
// example_file.zig
const SomeType = struct {};
const Self = @This();
```
If you hover on `SomeType`, you will get `Go to *SomeType*`
If you hover on `Self`, you will get `Go to *example_file*`

src/features/hover.zig
Techatrix: memory allocated with `server.arena` is only guaranteed to live until you finished handling a request. If you need something that lives until the server is stopped, you would use `server.allocator` and manage its lifetime. `server.arena` is exactly what you need.

Techatrix:
If fixing the `Self` example is too complex, focusing just on `SomeType` should be fine, especially because that example is far more common to encounter.

> Currently, type aliases get completely resolved to whatever's the underlying type. (In the Zig stdlib, this isn't a problem since AFAIK file-level structs are named the same as normal structs, e.g. `Allocator.zig`)

Maybe [resolveVarDeclAlias](https://github.com/zigtools/zls/blob/5c7bf22b5f630f5791d83c48183162808713ff74/src/analysis.zig#L369C14-L369C14) is what you need. It will resolve aliases but not completely resolve its type. 

> I have some local changes that adds `.type_function` variant to `Type.data`.

functions in `Type` can already be represented with the `Type.data.other` but a more explicit case may be helpful.

Techatrix (changes requested):
Everything is in order except that you should use `offsets.nodeToSlice` instead of `std.zig.Ast.getNodeSource` because ZLS uses a custom [lastToken](https://github.com/zigtools/zls/blob/ad6a02741045cecdd38b447c192adaa0a4eab750/src/ast.zig#L320) function that is safe to use on ast with syntax errors.

src/features/hover.zig
```diff
@@ -33,17 +34,27 @@ pub fn hoverSymbol(server: *Server, decl_handle: Analyser.DeclWithHandle, markup
...
+
                 break :def Analyser.getContainerFieldSignature(tree, field);
             } else {
                 break :def Analyser.nodeToString(tree, node) orelse return null;
             }
         },
         .param_payload => |pay| def: {
             const param = pay.param;
+
+            std.debug.assert(param.type_expr != 0);
```
Techatrix, line 55: Unfortunately I don't have a reproducer yet, but the following assertion failed while I was using your PR
```
thread 323332 panic: reached unreachable code
/opt/zig/0.11.0-dev.3886+0c1bfe271/files/lib/std/debug.zig:258:14: 0x32fb1c in assert (zls)
    if (!ok) unreachable; // assertion failure
             ^
/home/techatrix/zls/src/features/hover.zig:55:29: 0x48b547 in hoverSymbol (zls)
            std.debug.assert(param.type_expr != 0);
                            ^
/home/techatrix/zls/src/features/hover.zig:179:42: 0x3f29bd in hoverDefinitionGlobal (zls)
                .value = (try hoverSymbol(server, decl, markup_kind)) orelse return null,
                                         ^
/home/techatrix/zls/src/features/hover.zig:219:49: 0x3f1f40 in hover (zls)
        .var_access => try hoverDefinitionGlobal(server, source_index, handle),
                                                ^
/home/techatrix/zls/src/Server.zig:919:41: 0x39c3ea in hoverHandler (zls)
    const response = hover_handler.hover(server, source_index, handle);
                                        ^
/home/techatrix/zls/src/Server.zig:1506:35: 0x3600eb in processMessage (zls)
                break :blk handler(server, params) catch |err| {
                                  ^
/home/techatrix/zls/src/Server.zig:1362:26: 0x340d73 in processJsonRpc (zls)
    server.processMessage(message) catch |err| switch (message) {
                         ^
/home/techatrix/zls/src/main.zig:86:30: 0x34054a in loop (zls)
        server.processJsonRpc(json_message);
                             ^
/home/techatrix/zls/src/main.zig:428:13: 0x341828 in main (zls)
    try loop(server, record_file, replay_file);
            ^
/opt/zig/0.11.0-dev.3886+0c1bfe271/files/lib/std/start.zig:608:37: 0x32b0c4 in posixCallMainAndExit (zls)
            const result = root.main() catch |err| {
                                    ^
/opt/zig/0.11.0-dev.3886+0c1bfe271/files/lib/std/start.zig:367:5: 0x32abb1 in _start (zls)
    @call(.never_inline, posixCallMainAndExit, .{});
```

Techatrix (approved):

PR #1282, opened 2023-06-30
Hover improvements

FnControlOption:
Follow up to #1281

## Resolve underlying type recursively

new recursive underlying

<details><summary>Before</summary>

old recursive underlying

</details>

## Avoid duplicate "Go to &lt;type&gt;"

new duplicates

<details><summary>Before</summary>

old duplicates

</details>

## Add hover info for optional type and error union

optional type

error union

## Simplify resolved `if`/`switch` type when branches return same type

page allocator

## Resolve type to alias for "Go to &lt;type&gt;" when possible

new function

new variable

<details><summary>Before</summary>

old function

old variable

</details>

For reference, `Ast.Node.Index` is defined as `const Index = u32;`

This changes the behavior of "Go to std.mem.Allocator". Instead of opening `std/mem/Allocator.zig`, now it will open `std/mem.zig` and go to the line that has `const Allocator = @import("mem/Allocator.zig");`

allocator parameter

allocator alias

### DIFF
--- a/src/analysis.zig
+++ b/src/analysis.zig
@@ -3164,6 +3164,7 @@ fn addReferencedTypes(
     const node_tags = tree.nodes.items(.tag);
     const token_tags = tree.tokens.items(.tag);
     const main_tokens = tree.nodes.items(.main_token);
+    const datas = tree.nodes.items(.data);
 
     switch (type_handle.type.data) {
         .pointer,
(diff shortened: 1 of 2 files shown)


nullptrdevs:
Maybe it's too early in the morning here and I'm not getting the full picture ..but I have reservations
about the "Resolve each type as an alias when possible" change as showcased in the image on the right:

1 Knowing the underlying type at a glance is valuable,
2 It presents no information that couldn't be gleaned by reading the line above it.

InKryption:
Is there some way in which both the alias and the actual type could be provided?

nullptrdevs:
This looks excellent!


Regarding

> This changes the behavior of "Go to std.mem.Allocator". Instead of opening std/mem/Allocator.zig, now it will open std/mem.zig and go to the line that has `const Allocator = @import("mem/Allocator.zig");`

I have no strong opinion on the matter, this is what Go To Def does as it is, and although it might be convenient to provide the choice it might get cramped.

Technically Go To Def should pop open `std/mem/Allocator.zig`, and Go To Decl land you at `const Allocator = @import("mem/Allocator.zig");`, but I believe Auguste feels strongly about this being the way it is.

leecannon (approved):

PR #1309, opened 2023-07-10
transition from tres to std.json

Techatrix:
[tres](https://github.com/ziglibs/tres) provides json parsing and stringify on `std.json.Value` but now that the new `std.json` provides these capabilities there is no reason to keep an unnecessary dependency in zls.

[companion PR](https://github.com/zigtools/zig-lsp-codegen/pull/9) to zig-lsp-codegen

please update faster [ziglang.org](https://youtu.be/dQw4w9WgXcQ)


SuperAuguste:
Huge move, congrats!

PR #1311, opened 2023-07-11
multithreaded job processing

Techatrix:
Independent requests, diagnostics and build file loading will now be processed in parallel. This means faster message handling as well as near instant startup time ~~and more race conditions~~.

I would appreciate if anyone who reads this would try out this PR to see if there are any issues that need to be addressed. 
Also, should multithreading be enabled by default? I already made sure that ZLS can still be build as a single threaded executable.

I've also redesigned the public API of Server that plays nice with [zls-as-lib-demo](https://github.com/zigtools/zls-as-lib-demo).

Future Improvements:
- build on save using the Zig Compile Server #1076 (I am really excited for this one)
- request cancelling, partial results & progress report

Tracy is starting to look interesting
Screenshot from 2023-07-11 02-59-06

### DIFF
--- a/src/features/hover.zig
+++ b/src/features/hover.zig
@@ -16,6 +16,7 @@ const data = @import("../data/data.zig");
 
 pub fn hoverSymbol(
     server: *Server,
+    arena: std.mem.Allocator,
     decl_handle: Analyser.DeclWithHandle,
     markup_kind: types.MarkupKind,
     original_doc_str: ?[]const u8,
@@ -26,14 +27,14 @@ pub fn hoverSymbol(
     const handle = decl_handle.handle;
     const tree = handle.tree;
 
-    var type_references = Analyser.ReferencedType.Set.init(server.arena.allocator());
+    var type_references = Analyser.ReferencedType.Set.init(arena);
     var reference_collector = Analyser.ReferencedType.Collector.init(&type_references);
-    var doc_str = original_doc_str orelse try decl_handle.docComments(server.arena.allocator());
+    var doc_str = original_doc_str orelse try decl_handle.docComments(arena);
 
     const def_str = switch (decl_handle.decl.*) {
         .ast_node => |node| def: {
             if (try server.analyser.resolveVarDeclAlias(.{ .node = node, .handle = handle })) |result| {
-                return try hoverSymbol(server, result, markup_kind, doc_str);
+                return try hoverSymbol(server, arena, result, markup_kind, doc_str);
             }
 
             var buf: [1]Ast.Node.Index = undefined;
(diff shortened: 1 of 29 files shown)


FnControlOption:
Currently giving this a try while starting work on peer type resolution in zls; issues I'm having so far:

- Does not add `_ = foo` for unused `foo`
- VS Code is printing "[Error] Stopping server timed out"
- No error with `-Dsingle-threaded=true`

leecannon:
Ditto to VS Code printing "[Error] Stopping server timed out" even though it seems to just keep going and works fine.

Seems to be some weirdness with go to in hover, hovering over `errorable` in the below:
```zig
const E = error{ X, Y };
const S = struct { alpha: u32 };

fn errorable() E!S {
    return undefined;
}
```
On master:
image
This PR:
image

src/features/hover.zig
```diff
@@ -115,8 +119,9 @@ pub fn hoverSymbol(server: *Server, decl_handle: Analyser.DeclWithHandle, markup
         for (referenced_types, 0..) |ref, index| {
             if (index > 0)
                 try writer.print(" | ", .{});
-            const loc = offsets.tokenToPosition(ref.handle.tree, ref.token, server.offset_encoding);
-            try writer.print("[{s}]({s}#L{d})", .{ ref.str, ref.handle.uri, loc.line + 1 });
+            const source_index = offsets.tokenToIndex(ref.handle.tree, ref.token);
+            const line = 1 + std.mem.count(u8, ref.handle.tree.source[0..source_index], "\n");
+            try writer.print("Go to [{s}]({s}#L{d})", .{ ref.str, ref.handle.uri, line });
```
FnControlOption, line 124:
```suggestion
            try writer.print("[{s}]({s}#L{d})", .{ ref.str, ref.handle.uri, line });
```

Techatrix:
> Seems to be some weirdness with go to in hover, hovering over `errorable` in the below:

I overlooked a change while resolving merge conflicts. This should be resolved now

Techatrix:
> * Does not add `_ = foo` for unused `foo`

I've also encountered this issue and though this was a configuration issue. Is is it possible that you are using VSCode and have set `zig.astCheckProvider` to `extension` because that causes vscode-zig to set `enable_ast_check_diagnostics` to `false` that stops ZLS from generating ast-check diagnostics and therefore won't provide any code actions including autofix.

> [Error] Stopping server timed out

I suspect that I am not responding to some request. I will need to do some more investigating.

FnControlOption:
> is it possible that you are using VSCode and have set `zig.astCheckProvider` to `extension`

Nope, it's set to `zls`. It might be that the server simply crashed before it could add `_ = foo` but I'm not certain 🤔

leecannon:
Sometimes zls stops responding and the server has to be restarted.

Attaching a debugger shows multiple threads sitting here: https://github.com/zigtools/zls/blob/1c749e4d0a7500963354d8c321ca1115bf94841e/src/features/diagnostics.zig#L232

leecannon:
The issue occurs when multiple zig child processes are spawned to do ast check at the same time.

Adding a mutex to `getDiagnosticsFromAstCheck` stops the lockup but is there a better way?

leecannon:
Can't recreate the lockup any more :+1:.

I will keep trying to break it but so far this is looking good, also I will hold off on merging PRs for a while to prevent so many conflicts building up here.

FnControlOption:
Rebased this branch on commit 952d51b in `master` and currently doing `git merge master`. Will update with merge commit when finished

leecannon (approved):

PR #1312, opened 2023-07-11
Revert #1284

leecannon:
As zig comments are not markdown it turns out #1284 is not a good idea as it means that comments that look fine in source code just run together when zls renders them, example `std.HashMapUnmanaged`:

Comment
```
/// A HashMap based on open addressing and linear probing.
/// A lookup or modification typically occurs only 2 cache misses.
/// No order is guaranteed and any modification invalidates live iterators.
/// It achieves good performance with quite high load factors (by default,
/// grow is triggered at 80% full) and only one byte of overhead per element.
/// The struct itself is only 16 bytes for a small footprint. This comes at
/// the price of handling size with u32, which should be reasonable enough
/// for almost all uses.
/// Deletions are achieved with tombstones.
```

On master
image

With this revert applied
image

### DIFF
--- a/src/analysis.zig
+++ b/src/analysis.zig
@@ -45,21 +45,21 @@ pub fn invalidate(self: *Analyser) void {
     _ = self.arena.reset(.free_all);
 }
 
-pub fn getDocCommentsBeforeToken(allocator: std.mem.Allocator, tree: Ast, base: Ast.TokenIndex) !?[]const u8 {
+pub fn getDocCommentsBeforeToken(allocator: std.mem.Allocator, tree: Ast, base: Ast.TokenIndex, format: types.MarkupKind) !?[]const u8 {
     const tokens = tree.tokens.items(.tag);
     const doc_comment_index = getDocCommentTokenIndex(tokens, base) orelse return null;
-    return try collectDocComments(allocator, tree, doc_comment_index, false);
+    return try collectDocComments(allocator, tree, doc_comment_index, format, false);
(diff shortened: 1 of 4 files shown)


FnControlOption:
zls `master` has the same format as Zig autodocs: https://ziglang.org/documentation/master/std/#A;std:HashMapUnmanaged

It'll be a lot of work, but docs in `std` should be updated to look like either:

```zig
/// - A HashMap based on open addressing and linear probing.
/// - A lookup or modification typically occurs only 2 cache misses.
/// - No order is guaranteed and any modification invalidates live iterators.
/// - It achieves good performance with quite high load factors (by default,
///   grow is triggered at 80% full) and only one byte of overhead per element.
/// - The struct itself is only 16 bytes for a small footprint. This comes at
///   the price of handling size with u32, which should be reasonable enough
///   for almost all uses.
/// - Deletions are achieved with tombstones.
```

or:

```zig
/// A HashMap based on open addressing and linear probing.
///
/// A lookup or modification typically occurs only 2 cache misses.
///
/// No order is guaranteed and any modification invalidates live iterators.
///
/// It achieves good performance with quite high load factors (by default,
/// grow is triggered at 80% full) and only one byte of overhead per element.
///
/// The struct itself is only 16 bytes for a small footprint. This comes at
/// the price of handling size with u32, which should be reasonable enough
/// for almost all uses.
///
/// Deletions are achieved with tombstones.
```

PR #1317, opened 2023-07-12
support argument placeholder

### LINKED ISSUE #685: Feature flag request: Separate argument injection from snippets

jiacai2050:
Close #685

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,7 @@ The following options are currently available.
 | Option | Type | Default value | What it Does |
 | --- | --- | --- | --- |
 | `enable_snippets` | `bool` | `true` | Enables snippet completions when the client also supports them |
+| `enable_argument_placeholders` | `bool` | `false` | Whether to enable function argument placeholder completions |
 | `enable_ast_check_diagnostics` | `bool` | `true` | Whether to enable ast-check diagnostics |
 | `enable_autofix` | `bool` | `true` | Whether to automatically fix errors on save. Currently supports adding and removing discards. |
 | `enable_import_embedfile_argument_completions` | `bool` | `true` | Whether to enable import/embedFile argument completions |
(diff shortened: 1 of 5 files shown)


leecannon:
You need to add the option to `src/config_gen/config.json` then run `zig build gen`.

Also the default should be true to match the current behaviour.

jiacai2050:
PTAL

leecannon:
This isn't working correctly.

Below is the result of completion of `std.debug.print`, the text is what is actually inserted into the editor:

| | `enable_argument_placeholders` true | `enable_argument_placeholders` false |
|-|-|-|
| `enable_snippets` true  | `std.debug.print(comptime fmt: []const u8, args: anytype)`| `std.debug.print(${})` |
| `enable_snippets` false | `std.debug.print(${1:comptime fmt: []const u8}, ${2:args: anytype})` | `std.debug.print(${})` |

Example:
image

This is what should be inserted into the editor:
| | `enable_argument_placeholders` true | `enable_argument_placeholders` false |
|-|-|-|
| `enable_snippets` true  | `std.debug.print(comptime fmt: []const u8, args: anytype)` | `std.debug.print()` |
| `enable_snippets` false | `std.debug.print` | `std.debug.print`|

So the only combination working as expected is when both are true.

jiacai2050:
Ooops, I have make changes to make result right, but I have one question, 

when enable_snippets = true, enable_argument_placeholders = false, if we output `std.debug.print()`, cursor will be put after `)`

If we output `std.debug.print(${})`, cursor will be put after `(`, which is more reasonable to me.

leecannon:
Correct with snippets but not argument placeholders, it should be `std.debug.print(^)` (`^` used to signify cursor location after accepting completion) but what was happening was `std.debug.print(${})^` with the `${}` actually inserted into the document and not handled specially by the editor.

I haven't check the behaviour of the latest commit, as I'm still at work and should not even being looking at this right now :),

leecannon:
I've added a commit which ensures the placement of the cursor is correct when argument placeholders are disabled but snippets are enabled.

This PR is looking good to merge however it won't be merged until #1311 is to prevent more conflicts building up on that PR, don't worry though it should only be a couple of days.

leecannon (approved):

PR #1320, opened 2023-07-12
Enable hover for `var @"foo-bar"`

FnControlOption:
For example:

```zig
var @"foo-bar" = "foo";

const Foobar = struct { @"foo \" bar": i32 };
var foobar = Foobar{ .@"foo \" bar" = 123 };
```

**Warning:** Other than that there doesn't seem to be any noticeable slowdown on my potato, I don't know what effect this has on performance... Guidance on how to benchmark this would be much appreciated!

### DIFF
--- a/src/analysis.zig
+++ b/src/analysis.zig
@@ -3242,20 +3242,29 @@ pub fn resolveExpressionTypeFromAncestors(
 }
 
 pub fn identifierFromPosition(pos_index: usize, handle: DocumentStore.Handle) []const u8 {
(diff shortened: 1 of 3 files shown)


leecannon:
Hover works on `other` but not on `@"foo \" bar"`:

```zig
pub fn main() !void {
    var t: Foobar = undefined;
    _ = t.@"foo \" bar";
    _ = t.other;
}

const Foobar = struct {
    /// Test
    @"foo \" bar": i32,

    /// Test
    other: i32,
};
```

leecannon (approved):

PR #1321, opened 2023-07-13
transition from tres to std.json

Techatrix:
attempt number 2 after #1309
I completely lost my mind trying to reduce a miscompilation.


SuperAuguste:
Wow, that miscompilation is insane, nice find!

PR #1325, opened 2023-07-14
ignore unknown fields while parsing

Techatrix:
neovim seems to have introduced its own LSP Specification that includes addition properties like [`hierarchicalWorkspaceSymbolSupport`](https://github.com/neovim/neovim/blob/9176b5e10a6b32ff65c8ba3f532e3bd55c168ec6/runtime/lua/vim/lsp/protocol.lua#L847)


JohnCoconut:
Is this also happening in emacs' `lsp-mode`? I have the same problem.

leecannon (approved):

PR #1331, opened 2023-07-16
Search for r_brace at the AST level, not the source code level.

gauzey:
We were doing a search for the right brace at the source code level in makeScopeAt. This was fallible, since it only searched up to 16 characters from the l_brace. Another problem was that this search was not being done for any parent function scopes. So the parent scope would end up being shorter than its child block, causing any declaration searches within the block to terminate early.

It seems like this was a known issue based on the TODO comment in the handler for block_two nodes in makeScopeAt.

Before this change, this would fail to trigger any completions on the period after self:
```
fn doThing(self: Self) void {
    doSomethingElseHere(self.
}
```
This change fixes that. It also fixes this issue:
```
const pipeline = struct {
    const PolygonMode = enum(c.enum_onyx_polygon_mode) {
        fill = c.ONYX_POLYGON_MODE_FILL,
        line = c.ONYX_POLYGON_MODE_LINE,
        point = c.ONYX_POLYGON_MODE_POINT,
    };

    pub fn basic(polygon_mode: Poly
};

```
Here, I was not getting any completion suggestions after typing the `y` in Poly.

The fix I'm using is to just find the left brace token for the block, and then just search for the corresponding right brace. It's brain dead simple, which leads me to believe there might be a reason it was not already being use. Grammatically, I think it is a reasonable assumption that a scope that opens on an l_brace will end on its corresponding r_brace. But, I'm still learning zig, and there may be some language constructs I'm not familiar with that break this assumption.

I'm still testing this locally. I currently only have this change applied to block_two and container_decl_two, which are so far the two situations I've encountered in my coding where the old behavior was not working and this change has fixed it (the tow examples above). As I code more, I may find other scope types where it will make sense to extend this handling change to them. 

I don't expect this PR to go through as it stands, as I would like to know if it is a sound change from those who know the language better and what other scopes it might apply to.

I'm mainly submitting this PR at the moment for visibility, feedback, and to highlight the issue.

### DIFF
--- a/src/ast.zig
+++ b/src/ast.zig
@@ -6,6 +6,7 @@ const std = @import("std");
 const offsets = @import("offsets.zig");
 const Ast = std.zig.Ast;
 const Node = Ast.Node;
+const TokenTag = std.zig.Token.Tag;
 const full = Ast.full;
 
 fn fullPtrTypeComponents(tree: Ast, info: full.PtrType.Components) full.PtrType {
@@ -52,6 +53,32 @@ fn fullPtrTypeComponents(tree: Ast, info: full.PtrType.Components) full.PtrType 
         }
     }
     return result;
+}
+
+fn findMatchingRBrace(token_tags: []const TokenTag, l_brace: Ast.TokenIndex) Ast.TokenIndex {
+    std.debug.assert(token_tags[l_brace] == TokenTag.l_brace);
+
+    const start = l_brace + 1;
+    var depth: i32 = 0;
+    var offset: Ast.TokenIndex = 0;
+
+    for (token_tags[start..], 0..) |tag, i| {
+        if (tag == TokenTag.l_brace) {
+            depth += 1;
+        }
+        if (tag == TokenTag.r_brace) {
+            if (depth == 0) {
+                offset = @intCast(i);
+                break;
+            }
+            depth -= 1;
+        }
+    }
+
+    std.debug.assert(depth == 0);
+    std.debug.assert(token_tags[start + offset] == std.zig.Token.Tag.r_brace);
+
+    return start + offset;
 }
 
 pub fn ptrTypeSimple(tree: Ast, node: Node.Index) full.PtrType {
@@ -657,11 +684,15 @@ pub fn lastToken(tree: Ast, node: Ast.Node.Index) Ast.TokenIndex {
             }
             n = datas[n].rhs;
         },
+        .block_two => {
+            return findMatchingRBrace(token_tags, main_tokens[n]);
+        },
+        .container_decl_two => {
+            return findMatchingRBrace(token_tags, main_tokens[n] + 1);
+        },
         .array_init_dot_two,
-        .block_two,
         .builtin_call_two,
         .struct_init_dot_two,
-        .container_decl_two,
         .tagged_union_two,
         => {
             if (datas[n].rhs != 0) {
@@ -673,15 +704,9 @@ pub fn lastToken(tree: Ast, node: Ast.Node.Index) Ast.TokenIndex {
             } else {
                 switch (tags[n]) {
                     .array_init_dot_two,
-                    .block_two,
                     .struct_init_dot_two,
                     => end_offset += 1, // rbrace
                     .builtin_call_two => end_offset += 2, // lparen/lbrace + rparen/rbrace
-                    .container_decl_two => {
-                        var i: u32 = 2; // lbrace + rbrace
-                        while (token_tags[main_tokens[n] + i] == .container_doc_comment) i += 1;
-                        end_offset += i;
-                    },
                     .tagged_union_two => {
                         var i: u32 = 5; // (enum) {}
                         while (token_tags[main_tokens[n] + i] == .container_doc_comment) i += 1;
--- a/src/ast.zig
+++ b/src/ast.zig
@@ -55,7 +55,10 @@ fn fullPtrTypeComponents(tree: Ast, info: full.PtrType.Components) full.PtrType 
     return result;
 }
 
-fn findMatchingRBrace(token_tags: []const TokenTag, l_brace: Ast.TokenIndex) Ast.TokenIndex {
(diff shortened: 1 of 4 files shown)


gauzey:
I will look into that failing test.

gauzey:
Added brace matching for switch statements.

src/ast.zig
```diff
@@ -54,6 +55,32 @@ fn fullPtrTypeComponents(tree: Ast, info: full.PtrType.Components) full.PtrType
     return result;
 }
 
+fn findMatchingRBrace(token_tags: []const TokenTag, l_brace: Ast.TokenIndex) Ast.TokenIndex {
+    std.debug.assert(token_tags[l_brace] == TokenTag.l_brace);
+
+    const start = l_brace + 1;
+    var depth: i32 = 0;
```
Techatrix, line 62: Is it necessary to allow negative depth? Would it make sense to break early if there are too many closing braces?

Techatrix (changes requested):
This is definitely closer to what I envisioned when I added the 16 characters search.

Can you add a completion tests in [completion.zig](https://github.com/zigtools/zls/blob/master/tests/lsp_features/completion.zig) so that we don't regress this behavior.

src/ast.zig
gauzey: Yes, a negative depth would be a syntactic error and we could end on that. I can make that change.

src/ast.zig
```diff
@@ -54,6 +55,32 @@ fn fullPtrTypeComponents(tree: Ast, info: full.PtrType.Components) full.PtrType
...
+            if (depth == 0) {
+                offset = @intCast(i);
+                break;
+            }
+            depth -= 1;
+        }
+    }
+
+    std.debug.assert(depth == 0);
+    std.debug.assert(token_tags[start + offset] == std.zig.Token.Tag.r_brace);
```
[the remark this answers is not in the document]
gauzey: Yes, I'm glad you brought this up since I forgot to ask about it. `lastToken` does not return an error, so I was not sure how to fail gracefully in these failure situations. Is there some null-like token index that should be returned? Should `lastToken` be modified to return an error? Should I just return the last token I see and hope for the best?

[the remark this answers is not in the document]
Techatrix: changing the return type of `lastToken` should probably be avoided because it is used in many places. If you were to return null or an error, the caller of `lastToken` would then need to figure out a suitable last token which isn't easy to do. Given that this only happens on code that is mostly syntactical nonsense, there is no harm in returning a "bad" result from `lastToken`.
Here is my sugesstion: 
change the return type of `findMatchingRBrace` to `?Ast.TokenIndex` and fall back to the previous implementation in `lastToken` if `findMatchingRBrace` returns null.
Returns the very last token of the source file would also be possible.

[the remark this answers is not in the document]
gauzey: I agree, there is not really much the caller could do in the case of broken syntax even if they got an error. I think I favor returning the last token vs falling back to the old implementation. Neither one is going to do the right thing, but the former is at least a little more clear about that.

src/ast.zig
gauzey: Oh, nevermind, I looked at the code and there is no way for there to be a negative depth. Depth starts at 0 and is only ever decremented when we see an r_brace, but not before checking if it is 0 and breaking if it is.

gauzey:
Looks like a fuzz test failed. Not quite sure how to reproduce it locally to debug.

gauzey:
Oh, it looks like `sus` failed to build

gauzey:
Anyway, I've added regressions tests. A nice perk was that I was also able to uncomment two other tests which now pass. Let me know if there is anything else, otherwise I think I took care of everything.

src/ast.zig
Techatrix: Shouldn't the type be changed to `u32` or `usize` then?

src/ast.zig
```diff
@@ -54,6 +55,32 @@ fn fullPtrTypeComponents(tree: Ast, info: full.PtrType.Components) full.PtrType
     return result;
 }
 
+/// Given an l_brace, find the corresponding r_brace.
+/// If no corresponding r_brace is found, return null.
+/// Useful for finding the extent of a block/scope if the syntax is valid.
+fn findMatchingRBrace(token_tags: []const TokenTag, l_brace: Ast.TokenIndex) ?Ast.TokenIndex {
+    std.debug.assert(token_tags[l_brace] == TokenTag.l_brace);
```
Techatrix, line 62: You can avoid having to specify `TokenTag` if you instead do `token_tags[l_brace] == .l_brace`.
This is just nitpicking, so feel free to just ignore this comment. You said that you were still learning zig so I though I would let you know if you didn't know about this already.

Techatrix (changes requested):

src/ast.zig
gauzey: I decided to leave it as `i32` as a bit of protection against any future changes to the code that might make some use of negative values, in which case an unsigned integer type left as is could lead to a negative wrap around bug. That seemed to outweigh the extra precision of going to `u32`, since more than a billion tokens in a file unlikely, and even then I'd probably opt for `isize`.

I know zig has some protection against unsigned wrap around in some build modes, but coming from a C++ background I fear bugs related to unsigned wrapping, and I generally favor signed types if the integer is used for any arithmetic more complicated than simple incrementing.

That is my reasoning. I'll change it if you think it should still be changed.

src/ast.zig
gauzey: Ah good to know! Thank you for pointing it out.

src/ast.zig
Techatrix: In contrast to C++, signed/unsigned integer overflow is safety-checked undefined behavior which means that you can safely choose the integer type that most closely represents your use-case and don't have to worry about UB or overflow bugs.
That being said, I'm starting to bikeshed so how about I approve this PR and we can go on with our lives. :)

Techatrix (approved):

Issue #1341, opened 2023-07-21
Support for type detection for if expression


sgwong:
zls version: 0.11.0-dev.599+ed85a30

the zls version above doesn't support for certain type detection for if expression, check the sample below.

```zig
const a: u32 = 5;
const b: u32 = a + 5;
const at = true;
const bt = false;
const StA = struct {
    a: u32,
    b: bool,
    pub fn init() StA {
        return StA{ .a = 5, .b = true };
    }
};
fn StB() type {
    return struct { a: u32, b: bool };
}
const StC = struct {
    a: u32,
    b: bool,

    pub fn init() StC {
        return StC{ .a = 5, .b = true };
    }

    pub fn init2() StC {
        return StC{ .a = 6, .b = true };
    }
};

const strA = "abc";
const strB = "abc";

const f = if (b > 5) a else b; //this works: f:u32

const c = if (b > 5) true else false; //not working

const d = if (b > 5) a else 5; //not working

const e = if (b > 5) strA else strB; //not working

const ct = if (b > 5) at else bt; //not working

const cst = if (b > 5) StA{ .a = 5, .b = true } else StA{ .a = 6, .b = false }; //this works: Cst: StA

const dst = if (b > 5) StA.init() else StA{ .a = 6, .b = false }; //this works: dst: StA

const est = if (b > 5) StC{ .a = 6, .b = false } else StC.init(); //this works: est: StC

const fst = if (b > 5) StC.init() else StC.init2(); //this works: fst: StC

const gst = StC.init();
const hst = StC.init2();

const ist = if (b > 5) StC.init() else gst; //this works: ist: StC

const jst = if (b > 5) StB(){ .a = 5, .b = true } else StB(){ .a = 6, .b = false }; //this works: jst: StB()

pub fn main() void {
    _ = f;
    _ = c;
    _ = d;
    _ = e;
    _ = ct;
    _ = cst;
    _ = dst;
    _ = est;
    _ = fst;
    _ = gst;
    _ = hst;
    _ = ist;
    _ = jst;
}

```

PR #1342, opened 2023-07-21
Simplify resolved types of primitive values

FnControlOption:
## Simplify resolved types of primitive values

Instead of storing a primitive value as `Type.data.other`, we can just directly map it to `Type.data.primitive`. This allows `true` and `false` to coerce into `bool`.

Partially addresses #1341

```zig
const a: u32 = 5;
const b: u32 = a + 5;
const c = if (b > 5) true else false; // bool
```

## Fix resolved types of number literals

```zig
const a = true;
const b = if (a) 1 else 2; // comptime_int
const c = if (a) 'a' else 'b'; // comptime_int
const d = if (a) 1.2 else 3.4; // comptime_float
```

Technically, these should be resolved by the comptime interpreter, but this works for now.

## Resolve type of enum in switch capture

```zig
const E = enum { a, b, c };
const e: E = .a;
const f = switch (e) {
    .a, .b => |foo| foo, // E
    .c => |bar| bar, // E
};
```

## Resolve type of continue/break/return/unreachable (noreturn)

```zig
const std = @import("std");
pub fn main() void {
    std.log.info("{}", .{@TypeOf(if (true) return else unreachable)});
    const x = if (true) return else unreachable; // noreturn
    _ = x;
}
```

## Add hover for `slice.len` and `slice.ptr`

Screenshot 2023-07-23 at 5 55 21 AM

Screenshot 2023-07-23 at 5 55 32 AM

## Resolve type of index capture in `for` loop

Screenshot 2023-07-23 at 6 12 33 AM

## Resolve type of tag capture on `inline` prong of `switch`

```zig
const U = union(enum) { a: i32, b: i32, c: i32 };
const u: U = .{ .a = 42 };
const g = switch (u) {
    inline .a, .b => |_, foo| foo, // @typeInfo(U).Union.tag_type.?
    inline .c => |_, bar| bar, // @typeInfo(U).Union.tag_type.?
};
```

### DIFF
--- a/src/Server.zig
+++ b/src/Server.zig
@@ -1117,9 +1117,7 @@ pub fn generalReferencesHandler(server: *Server, arena: std.mem.Allocator, reque
         .field_access => |loc| z: {
             const held_loc = offsets.locMerge(loc, name_loc);
             const a = try analyser.getSymbolFieldAccesses(arena, handle, source_index, held_loc, name);
-            if (a) |b| {
-                if (b.len != 0) break :z b[0];
-            }
+            if (a.declarations.len != 0) break :z a.declarations[0];
 
             break :z null;
         },
--- a/src/analyser/primitive.zig
+++ b/src/analyser/primitive.zig
@@ -1,0 +1,193 @@
+const std = @import("std");
+
+pub const PrimitiveType = union(enum) {
+    uint: u16,
+    int: u16,
+    null,
+    undefined,
+    isize,
+    usize,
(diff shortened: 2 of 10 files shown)


leecannon:
@FnControlOption Can you fix the conflicts and get this compiling on 0.11 then I will review this and try to get it merged before we cut 0.11.

Techatrix:
This PR has gone stale.

PR #1346, opened 2023-07-23
Fix snippet being added to parenthesised function completion

SuperAuguste:
Hacky but here's the premise
Assume function `fn bruh(a: u8) void`, and `|` is the cursor.

```zig
// old behavior
br| -> bruh(a: u8)
br|( -> bruh(a: u8)(

// new behavior
br| -> bruh(a: u8)
br|( -> bruh(
```


SuperAuguste:
ahh fuzzer issue (one among many) - need to give sus some love :'(

Techatrix (approved):

PR #1347, opened 2023-07-24
only run ci in zigtools org

jiacai2050:


SuperAuguste:
What does this fix? Neither of these should fail on non zigtools repos as far as I know.

jiacai2050:
When I fork zls, I don't expect CI to be enabled by default, this is because zls is tracking Zig master, so the CI might get broken, and I will receive an email for that.

This is the reason I submit this PR.

SuperAuguste:
That makes sense, thanks.

SuperAuguste (approved):

PR #1348, opened 2023-07-24
fix single threaded build

Techatrix:
How about we continue handling requests after the first one? I know, revolutionary!


SuperAuguste (approved):
😂

PR #1361, opened 2023-07-29
build on save diagnostics

Techatrix:
just like vscode-zig's build on save feature but integrated into ZLS.


leecannon:
The only annoying thing about this is making any modification removes all of the diagnostics, unlike with the extension where the diagnostics stick around until the next build.

leecannon (approved):

PR #1374, opened 2023-08-02
Updates Nix lock to newer compiler version.

Techatrix:
I just installed nix and ran `nix flake update`. @acristoffers did I do this correctly?

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -107,11 +107,11 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1690327932,
-        "narHash": "sha256-Fv7PYZxN4eo0K6zXhHG/vOc+e2iuqQ5ywDrh0yeRjP0=",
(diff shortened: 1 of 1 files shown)


acristoffers:
For the compiler yes, that's correct. If the dependencies in the zon file changed then their new addresses need to be manually copied to the nix file.

acristoffers:
Ah, and just a tip: always run `nix build` after that to make sure the build actually works. I'm totally guilty of not doing it and having to do back-to-back pull requests to fix things.

sagehane:
If the goal is only to update the lock-file, consider `nix flake update --commit-lock-file`. It makes for an idiomatic commit message that is fairly descriptive. https://github.com/zigtools/zls/commit/edf83bcf008559a8c6c4d63982990df013c85b5d for an example.

leecannon (approved):

Issue #1376, opened 2023-08-02
hover msg has something extra at the end


xphoniex:
here's zig/zls:

![Screenshot from 2023-08-02 06-55-44](https://github.com/zigtools/zls/assets/14107758/6ee72079-e520-4bd7-9c37-8f7eb7ec80af)

```
  // fn main() !void (fn () !void)
pub fn main() !void {
```

here's rust/rust-analyzer:

![Screenshot from 2023-08-02 06-56-36](https://github.com/zigtools/zls/assets/14107758/dfed274c-da97-47d5-8832-e96166ae1721)

```
  // fn double_it(num: &mut u32)
fn double_it(num: &mut u32) {
```

is the part in parenthesis `(fn () !void)` necessary?

PR #1377, opened 2023-08-02
Remove the extra part from hover message

xphoniex:
Should fix https://github.com/zigtools/zls/issues/1376.

Please let me know if there should be any cases where we have to show `resolved_type_str` too.

### DIFF
--- a/src/features/hover.zig
+++ b/src/features/hover.zig
@@ -30,6 +30,8 @@ pub fn hoverSymbol(
     var reference_collector = Analyser.ReferencedType.Collector.init(&type_references);
     var doc_str = original_doc_str orelse try decl_handle.docComments(arena);
 
+    var is_fn = false;
+
     const def_str = switch (decl_handle.decl.*) {
         .ast_node => |node| def: {
             if (try analyser.resolveVarDeclAlias(.{ .node = node, .handle = handle })) |result| {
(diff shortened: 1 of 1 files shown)


Techatrix:
The resolved_type generally should be shown. Hovering over functions should be the exception where it's unnecessary. The hover when markdown is enabled should also be updated.

xphoniex:
Can you give me a few examples (in code), what should show and what shouldn't, so I can test locally before pushing the fix?

xphoniex:
Check https://github.com/zigtools/zls/pull/1377/commits/a77cac1ba3ba8cd66d471285f835c8432a33ec05 :

Screenshot from 2023-08-02 10-15-46

Screenshot from 2023-08-02 10-16-25

src/features/hover.zig
```diff
@@ -122,7 +129,11 @@ pub fn hoverSymbol(
             try writer.print("[{s}]({s}#L{d})", .{ ref.str, ref.handle.uri, line });
         }
     } else {
-        try writer.print("{s} ({s})", .{ def_str, resolved_type_str });
+        if (is_fn) {
+            try writer.print("{s}", .{def_str});
+        } else {
+            try writer.print("{s} ({s})", .{ def_str, resolved_type_str });
```
Techatrix, line 135: I know this is unrelated to your fix but while were modifying hover. I think there should be a line break between the definition string and resolved type just like in the markdown version.

xphoniex: [diff](https://github.com/zigtools/zls/compare/a77cac1ba3ba8cd66d471285f835c8432a33ec05..5d84a617b9bd714735bc45f68098ef210ae58edb)

PR #1382, opened 2023-08-04
update usage of std.zig.system.NativePaths

Techatrix:
final breaking change in zig 0.11.0


leecannon (approved):

leecannon:
Took a while to convince the zig fetching action to actually grab the latest tarball https://github.com/goto-bus-stop/setup-zig/issues/18#issuecomment-1665240334.

PR #1393, opened 2023-08-08
explore duplicate packages in build runner

Techatrix:
This will make it easier to get stuck in cycles in the build runner. But in my defense, I think before It was still easily doable to create a build.zig that will loop.
fixes #1388

PR #1394, opened 2023-08-08
build_runner:  recurse into dependant steps and handle GeneratedFile correctly

leecannon:
Resolves https://github.com/zigtools/zls/pull/1393#issuecomment-1670145228


nullptrdevs:
Nifty :+1:

Techatrix (approved):
Just like nullptrdevs sugested, maybe we should also consider running depedent steps.

PR #1397, opened 2023-08-10
fix typo and formatting ts example in the `Range` document.

sina-devel:


Techatrix:
`lsp.zig` is generated by [zig-lsp-codegen](https://github.com/zigtools/zig-lsp-codegen) using a [metaModel.json](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/metaModel/metaModel.json) that was probably created from typescript definitions because `Range` has this typo [here](https://github.com/microsoft/vscode-languageserver-node/blob/07b9d293ea775dfd263562da3e8321d40397a3d0/types/src/main.ts#L177C17-L177C17). This appears to be the original source of the typo that should be corrected orelse this typo can easily reemerge when we regenerate `lsp.zig`.

PR #1408, opened 2023-08-15
config related fixes

Techatrix:
reverted b47781897acca54b3fffbe052c6daf3ab08fe8b0 because realpath should be avoided.
reverted eb00a8ab15c40ac11b75da3c24886cdcd4680346 because config options still need to happen after startup because as an example you may not have zig set in path


SuperAuguste (approved):
thank u :)

PR #1409, opened 2023-08-16
generate version data file in build phase

Techatrix:
Instead of having to run `zig build gen` whenever builtins get changed so that [master.zig](https://github.com/zigtools/zls/blob/master/src/data/master.zig) is updated, I thought that this could be converted into a build step.

I am not entirely satisfied with how the solution ended up in `build.zig` because caching prevents the version data from being updated whenever the website changes but never caching forces you to always have a internet connection when you build ZLS.

What I did is to invalidate the cache once every day based on the current timestamp which feels more like a hacky trick rather than a proper solution.
https://github.com/zigtools/zls/blob/b55525e961078b08e7e8a4f55c08090041560e1a/build.zig#L147-L153

Some feedback on this would be appreciated. Or just tell me to scrap this idea which is also fine.

PR #1410, opened 2023-08-17
nix: add missing dependency on zig language reference file

Techatrix:
I broke nix build in #1409 because you can't just download a file without declaring it as a dependency.

@acristoffers I hope you don't mind the ping. Did I do this correctly?
The `langref.html.in` file dependency would need to be manually updated whenever it changes. Don't know if there is a better way to make it easier to update. I thought about just depending on the `zig-lang/zig` repo which contains `langref.html.in` but that seems to be a bit wasteful just to get a single file.


acristoffers:
Hi @Techatrix. I don't mind pinging ^^

Yep, that's correct. I thought about getting it from zig-overlay, but they don't pull the source, they get the prebuilt binaries, so no help there either. I also don't like it that we have to keep updating hashes to get things, but for now I'm out of ideas of how to do it better :( And this particular case is even worse, as we have don't have this file in the .zon, so it is yet another source to check.

leecannon (approved):

PR #1426, opened 2023-08-27
use std.time.Timer instead of timestamps

Techatrix:


leecannon (changes requested):
`Timer.read` returns nanoseconds not milliseconds.

leecannon (approved):

PR #1433, opened 2023-09-05
improve type resolution on branching types

Techatrix:
Here is an example of what can now be resolved which previously yielded no completions:
showcase

### DIFF
--- a/src/analysis.zig
+++ b/src/analysis.zig
@@ -1710,8 +1710,14 @@ pub const TypeWithHandle = struct {
         var deduplicator = Deduplicator{};
         defer deduplicator.deinit(allocator);
 
-        for (entries) |e|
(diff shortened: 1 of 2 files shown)


xdBronch:
i found some undesirable behavior with this when it tries to resolve large/complex types. for example when trying to get autocomplete for `std.os.system` ZLS freezes, uses nearly 9gb of ram, then dies. i'm not sure how it could be worked around because this feature seems very useful otherwise.

Techatrix:
The issue with `std.os.system` is caused by analysis being stuck in an infinite loop. This will be relatively easy to resolve. If you have any more issues, please let be know.

leecannon:
I've been using this PR locally for about a week and have not seen any issues, will merge once CI is unblocked.

leecannon (approved):

PR #1444, opened 2023-09-16
Inlay Type Hints in Variable Declarations

### LINKED ISSUE #318: [Feature Request] Type inference in editor
Hi,
I think it would be a great idea to add a feature where zls / ZLS for VSCode could show the type inferred to certain variables like how rust-analyzer does it : 
image
(Notice the "i32")

Koranir:
Screenshot from 2023-09-17 19-03-00

Adds inlay hints to variable declarations - useful for making sure a type is what _you_ think _zls_ thinks it is without having to hover over a variable name.
fixes #318

### DIFF
--- a/src/features/inlay_hints.zig
+++ b/src/features/inlay_hints.zig
@@ -212,6 +212,7 @@ fn writeBuiltinHint(builder: *Builder, parameters: []const Ast.Node.Index, argum
     }
 }
 
+/// Takes a variable declaration AST node. If the type is inferred, attempt to infer it and display it as a hint.
 fn writeVariableDeclHint(builder: *Builder, decl_node: Ast.Node.Index) !void {
     const tracy_zone = tracy.trace(@src());
     defer tracy_zone.end();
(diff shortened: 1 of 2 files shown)


Koranir:
Can confirm that #1445 solves the issue with inferring types.

src/features/inlay_hints.zig
```diff
@@ -188,29 +189,82 @@ fn writeBuiltinHint(builder: *Builder, parameters: []const Ast.Node.Index, argum
...
+    if (try decl_handle.resolveType(builder.analyser)) |resolved_type| {
+        try builder.analyser.referencedTypes(
+            resolved_type,
+            &type_str,
+            &reference_collector,
+        );
+    }
+
+    const hint = tree.fullVarDecl(decl_node) orelse return;
+    if (hint.ast.type_node == 0) {
```
Techatrix, line 251: you can avoid one level of indentation by doing an early return. You should also move this check closer to the start of the function to avoid needlessly computing the inlay hint and then discarding it.

Techatrix (changes requested):
This will need some test to make sure it is working correctly and avoid future regressions.
I've added two commits to your branch that should allow you to write some new tests [like this](https://github.com/zigtools/zls/blob/a630c609d446a810f809a03d2ba3e4383d69ac0b/tests/lsp_features/inlay_hints.zig#L91-L95).
If there is anything that I can do to help, don't hesitate to ask.

src/features/inlay_hints.zig
```diff
@@ -188,29 +189,82 @@ fn writeBuiltinHint(builder: *Builder, parameters: []const Ast.Node.Index, argum
...
+    ) orelse return;
+
+    var type_str: []const u8 = "unknown";
+    if (try decl_handle.resolveType(builder.analyser)) |resolved_type| {
+        try builder.analyser.referencedTypes(
+            resolved_type,
+            &type_str,
+            &reference_collector,
+        );
+    }
```
Techatrix, line 248: You can probably achieve the same with `resolveTypeOfNode`.

```suggestion
    const resolved_type = (try builder.analyser.resolveTypeOfNode(.{ .node = decl_node, .handle = handle })) orelse return;

    var type_references = Analyser.ReferencedType.Set.init(builder.arena);
    var reference_collector = Analyser.ReferencedType.Collector.init(&type_references);

    var type_str: []const u8 = "unknown";
    try builder.analyser.referencedTypes(
        resolved_type,
        &type_str,
        &reference_collector,
    );
```

src/features/inlay_hints.zig
```diff
@@ -211,6 +212,56 @@ fn writeBuiltinHint(builder: *Builder, parameters: []const Ast.Node.Index, argum
...
+        decl_node,
+    ) orelse return;
+    const source_index = offsets.tokenToIndex(tree, tree.firstToken(decl_node));
+    const decl_handle = try builder.analyser.lookupSymbolGlobal(
+        handle,
+        decl_name,
+        source_index,
+    ) orelse return;
+
+    var type_str: []const u8 = "unknown";
```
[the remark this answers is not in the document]
Koranir: > I personally think that showing "unknown" has no informational value.

I think that knowing whether zls recognizes your type or not does have some value, but I suppose there being no hint shown would also convey that.

src/features/inlay_hints.zig
```diff
@@ -188,29 +187,77 @@ fn writeBuiltinHint(builder: *Builder, parameters: []const Ast.Node.Index, argum
...
+    const decl_str = offsets.tokensToSlice(tree, tree.firstToken(hint.ast.init_node) - 2, assign_token);
+    var decl_spaces: usize = 0;
+    for (decl_str) |char| {
+        if (char == ' ') {
+            decl_spaces += 1;
+        }
+    }
+
+    try builder.hints.append(builder.arena, .{
+        .index = offsets.tokenToIndex(tree, assign_token) - decl_spaces,
```
Techatrix, line 251: I should have looked more closely at your code that finds the inlay hint index because you can get the end of a token in the source with `offsets.tokenToLoc(tree, token_index).end`. `hint.ast.mut_token + 1` is always the variable name token.
```suggestion
    try builder.hints.append(builder.arena, .{
        .index = offsets.tokenToLoc(tree, hint.ast.mut_token + 1).end,
```

Techatrix (changes requested):
just one more thing and then we are done. everything else looks great.

src/features/inlay_hints.zig
```diff
@@ -69,23 +69,22 @@ const Builder = struct {
...
             );
-            defer last_index = index;
+            defer last_index = hint.index;
             defer last_position = position;
             converted_hint.* = types.InlayHint{
                 .position = position,
                 .label = .{ .string = hint.label },
                 .kind = hint.kind,
-                .tooltip = .{ .MarkupContent = hint.tooltip },
+                .tooltip = if (hint.tooltip != null) .{ .MarkupContent = hint.tooltip.? } else null,
```
Techatrix, line 85: You can do `if (some_optional) |inner_value| { ... } else { ... }` to conditionally capture the value inside an optional so that you can avoid the `.?` which is slightly less idiomatic.
This is how I would do it:
```suggestion
                .tooltip = if (hint.tooltip) |tooltip| .{ .MarkupContent = tooltip } else null,
```
I don't expect you to do this change. I just wanted you to inform you about this since you said that you are new to zig and you may not have known about this language feature.

Koranir: Thanks!

Techatrix (approved):

Issue #376, opened 2021-08-17
enum autocomplete suggestions are wrong


Jarred-Sumner:
For `StdIo` in `child_process.zig`, the possible values are:
```zig
pub const StdIo = enum {
    Inherit,
    Ignore,
    Pipe,
    Close,
};
```

However, the suggested values are `Error` and `Warn` -- neither of which are valid values.

<img width="774" alt="CleanShot 2021-08-16 at 18 07 48@2x" src="https://user-images.githubusercontent.com/709451/129647836-fcb3cdf4-226d-4237-8360-41ea98e2c578.png">

This isn't a specific issue with `ChildProcess`. I've noticed that the enum autocomplete suggestions are incorrect more generally. 

Prefixing with the type name works better, however the completion kind displayed are incorrect. 
<img width="554" alt="CleanShot 2021-08-16 at 18 11 29@2x" src="https://user-images.githubusercontent.com/709451/129648048-a4649d2c-896b-4a45-90e7-3b06988d5286.png">

SuperAuguste:
This is just a fundamental design limitation with zls; hopefully we'll be able to chisel away at it with the comptime interpreter.

Techatrix:
The following code sample works in ZLS master. (cc @nullptrdevs)
```zig
const std = @import("std");
fn main() !void {
    var child_process: std.process.Child = undefined;
    child_process.stderr_behavior = .<cursor here>
}
```

PR #1449, opened 2023-09-18
More enum completions

nullptrdevs:
enumc2
enumc1
enumc4
enumc3

Limitations:
Doesn't work for container fields that have an optional enum type (?Enum).
Can be "peeked/previewed" => `exe.wasi_exec_model.? = .`

Known side effects:
Currently `=` and `==`/`!=` are treated the same, ie completions are provided for `f() = .`.


Techatrix:
partially addresses #376

leecannon (approved):

PR #1450, opened 2023-09-20
update known-folders

leecannon:


Techatrix (approved):
I've also updated known folders in flake.nix to keep it in sync with build.zig.zon

PR #1451, opened 2023-09-20
prevent duplicate CI runs

leecannon:
We are running all CI jobs twice on PRs:
image

[This](https://github.com/orgs/community/discussions/26276#discussioncomment-3251140) is the easiest way to stop it.

### DIFF
--- a/.github/workflows/build_runner.yml
+++ b/.github/workflows/build_runner.yml
@@ -2,10 +2,14 @@ name: BuildRunner
 
 on:
   push:
+    branches:
+      - master
     paths:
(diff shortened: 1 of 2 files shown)


SuperAuguste:
Nice catch, this has been bugging be for a bit. What if you want to work on [a branch](https://github.com/zigtools/zls/tree/techatrix/stage2-sema) (:^)) and have the CI run but not necessarily have a PR open for that branch? What about PRs onto non-master branches. This is a surprisingly complex problem hmm.

leecannon:
Yea it feels rather janky.

PR #1453, opened 2023-09-20
make variable declaration and parameter name hints separately configurable

leecannon:
Adds "inlay_hints_show_variable_declaration" and "inlay_hints_show_parameter_name" to allow control of what hints are displayed.


Techatrix:
how about this?
```zig
inlay_hints: enum {
    none,
    only_parameter_names,
    only_types,
    all,
} = .all,
```

leecannon:
I want parameter names but not types 😅

leecannon:
The problem I have with using an enum is the moment there is one more type of inlay hint we have to go back to using multiple booleans anyway or you only have three choices none, all or just one type of hint.

For example I already want to break the variable declaration hints into separate local variable versus global variable options.

Techatrix (approved):
We should also update vscode-zig.

Issue #1235, opened 2023-06-14
zig confirm suggestion "replace" instead of "insert"


haoyuan80s:
### Zig Version

0.11.0-dev.3380+7e0a02ee2

### Zig Language Server Version

0.11.0-dev.479+c28a59c

### Steps to Reproduce

<|>: cursor position
```
fn add(
    c<|>xxx: i32,
) i32 {
    return xxx + 1;
}
```
![image](https://github.com/zigtools/zls/assets/9064146/4574a38d-6fa0-4020-93fe-382b4c69fb57)
confirm the first selection (tab in my nvim keybinding)

x

### Expected Behavior

insert test:
![image](https://github.com/zigtools/zls/assets/9064146/a3011a35-d05a-493a-a4ad-47f9d310c98f)

### Actual Behavior

replace text:
![image](https://github.com/zigtools/zls/assets/9064146/643372dc-0859-47b9-b907-2c7f042455bf)

* Note: it works (aka inserts) if the argument is a single letter (e.g. `x`)

Techatrix:
this conflicts with #907. Maybe this should be configurable.

PR #1458, opened 2023-09-23
Add `completions_with_replace` flag

almmiko:
Adds `completions_with_replace` flag to determine replace behavior.

Fixes: https://github.com/zigtools/zls/issues/1235


SuperAuguste (approved):
Approving this as this is a good compromise, but in the future, we'd like to review how we do completion replacement in a manner acceptable for all so a flag like this may not be necessary.

PR #1461, opened 2023-09-24
separately cache callsite references

Techatrix:
caching callsite references based on the function node index collide with type resolution of the function node itself.


SuperAuguste (approved):
Oops! Nice fix :)

PR #1467, opened 2023-09-25
remove unnecessary config options

Techatrix:
`enable_import_embedfile_argument_completions` existed because this option has been buggy in the past but not anymore.

`include_at_in_builtins` is a workaround for builtin completions in sublime text 3 which can instead be detected by
inspecting the `clientInfo` field in initialize request.

`max_detail_length` is a workaround for long completion detail entries bricking the preview window in Sublime Text #261 which can also be detected through the `clientInfo` field.

`operator_completions` this option has been added when this feature was implemented but is there anyone who needs it?


SuperAuguste (approved):
LGTM

PR #1470, opened 2023-09-25
simplify type resolution on switch payload

Techatrix:
This also allows hovering over a switch case capture on an enum.
```zig
const E = enum { foo };
fn func(e: E) void {
    switch (e) {
        .foo => |b<hover here>ar| {},
    }
}
```


SuperAuguste (approved):
LGTM

PR #1471, opened 2023-09-26
Restrict Maximum Whitespace in Inlay Hints

Koranir:
## Before:
image
## After:
image

### DIFF
--- a/src/features/inlay_hints.zig
+++ b/src/features/inlay_hints.zig
@@ -213,6 +213,25 @@ fn writeBuiltinHint(builder: *Builder, parameters: []const Ast.Node.Index, argum
     }
 }
 
+// Restrict whitespace to only one space at a time.
+fn reduce_string_whitespace(str: []const u8, arena: std.mem.Allocator) ![]const u8 {
+    // Overallocates by a small amount if whitespace is reduced, but it should be fine.
+    var reduced_type_str = try std.ArrayList(u8).initCapacity(arena, str.len);
+    var skip = false;
+    for (str) |char| {
+        if (char == '\n' or char == ' ') {
+            if (!skip) {
+                reduced_type_str.appendAssumeCapacity(' ');
+            }
+            skip = true;
+        } else {
+            reduced_type_str.appendAssumeCapacity(char);
+            skip = false;
+        }
+    }
+    return reduced_type_str.items;
+}
+
 /// Takes a variable declaration AST node. If the type is inferred, attempt to infer it and display it as a hint.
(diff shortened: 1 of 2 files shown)


xdBronch:
imo the total number of characters should be restricted in cases like this, 68 characters dedicated to the type is quite a lot. maybe something like `const thing2: struct {...} = thing();` above a certain threshold? this is definitely an improvement though :+1:

src/features/inlay_hints.zig
```diff
@@ -237,6 +237,24 @@ fn writeVariableDeclHint(builder: *Builder, decl_node: Ast.Node.Index) !void {
     );
     if (type_str.len == 0) return;
 
+    // Restrict whitespace to only one space at a time.
+    var reduced_type_str = std.ArrayList(u8).init(builder.arena);
+    // Overallocates by a small amount if whitespace is reduced, but it should be fine.
+    try reduced_type_str.ensureTotalCapacity(type_str.len);
```
Techatrix, line 243: this can be combined into a call to `std.ArrayListUnmanaged(u8).initCapacity(builder.arena, type_str.len)`

src/features/inlay_hints.zig
```diff
@@ -237,6 +237,24 @@ fn writeVariableDeclHint(builder: *Builder, decl_node: Ast.Node.Index) !void {
...
+    var skip = false;
+    for (type_str) |char| {
+        if (char == '\n' or char == ' ') {
+            if (!skip) {
+                try reduced_type_str.append(' ');
```
Techatrix, line 248: this should call `appendAssumeCapacity`, same goes for the call a few lines below.

Techatrix (changes requested):
I would suggest to refactor this into a separate function.

Detecting long type hints and then reducing them to something like `struct { ... }` can be made as a separate contribution because this change is still better than status quo.

Techatrix (approved):

src/features/inlay_hints.zig
```diff
@@ -213,6 +213,25 @@ fn writeBuiltinHint(builder: *Builder, parameters: []const Ast.Node.Index, argum
     }
 }
 
+// Restrict whitespace to only one space at a time.
+fn reduce_string_whitespace(str: []const u8, arena: std.mem.Allocator) ![]const u8 {
+    // Overallocates by a small amount if whitespace is reduced, but it should be fine.
+    var reduced_type_str = try std.ArrayList(u8).initCapacity(arena, str.len);
```
nullptrdevs, line 219:
```suggestion
    var reduced_type_str = try std.ArrayListUnmanaged(u8).initCapacity(arena, str.len);
```

PR #1476, opened 2023-09-27
Add semicolon when it's missing

### LINKED ISSUE #1475: Feature Request: Add auto semicolon completion
Since zls already auto format the unused variable and for semicolon it exactly gives the correct message in which column semicolon is missing so auto completing this would be a better experience as i am a JS dev so most of the time i forgot to add semicolon and the prettier formatter auto complete that. I guess it would be a great feature.

and

instead of this
```zig
const verices = [_]f32{
        -0.5, -0.5, 0,
        0.5,  -0.5, 0,
        0,    0.5,  0,
    };
```

this would be better
```zig
const verices = [_]f32{
        -0.5, -0.5, 0,
         0.5, -0.5, 0,
         0,    0.5, 0,
    };
```

SuperAuguste:
Closes #1475

Probably needs some fuzzing and manual testing but this is it!

I wonder if applying multiple autofixes at once would be desirable as now this can happen pretty easily (see `var a = 1`, which adds a `;` on the first save and then an unused discard on the second.)

Anyways back to my job I go! 🏃‍♂️

### DIFF
--- a/src/Server.zig
+++ b/src/Server.zig
@@ -339,10 +339,8 @@ fn getAutofixMode(server: *Server) enum {
 
 /// caller owns returned memory.
 fn autofix(server: *Server, arena: std.mem.Allocator, handle: *DocumentStore.Handle) error{OutOfMemory}!std.ArrayListUnmanaged(types.TextEdit) {
-    if (handle.tree.errors.len != 0) return .{};
-
     var diagnostics = std.ArrayListUnmanaged(types.Diagnostic){};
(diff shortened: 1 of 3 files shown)


SuperAuguste:
Is this something anyone actually wants? Not sure if it's useful

Techatrix:
If this feature has no false-positives then it wouldn't hurt, right?

Techatrix (approved):

PR #1478, opened 2023-09-28
apply change on intermediate document in applyContentChanges

Techatrix:
this was a bug that I introduced in #1460


nullptrdevs (approved):
All good.

PR #1482, opened 2023-09-30
resolve config options in tests through the build system

Techatrix:
previously when running tests, the zig path, zig lib path and global cache path where all resolved the same way they would be when running ZLS normally that is by looking in PATH and using known_folders.

This means that as an example the zig version that was used to run `zig build` may not be same as the zig version used by ZLS in its tests. Instead these config options are now resolved through the build system and supplied as an option module.

This achieved my goal of running tests in nix which wasn't possible before.


SuperAuguste (approved):
LGTM

Issue #1230, opened 2023-06-07
Incorrect type inference when slicing a slice with comptime-known range


recursiveGecko:
### Zig Version

0.11.0-dev.3322+82632aff2

### Zig Language Server Version

0.11.0-dev.469+4f0762a

### Steps to Reproduce

Using VS Code with the Zig extension, with the following code:

```zig

const my_slice: []const f32 = &.{1, 2, 3, 4};
std.debug.print("{d:.3}", .{ my_slice[0..2] });
```

Enter a dot after `my_slice[0..2]` to trigger auto-completion.

### Expected Behavior

Since the range is comptime-known, the type of `my_slice[0..2]` is `*[2]f32`, so typing `.` after `my_slice[0..2]` should open an autocomplete window that includes `*`.

### Actual Behavior

Autocompleted suggestions are `len` and `ptr`, but not `*`.

This can be further validated by assigning the subslice to a variable and inspecting its type (by hovering over it), which is reported as `[]const f32`.
```zig
const my_subslice = my_slice[0..2];
```

Techatrix:
ZLS is not capable of resolving comptime-known values so it assumes that values are runtime-known. A special case that would detect a number literal being used as the start and end index when slicing could be added but anything more complex would be harder to implement.

PR #1484, opened 2023-09-30
analysis: support many and c pointer completions

### LINKED ISSUE #1483: many pointer completions
It seems like zls hasn't implemented completions for multipointer indexing.

```zig
const T = struct { foo: u8 };
const t_slice: []T = undefined;
const t_many: [*]T = undefined;
const works = t_slice[0].;
const doesnt_work = t_many[0].;
```

I would like to maybe work on this but I'm not sure where to start.  If anyone wants to coach me a little to help get me started, I will give it a try.  Or if it would just be easier for someone more familiar with the codebase to implement, I will refrain.

travisstaloch:
closes #1483

previously, resolveBracketAccessType() only supported slice pointer types.  i've added supoort for many and c pointer types.  this allows for `manyptr[0].` completions.

i've added new test cases to completion.zig for bracket access to many and c pointers along with array access.  i've also added cases of slicing a many pointer and an optional many pointer in a for loop.

### DIFF
--- a/src/analysis.zig
+++ b/src/analysis.zig
@@ -754,15 +754,18 @@ fn resolveBracketAccessType(analyser: *Analyser, lhs: TypeWithHandle, rhs: enum 
             .handle = lhs.handle,
         };
     } else if (ast.fullPtrType(tree, lhs_node)) |ptr_type| {
-        if (ptr_type.size == .Slice) {
-            if (rhs == .Single) {
-                return ((try analyser.resolveTypeOfNodeInternal(.{
-                    .node = ptr_type.ast.child_type,
-                    .handle = lhs.handle,
-                })) orelse return null).instanceTypeVal();
-            }
-            return lhs;
+        switch (ptr_type.size) {
+            .Slice, .C, .Many => {
+                if (rhs == .Single) {
+                    return ((try analyser.resolveTypeOfNodeInternal(.{
+                        .node = ptr_type.ast.child_type,
+                        .handle = lhs.handle,
+                    })) orelse return null).instanceTypeVal();
+                }
+            },
+            else => {},
         }
+        return lhs;
     }
 
     return null;
--- a/src/analysis.zig
+++ b/src/analysis.zig
@@ -762,10 +762,10 @@ fn resolveBracketAccessType(analyser: *Analyser, lhs: TypeWithHandle, rhs: enum 
                         .handle = lhs.handle,
                     })) orelse return null).instanceTypeVal();
                 }
+                return lhs;
             },
-            else => {},
+            .One => {},
         }
-        return lhs;
     }
 
     return null;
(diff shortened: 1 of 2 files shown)


src/analysis.zig
```diff
@@ -754,15 +754,18 @@ fn resolveBracketAccessType(analyser: *Analyser, lhs: TypeWithHandle, rhs: enum
...
+                if (rhs == .Single) {
+                    return ((try analyser.resolveTypeOfNodeInternal(.{
+                        .node = ptr_type.ast.child_type,
+                        .handle = lhs.handle,
+                    })) orelse return null).instanceTypeVal();
+                }
+            },
+            else => {},
         }
+        return lhs;
```
Techatrix, line 768: this line should probably be inside the switch case because otherwise you will get completions when indexing a single item pointer.

src/analysis.zig
```diff
@@ -754,15 +754,18 @@ fn resolveBracketAccessType(analyser: *Analyser, lhs: TypeWithHandle, rhs: enum
...
+        switch (ptr_type.size) {
+            .Slice, .C, .Many => {
+                if (rhs == .Single) {
+                    return ((try analyser.resolveTypeOfNodeInternal(.{
+                        .node = ptr_type.ast.child_type,
+                        .handle = lhs.handle,
+                    })) orelse return null).instanceTypeVal();
+                }
+            },
+            else => {},
```
Techatrix, line 766: nitpick:
```suggestion
            .One => {},
```

src/analysis.zig
xdBronch, line 768: shouldnt this be inside the first switch case? it was inside the `if (ptr_type.size == .Slice)` before

Techatrix (changes requested):
I also just realized that the behavior when `rhs == .Range and ptr_type.size != .Slice` need to be properly implemented:
```zig
const T = struct {alpha: u32};
const slice: []T = undefined;
const foo = slice[0..1]; // works as expected
const many: [*]T = undefined;
const bar = many[0..1]; // doesn't work, expected []T
const c: [:c]T = undefined;
const baz = many[0..1]; // doesn't work, expected []T
```
Note that the expected behavior above is actually incorrect because slicing with comptime known values should yield an array, but ZLS can't do comptime. see https://github.com/zigtools/zls/issues/1230
Everything suggested here can be left as a follow up issue and doesn't have to be resolved in this PR for approval.

src/analysis.zig
travisstaloch: addressed in new commit

src/analysis.zig
travisstaloch: addressed in new commit

src/analysis.zig
travisstaloch: thanks! looks like agree with Techatrix. addressed in new commit.

Techatrix (approved):

travisstaloch:
> Note that the expected behavior above is actually incorrect because slicing with comptime known values should yield an array, but ZLS can't do comptime. see #1230

Oh i see what you mean. zls thinks that `slice_or_manyptr[0..2]` is a slice when infact its a `*const [2]T`

PR #1492, opened 2023-10-02
lazily compute resources of DocumentStore handles

Techatrix:
Instead creating the DocumentStore (and ZIR if needed) whenever the document gets modified, it is more efficient
to compute them lazily needed. This will mean that series of text edits without any requests on them which often occur while typing will be processed quicker with less latency.

I've also tested lazily computing the Ast but it turned out that the Ast was (almost) always needed which made this unnecessary.

### DIFF
--- a/src/ComptimeInterpreter.zig
+++ b/src/ComptimeInterpreter.zig
@@ -25,7 +25,7 @@ document_store: *DocumentStore,
 uri: DocumentStore.Uri,
 namespaces: std.MultiArrayList(Namespace) = .{},
 
-pub fn getHandle(interpreter: *ComptimeInterpreter) *const DocumentStore.Handle {
+pub fn getHandle(interpreter: *ComptimeInterpreter) *DocumentStore.Handle {
     // This interpreter is loaded from a known-valid handle so a valid handle must exist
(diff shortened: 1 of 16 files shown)


nullptrdevs (commented):
Been meaning to ask, are these intentionally not pointers (completions.zig, ~line 844)?
```zig
fn completeFileSystemStringLiteral(
    arena: std.mem.Allocator,
    store: DocumentStore,
    handle: DocumentStore.Handle,
```

Techatrix:
Ideally speaking, I would argue that pass by value should be preferred over pass by reference/pointer in zig since it gives the compiler more opportunities to optimize the code. Unfortunately in its current state, zig is sometimes failing to generate the most optimal code (#17580) but hopefully this won't be the case in the future.

This has been the main motivation why I did this. Another example of this issue would be the `Config` struct which is sometimes passed as a pointer and sometimes not. Despite that this is still an inconsistency in the codebase and I can change `completeFileSystemStringLiteral` to take its parameters as pointers if you wish.
Many uses of `DocumentStore` and `Handle` require mutability in this PR so pass by value is rarely possible anyway.

nullptrdevs (approved):
Should be gud .

PR #1496, opened 2023-10-05
Rename '@fabs' builtin to '@abs'

antlilja:
This PR renames `@fabs` to `@abs` like the zig compiler has done.


nullptrdevs (approved):
Thanks

PR #1498, opened 2023-10-06
snippets: show top level snippets after typing (first char)

nullptrdevs:
Adds "std" and "import".


src/features/completions.zig
```diff
@@ -1416,7 +1416,8 @@ fn completeFileSystemStringLiteral(
 }
 
 pub fn completionAtIndex(server: *Server, analyser: *Analyser, arena: std.mem.Allocator, handle: *const DocumentStore.Handle, source_index: usize) error{OutOfMemory}!?types.CompletionList {
-    const at_line_start = offsets.lineSliceUntilIndex(handle.tree.source, source_index).len == 0;
+    const line_slice = offsets.lineSliceUntilIndex(handle.tree.source, source_index);
+    const at_line_start = line_slice.len == 0 or line_slice.len == 1;
```
Techatrix, line 1420: do snippet completions not appear for you without this change? I just checked VSCode, ST4 and Helix and it worked fine before.
```suggestion
    const at_line_start = offsets.lineSliceUntilIndex(handle.tree.source, source_index) <= 1;
```

Techatrix: I even get completions without explicitly requesting them:

just typing 'main':
A
explicitly request completions then type 'main':
B
type 'm' the request completions:
C

Techatrix (approved):

PR #1499, opened 2023-10-06
Fix UAF in uriAssociatedWithBuild

SuperAuguste:
Found with comptime interpreter enabled 😅

### DIFF
--- a/src/DocumentStore.zig
+++ b/src/DocumentStore.zig
@@ -703,11 +703,17 @@ fn uriAssociatedWithBuild(
     defer tracy_zone.end();
 
     var checked_uris = std.StringHashMapUnmanaged(void){};
-    defer checked_uris.deinit(self.allocator);
+    defer {
+        var kit = checked_uris.keyIterator();
+        while (kit.next()) |key| {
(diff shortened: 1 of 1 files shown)


Techatrix (changes requested):
Pretty sure this change can create a memory leak because the `package_uri` doesn't always get stored in `checked_uris` when one of these early returns get executed or an error is returned:
- [DocumentStore.zig#L719](https://github.com/zigtools/zls/blob/c8255d47ff98c36bcdcedf1e570f426a0c05ba03/src/DocumentStore.zig#L719)
- [DocumentStore.zig#L738](https://github.com/zigtools/zls/blob/c8255d47ff98c36bcdcedf1e570f426a0c05ba03/src/DocumentStore.zig#L738)
- [DocumentStore.zig#L740](https://github.com/zigtools/zls/blob/c8255d47ff98c36bcdcedf1e570f426a0c05ba03/src/DocumentStore.zig#L740)

SuperAuguste:
oof i totally missed that, thanks 😅

i guess we can use an arena or add it to a big list or something

Issue #1219, opened 2023-05-31
cannot show all struct fields when `usingnamespace` and `@field` together


axojhf:
### Zig Version

0.11.0-dev.3329+1ab008d89

### Zig Language Server Version

5d53f0104fe6742a28592094f9f7de5060a6f07c

### Steps to Reproduce

Clone `https://github.com/ZigEmbeddedGroup/espressif-esp`, using Visual Studio Code to open `src/example/blinky.zig`, the fields of `microzig.chip` can not be fully displayed

### Expected Behavior

show all fields using `pub usingnamespace @field`

### Actual Behavior

Only show the `types` field

nullptrdevs:
38a199813a1018305d4e85233950c2e6c8ca30a1

PR #1511, opened 2023-10-11
Resolve type of `@field(..)` in the context of `usingnamespace`

nullptrdevs:
I admit ignorance on this one, was aiming for https://github.com/zigtools/zls/issues/1219

### DIFF
--- a/src/analysis.zig
+++ b/src/analysis.zig
@@ -529,6 +529,20 @@ fn resolveVarDeclAliasUncached(analyser: *Analyser, node_handle: NodeWithHandle,
     }
 }
 
+/// resolves `@field(lhs, field_name)`
+fn resolveFieldAccess(analyser: *Analyser, lhs: TypeWithHandle, field_name: []const u8) !?TypeWithHandle {
+    if (try analyser.resolveTaggedUnionFieldType(lhs, field_name)) |tag_type| return tag_type;
+
+    // If we are accessing a pointer type, remove one pointerness level :)
+    const left_type = (try analyser.resolveDerefType(lhs)) orelse lhs;
+
+    if (try analyser.resolvePropertyType(left_type, field_name)) |t| return t;
+
+    if (try left_type.lookupSymbol(analyser, field_name)) |child| return try child.resolveType(analyser);
+
+    return null;
+}
+
 fn findReturnStatementInternal(tree: Ast, fn_decl: Ast.full.FnProto, body: Ast.Node.Index, already_found: *bool) ?Ast.Node.Index {
     var result: ?Ast.Node.Index = null;
 
@@ -1305,18 +1319,7 @@ fn resolveTypeOfNodeUncached(analyser: *Analyser, node_handle: NodeWithHandle) e
 
             const symbol = tree.tokenSlice(datas[node].rhs);
 
-            if (try analyser.resolveTaggedUnionFieldType(lhs, symbol)) |tag_type|
-                return tag_type;
-
-            // If we are accessing a pointer type, remove one pointerness level :)
-            const left_type = (try analyser.resolveDerefType(lhs)) orelse lhs;
-
-            if (try analyser.resolvePropertyType(left_type, symbol)) |t|
-                return t;
-
-            if (try left_type.lookupSymbol(analyser, symbol)) |child| {
-                return try child.resolveType(analyser);
-            } else return null;
+            return try resolveFieldAccess(analyser, lhs, symbol);
         },
         .anyframe_type,
         .array_type,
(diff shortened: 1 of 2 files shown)


src/analysis.zig
```diff
@@ -1433,6 +1433,32 @@ fn resolveTypeOfNodeUncached(analyser: *Analyser, node_handle: NodeWithHandle) e
                 // reference to node '0' which is root
                 return TypeWithHandle.typeVal(.{ .node = 0, .handle = new_handle });
             }
+            if (std.mem.eql(u8, call_name, "@field")) {
+                if (params.len < 1) return null;
+                const quoted_param1_slice = handle.tree.tokenSlice(main_tokens[params[1]]);
```
Techatrix, line 1438: you probably wanna check if the first parameter is a string literal and maybe rename `quoted_param1_slice` to `field_name`

src/analysis.zig
```diff
@@ -1433,6 +1433,32 @@ fn resolveTypeOfNodeUncached(analyser: *Analyser, node_handle: NodeWithHandle) e
...
+                    switch (resolved_type.handle.tree.nodes.items(.tag)[member]) {
+                        .simple_var_decl => {},
+                        else => continue,
+                    }
+                    const name_token = resolved_type.handle.tree.nodes.items(.main_token)[member] + 1; // main is `var` or `const`
+                    if (!std.mem.eql(u8, offsets.tokenToSlice(resolved_type.handle.tree, name_token), quoted_param1_slice[1 .. quoted_param1_slice.len - 1])) continue;
+                    const ndata = resolved_type.handle.tree.nodes.items(.data);
+                    if (ndata[member].rhs == 0) break;
+                    return TypeWithHandle.typeVal(.{ .node = ndata[member].rhs, .handle = resolved_type.handle });
+                }
```
Techatrix, line 1460:
```suggestion
                const field_decl = try resolved_type.lookupSymbol(analyser, quoted_param1_slice) orelse return null;
                return try field_decl.resolveType(analyser);
```

tests/lsp_features/completion.zig
```diff
@@ -1253,6 +1253,45 @@ test "completion - usingnamespace" {
...
+        \\test {
+        \\    _ = chip.<cursor>;
+        \\}
+    , &.{
+        .{ .label = "inner", .kind = .Constant, .detail = "const chip_mod = struct" },
+        .{ .label = "types", .kind = .Constant, .detail = "const types = struct" },
+        .{ .label = "peripherals", .kind = .Constant, .detail = "const peripherals = struct" },
+        .{ .label = "chip1fn1", .kind = .Function, .detail = "fn chip1fn1() void" },
+        .{ .label = "chip1fn2", .kind = .Function, .detail = "fn chip1fn2(_: u32) void" },
+    });
```
Techatrix, line 1294: could this test case be made slightly smaller, maybe also add a test just for `@field`.

Techatrix (changes requested):

Techatrix (changes requested):
I just realized that its probably better to exactly match the code that resolves an ordinary field access [here](https://github.com/zigtools/zls/blob/61fec01a2006c5d509dee11c6f0d32a6dfbbf44e/src/analysis.zig#L1298-L1320).
```zig
/// resolves `@field(lhs, field_name)`
fn resolveFieldAccess(analyser: *Analyser, lhs: TypeWithHandle, field_name: []const u8) !?DeclWithHandle {}
```

nullptrdevs:
> I just realized that its probably better to exactly match the code that resolves an ordinary field access [here](https://github.com/zigtools/zls/blob/61fec01a2006c5d509dee11c6f0d32a6dfbbf44e/src/analysis.zig#L1298-L1320).
> 
> ```zig
> /// resolves `@field(lhs, field_name)`
> fn resolveFieldAccess(analyser: *Analyser, lhs: TypeWithHandle, field_name: []const u8) !?DeclWithHandle {}
> ```

I'm unable to find such fn in the whole codebase.

The other tidbits:
```zig
                var lhs = (try analyser.resolveTypeOfNodeInternal(.{
                    .node = params[0],
                    .handle = handle,
                })) orelse return null;

                if (try analyser.resolveTaggedUnionFieldType(lhs, field_name)) |tag_type| return tag_type;

                // If we are accessing a pointer type, remove one pointerness level :)
                const left_type = (try analyser.resolveDerefType(lhs)) orelse lhs;

                if (try analyser.resolvePropertyType(left_type, field_name)) |t| return t;

                if (try left_type.lookupSymbol(analyser, field_name)) |child| return try child.resolveType(analyser);

                return null;
```

Techatrix:
> I'm unable to find such fn in the whole codebase.

I meant to create a new function with the given interface, I should have made that more clear.

> the return type threw me off as it should be !?TypeWithHandle.

The reason why I picked `!?DeclWithHandle` in favor of `!?TypeWithHandle` is because you can easily convert a `DeclWithHandle` into a `TypeWithHandle` with the `resolveType` function and some (future) callsites need the Decl instead of the resolved type which would make the function more reusable.

> The other tidbits:

I guess you are suggesting to just duplicate the code, but I very much avoid this as you may modify one place and forget to edit the other. We do want to ensure that the behavior of `lhs.field_name` and `@field(lhs, field_name)` will always match.

nullptrdevs:
I simply can't see my way through to convert the results of
`analyser.resolveTaggedUnionFieldType(lhs, field_name)` and `analyser.resolvePropertyType(left_type, field_name)`
to `DeclWithHandle`.

Techatrix:
I see, I haven't thought of that, my bad...
`!?TypeWithHandle` should be fine then. I will see how I can work around my issue afterwards.

src/analysis.zig
```diff
@@ -1433,6 +1436,21 @@ fn resolveTypeOfNodeUncached(analyser: *Analyser, node_handle: NodeWithHandle) e
                 // reference to node '0' which is root
                 return TypeWithHandle.typeVal(.{ .node = 0, .handle = new_handle });
             }
+            if (std.mem.eql(u8, call_name, "@field")) {
+                if (params.len < 1) return null;
```
Techatrix, line 1440:
```suggestion
                if (params.len < 2) return null;
```

Techatrix (changes requested):

Techatrix (approved):

PR #1512, opened 2023-10-11
Inlay Hints for Control Flow Capture Values

Koranir:
Adds inlay hints hints for the capture values of the control flow statements (`if`, `while`, `for`, `switch`).
image
image
image

Fixed multiple capture values not working
image

### DIFF
--- a/src/features/inlay_hints.zig
+++ b/src/features/inlay_hints.zig
@@ -284,6 +284,11 @@ fn appendTypeHintString(builder: *Builder, type_token_index: u32, hint: []const 
     });
 }
 
+fn inferAppendTypeStr(builder: *Builder, token: Ast.TokenIndex) !void {
+    const type_str = try typeStrOfToken(builder, token) orelse return;
+    try appendTypeHintString(builder, token, type_str);
+}
+
 /// Takes a variable declaration AST node. If the type is inferred, attempt to infer it and display it as a hint.
 fn writeVariableDeclHint(builder: *Builder, decl_node: Ast.Node.Index) !void {
     const tracy_zone = tracy.trace(@src());
@@ -300,14 +305,9 @@ fn writeIfCaptureHint(builder: *Builder, if_node: Ast.Node.Index) !void {
     defer tracy_zone.end();
 
     const hint = builder.handle.tree.fullIf(if_node) orelse return;
-    if (hint.error_token == null and hint.payload_token == null) return;
 
-    if (hint.payload_token) |token| br: {
-        try appendTypeHintString(builder, token, try typeStrOfToken(builder, token) orelse break :br);
-    }
-    if (hint.error_token) |token| {
-        try appendTypeHintString(builder, token, try typeStrOfToken(builder, token) orelse return);
-    }
+    if (hint.payload_token) |token| try inferAppendTypeStr(builder, token);
+    if (hint.error_token) |token| try inferAppendTypeStr(builder, token);
 }
 
 fn writeForCaptureHint(builder: *Builder, for_node: Ast.Node.Index) !void {
@@ -315,7 +315,23 @@ fn writeForCaptureHint(builder: *Builder, for_node: Ast.Node.Index) !void {
     defer tracy_zone.end();
 
     const hint = builder.handle.tree.fullFor(for_node) orelse return;
-    try appendTypeHintString(builder, hint.payload_token, try typeStrOfToken(builder, hint.payload_token) orelse return);
+    const token_tags = builder.handle.tree.tokens.items(.tag);
+    var token_index: u32 = hint.payload_token;
+    var prepend: []const u8 = "";
+    while (token_tags[token_index] != .pipe) : (token_index += 1) {
+        if (token_tags[token_index] == .comma) continue;
+        if (token_tags[token_index] == .asterisk) {
+            prepend = "*";
+            continue;
+        }
+        const type_str = try typeStrOfToken(builder, token_index) orelse continue;
+        try appendTypeHintString(
+            builder,
+            token_index,
+            try std.fmt.allocPrint(builder.arena, "{s}{s}", .{ prepend, type_str }),
+        );
+        prepend = "";
+    }
 }
 
 fn writeWhileCaptureHint(builder: *Builder, while_node: Ast.Node.Index) !void {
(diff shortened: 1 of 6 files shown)


nullptrdevs:
No guarantees that this is the best way, or even the right way
```diff
diff --git a/src/features/inlay_hints.zig b/src/features/inlay_hints.zig
index c3f9151..2431a34 100644
--- a/src/features/inlay_hints.zig
+++ b/src/features/inlay_hints.zig
@@ -315,7 +315,13 @@ fn writeForCaptureHint(builder: *Builder, for_node: Ast.Node.Index) !void {
     defer tracy_zone.end();
 
     const hint = builder.handle.tree.fullFor(for_node) orelse return;
-    try appendTypeHintString(builder, hint.payload_token, try typeStrOfToken(builder, hint.payload_token) orelse return);
+    const token_tags = builder.handle.tree.tokens.items(.tag);
+    var token_index: u32 = hint.payload_token;
+    while (token_tags[token_index] != .pipe) : (token_index += 1) {
+        if (token_tags[token_index] == .comma) continue;
+        if (token_tags[token_index] == .asterisk) // Prepend `*` to the inlay hint
+        try appendTypeHintString(builder, token_index, try typeStrOfToken(builder, token_index) orelse return);
+    }
 }
 
 fn writeWhileCaptureHint(builder: *Builder, while_node: Ast.Node.Index) !void {
 ```

**Comment by Techatrix** [comment]
I would suggest the copy the code from analysis.zig [here](https://github.com/zigtools/zls/blob/61fec01a2006c5d509dee11c6f0d32a6dfbbf44e/src/analysis.zig#L4365-L4376) because it has already been tested, fuzzed and matches to code used by [AstGen](https://github.com/zigtools/zls/blob/61fec01a2006c5d509dee11c6f0d32a6dfbbf44e/src/stage2/AstGen.zig#L6511-L6521).

File: src/features/inlay_hints.zig [hunk 1]
```diff
@@ -254,22 +268,89 @@ fn writeVariableDeclHint(builder: *Builder, decl_node: Ast.Node.Index) !void {
...
+            continue;
+        }
+        const type_str = try typeStrOfToken(builder, token_index) orelse continue;
+        try appendTypeHintString(
+            builder,
+            token_index,
+            try std.fmt.allocPrint(builder.arena, "{s}{s}", .{ prepend, type_str }),
+        );
+        prepend = "";
+    }
```
nullptrdevs, line 334:
```suggestion
    const tree = builder.handle.tree;
    const hint = tree.fullFor(for_node) orelse return;
    const token_tags = tree.tokens.items(.tag);

    var capture_token = hint.payload_token;
    for (hint.ast.inputs) |_| {
        if (capture_token + 1 >= tree.tokens.len) break;
        const capture_is_ref = token_tags[capture_token] == .asterisk;
        const name_token = capture_token + @intFromBool(capture_is_ref);
        const type_str = try typeStrOfToken(builder, name_token) orelse continue;
        const prepend = if (capture_is_ref) "*" else "";
        try appendTypeHintString(
            builder,
            name_token,
            try std.fmt.allocPrint(builder.arena, "{s}{s}", .{ prepend, type_str }),
        );
        capture_token = name_token + 2;
    }
```

nullptrdevs (changes requested):
Techatrix is correct

nullptrdevs (approved):
Other than the type needing to be future-proofed this looks ready.

Some subjective remarks:
- I wish zig itself wasn't using `node` <-> `node_index` interchangeably
- Ditto `token` <-> `token_index`
- `hint` isn't a good fit for the name of the result of fns that return (full)X node
- These are so short, they could've been part of the `switch`

src/features/inlay_hints.zig
```diff
@@ -353,7 +401,46 @@ fn writeNodeInlayHint(
         .aligned_var_decl,
         => {
             if (!builder.config.inlay_hints_show_variable_declaration) return;
-            try writeVariableDeclHint(builder, node);
+            const var_decl = builder.handle.tree.fullVarDecl(node) orelse return;
```
Techatrix, line 404:
```suggestion
            const var_decl = builder.handle.tree.fullVarDecl(node).?;
```
The same goes for the cases below here.

src/features/inlay_hints.zig
```diff
@@ -254,22 +268,56 @@ fn writeVariableDeclHint(builder: *Builder, decl_node: Ast.Node.Index) !void {
...
+    const type_str = try typeStrOfToken(builder, token) orelse return;
+    try appendTypeHintString(builder, token, type_str);
+}
+
+fn writeForCaptureHint(builder: *Builder, for_node: Ast.Node.Index) !void {
+    const tracy_zone = tracy.trace(@src());
+    defer tracy_zone.end();
+
+    const tree = builder.handle.tree;
+    const full_for = tree.fullFor(for_node) orelse return;
```
Techatrix, line 302:
```suggestion
    const full_for = tree.fullFor(for_node).?;
```

src/features/inlay_hints.zig
```diff
@@ -254,22 +268,56 @@ fn writeVariableDeclHint(builder: *Builder, decl_node: Ast.Node.Index) !void {
         &type_str,
         &reference_collector,
     );
-    if (type_str.len == 0) return;
+    if (type_str.len == 0) return null;
 
-    // TODO: Remove once long type hints can be reduced, i.e. `struct { .. }`
```
Techatrix, line 259: Was it intentional to remove this TODO comment?

Techatrix (changes requested):
There is a missing test case for switch case captures and a test to see if `for("",0..) |_, here<usize>|` works.

I am also wondering if there is any need to add a new config option. Shouldn't `inlay_hints_show_variable_declaration` and `inlay_hints_show_capture_variables` be merged into a single option since they both are type hints or is there interest in having only one of them enabled? Keep in mind that something like #1531 may even add another option to the mix.

src/features/inlay_hints.zig
nullptrdevs: Wouldn't this be as simple as,
if hint startsWith "struct " => hint = "struct {...}"
if hint startsWith "fn " => hint = "fn"
enum, union(enum), ..

Techatrix: It may be reasonable to show small containers like `struct { foo: u32 }` or `struct { []const u8 }` but except that, I think your right.

Koranir: I added the comment back in, it would be a topic for a separate PR though.

Koranir:
When opting for inlay hints for variables,  I suppose you'd want it on all the variables.

Techatrix (approved):

PR #1533, opened 2023-10-16
completions: Unwrap the field's type if it's an optional

nullptrdevs:
Handles `x{.a_field_with_an_optional_type =`


nullptrdevs:
> [resolveUnwrapOptionalType](https://github.com/zigtools/zls/blob/1a282b30cccb6dd52438fa1ef6e6b887069e109f/src/analysis.zig#L618)

I did, just to check if it works, but it's not marked `pub`.

Techatrix:
you can make it public

Techatrix (approved):

PR #1536, opened 2023-10-17
Explore modifying Zig's parser to fit the needs of a language server

nullptrdevs:
## Feeling out how disruptive this would be

Ignoring missing semicolons (https://github.com/ziglang/zig/blob/eb5276c94eaab238551fdae9a2e77b0133e31cfb/lib/std/zig/Parse.zig#L876) solves
```zig
const E = enum {
    foo,
    bar,
    baz,
    fn foo(e: E) void {
        switch (e) {.<cursor>};
    }
};

pub const bar = struct {
    pub const baz = struct {
        pub const Foo = struct {
            a: u32 = 0,
        };

        pub fn qux() u32 {
            return Foo{
                .<cursor>
            };
        }
    };
};

pub var state: S = undefined;
pub const S = struct { foo: u32 };

pub fn main() void {
    state.<cursor>
    {
        _ = state.foo;
    }
    state.foo;
}
```

Haven't observed any side effects and  the number and type of errors is the same.

<details>
<summary>Somewhat of a changelog</summary>

```
> Added Ast.zig, Parse.zig, parser_test.zig, render.zig and tokenizer.zig

DocumentStore.zig
        const Parser = @import("stage2/Ast.zig");

        var zls_ast = try Parser.parse(self.allocator, text, .zig);
        errdefer zls_ast.deinit(self.allocator);
        var tree = Ast{
            .source = zls_ast.source,
            .tokens = zls_ast.tokens,
            .nodes = zls_ast.nodes,
            .extra_data = zls_ast.extra_data,
            .errors = zls_ast.errors,
        };
        
stage2/Ast.zig
        tokens: std.zig.Ast.TokenList.Slice,
        nodes: std.zig.Ast.NodeList.Slice,
        errors: []const std.zig.Ast.Error,
        
> All tests pass (as expected)
> Changed stage2/Parse.zig:876 to     try p.expectSemicolon(.expected_semi_after_decl, true);
> Tests that failed (I think the parser has a point here)
    try testCompletion(
        \\const S = struct { alpha: u32 };
        \\fn foo(comptime T: type) T {}
        \\const s = foo(S);
        \\const foo = s.<cursor>
    , &.{
        .{ .label = "alpha", .kind = .Field, .detail = "alpha: u32" },
    });
    try testCompletion(
        \\const S = struct { alpha: u32 };
        \\fn foo(any: anytype, comptime T: type) T {}
        \\const s = foo(null, S);
        \\const foo = s.<cursor>
    , &.{
        .{ .label = "alpha", .kind = .Field, .detail = "alpha: u32" },
    });
    
> Fixed up the tests
> These cases now work (num errors is the same which is ok and better than I expected)
pub const bar = struct {
    pub const baz = struct {
        pub const Foo = struct {
            a: u32 = 0,
        };

        pub fn qux() u32 {
            return Foo{
                .<cursor>
            };
        }
    };
};

pub var state: S = undefined;
pub const S = struct { foo: u32 };

pub fn main() void {
    state.<cursor>
    {
        _ = state.foo;
    }
    state.foo;
}

> Can this work?
        fn alias() void {
            var s = Alias{.<cursor>}; // Def node is after the current node
        }
        pub const Outer = struct {
            pub const Inner = struct {
                isf1: bool = true,
                isf2: bool = false,
            };
        };
        const Alias0 = Outer.Inner;
        const Alias = Alias0;

> Changed stage2/Parse.zig:2966 to
            else => return p.addNode(.{
                .tag = .unreachable_literal,
                .main_token = p.tok_i,
                .data = .{
                    .lhs = undefined,
                    .rhs = undefined,
                },
            }), //return null_node,
> now it does, but Sema prob won't like this (custom node_tags?)

> Some tests failed, the change revealed that those statements should be within a `test{..}` block.. another point for a working parser

> dot compl for the struct fields above don't work
> tweaked stage2/Parse.zig:3315
    {
        try p.warn(.expected_initializer); // return p.fail(.expected_initializer);
        p.tok_i += 1;
    } else p.tok_i += 3;
    
> result
            else => return p.addNode(.{.<nope>
                .tag = .unreachable_literal,.<works>
                .main_token = p.tok_i,.<works>
                .data = .{.<nope>
                    .lhs = undefined,.<works>
                    .rhs = undefined,.<works>
                    .<nope>
                },
                .<nope>
            }), //return null_node,
> works only if surrounded by existing fields...

> zig test src/stage2/Parse.zig 
All 340 tests passed.

> checked
const World = struct {
	name: []const u8,
	placeHolderOtherStructFields: [33693888]u8 = undefined,

	pub fn init(self: World, name: []const u8) void {
		self = .{};
	}
};
> works! but not if `self: *World`(check completions logic, not a parser issue)
```

</details>


src/stage2/Parse.zig
```diff
@@ -2963,7 +2963,14 @@ fn parsePrimaryTypeExpr(p: *Parse) !Node.Index {
...
             },
-            else => return null_node,
+            else => return p.addNode(.{
+                .tag = .unreachable_literal,
+                .main_token = p.tok_i,
+                .data = .{
+                    .lhs = undefined,
+                    .rhs = undefined,
+                },
+            }),
```
Techatrix, line 2973: What is the motivation for this change? I undid this change and ran tests and tried you example code from the PR description and everything was still working.

Techatrix:
Would you consider to try and upstream these changes? I think these changes quite different from a PR like https://github.com/ziglang/zig/pull/14391 where a feature that only gets used by third party projects is being added.

src/stage2/Parse.zig
nullptrdevs: I had placebo'd myself, but had already reverted it as part of the "Take out the guesswork and use a debugger" changes that I just pushed.

nullptrdevs:
```zig
// Test for completions after `.{` and every `,`
pub fn gamma(p: Parser) void {
    return p.addNode(.{
        .a = .some,
        .b = 2,
        .data = .{
            .lhs = undefined,
            .rhs = undefined,
        },
    });
}
const Parser = struct {
    const Node = struct {
        tag: u32,
        main_token: u32,
        data: struct {
            lhs: u32,
            rhs: u32,
        },
    };

    pub fn addNode(_: Node) void {}
};


fn foo(e: Enum) void {
    switch (e) {}
}

const Enum = enum {
    foo,
    bar,
    baz,
};
```

nullptrdevs:
Upkeep's been quite light, the only change being https://github.com/ziglang/zig/commit/f258a391daa31b3ba2c37d879db96fadc0c058f3

Techatrix:
Isn't only the `Parse.zig` and the `parse` function from `Ast.zig` needed? That could all be moved into `Parse.zig` to have only a single file.

nullptrdevs:
> Isn't only the `Parse.zig` and the `parse` function from `Ast.zig` needed? That could all be moved into `Parse.zig` to have only a single file.

The intention is to keep it (close to) a carbon copy as it makes it easy to update by just overwriting a/the file(s) and looking at the overlapping sections.

llogick:
I'm not sure why this PR is still open, but I'm growing fond of it and updated it for it's upcoming anniversary 😋

Cheers

llogick:
RIP

PR #1541, opened 2023-10-22
test signature help + fixes

### LINKED ISSUE #1438: Show parameters when calling functions by pointer
When calling functions by pointer, either as a `fn` or `*const fn`, ZLS won't show the list of parameters while typing or manually toggling.
```zig
const S = struct {
    foo: fn (a: i32, b: i32) i32,
    bar: *const fn (a: i32, b: i32) i32,
};

fn baz(a: i32, b: i32) i32 {
    return a + b;
}

pub fn main() void {
    const s = S{ .foo = baz, .bar = &baz };
    _ = s.foo(1, 2); // won't show parameters
    _ = s.bar(3, 4); // won't show parameters
    _ = s.bar.*(5, 6); // dereferencing bar will show parameters
}
```
Dereferencing makes it work, but this is annoying to do just to see parameters.
Screenshot 2023-09-09 at 4 48 54 PM

Ideally this would work for all cases.

Techatrix:
Adds test coverage for the signature help request and fixes when calling a function pointer (fixes #1438)

### DIFF
--- a/src/analysis.zig
+++ b/src/analysis.zig
@@ -708,6 +708,12 @@ fn resolveTaggedUnionFieldType(analyser: *Analyser, type_handle: TypeWithHandle,
     return null;
 }
 
+pub fn resolveFuncProtoOfCallable(analyser: *Analyser, type_handle: TypeWithHandle) error{OutOfMemory}!?TypeWithHandle {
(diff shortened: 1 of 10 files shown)


nullptrdevs (approved):
Minor conflict

PR #1548, opened 2023-10-24
Fix build runner module crawler

tomc1998:
The build runner was returning when it found a default module, which is a bug; it was missing some modules in projects with more complicated module graphs.


tomc1998:
Oh, looks like special/build_runner.zig was renamed to `build_runner/0.11.0.zig`? possibly the same fix needs to be made to `build_runner.master.zig` also..?

leecannon:
This was already fixed for master build runner in #1393

leecannon (approved):

PR #1557, opened 2023-10-28
Some fixes

nullptrdevs:
- fix cimports
- fix wrong insertText


Techatrix (commented):
what about `handle.import_uris = try self.collectImportUris(handle.*);`?

Techatrix (approved):

Issue #1218, opened 2023-05-31
No hover/go-to on zig-gamedev/zgui, culprit is `usingnamespace`


foxnne:
### Zig Version

0.11.0-dev.3323+706bdf651

### Zig Language Server Version

5d53f0104fe6742a28592094f9f7de5060a6f07c

### Steps to Reproduce

- Clone zig-gamedev
- Open sample using zgui
- Autocomplete should work, but not hover or go-to definition.

Note: zgui uses `usingnamespace` to bring all of the elements into `main.zig`. Removing this and putting everything into `gui.zig` and correcting the `build.zig` for this fixes the issue. 

### Expected Behavior

Hover and go-to to work just like autocomplete

### Actual Behavior

Autocomplete works great but not hover/go-to.

foxnne:
zig-gamedev just merged PR removing the usingnamespace usage, but it should be easy to repro, just using module and usingnamespace should cause the same issue.

Techatrix:
fixed by #1566.

nullptrdevs:
> fixed by #1566.

Had left this issue open because looking at the PR diff the issue is/was? with
`pub usingnamespace @import("gui.zig");`.

PR #1566, opened 2023-10-31
Support more than `.string_literal` as second param to `@field`

nullptrdevs:


src/analysis.zig
```diff
@@ -1446,17 +1446,56 @@ fn resolveTypeOfNodeUncached(analyser: *Analyser, node_handle: NodeWithHandle) e
...
                 })) orelse return null;
 
-                return try resolveFieldAccess(analyser, lhs, field_name[1 .. field_name.len - 1]);
+                return try resolveFieldAccess(analyser, lhs, field_name[1 .. field_name.len - 1])
+                // Try `@"field_name"`
+                orelse try resolveFieldAccess(
+                    analyser,
+                    lhs,
+                    try std.fmt.allocPrint(analyser.arena.allocator(), "@{s}", .{field_name}),
+                );
```
Techatrix, line 1498: This is a workaround for an issue with DocumentScope. It adds a declaration with the name `@"chip1"` instead of `chip1`. This should be fixed in DocumentScope and not here.
And also, this is a very confusing place to add a comment.

src/analysis.zig
```diff
@@ -1446,17 +1446,56 @@ fn resolveTypeOfNodeUncached(analyser: *Analyser, node_handle: NodeWithHandle) e
...
+                                    const elem_ty_tree = pointer.elem_ty.handle.tree;
+                                    if ((str_node > elem_ty_tree.nodes.len - 1) or (elem_ty_tree.nodes.items(.tag)[str_node] != .string_literal)) return null;
+                                    break :fld_name_blk offsets.nodeToSlice(elem_ty_tree, str_node);
+                                },
+                                else => return null,
+                            }
+                        },
+                        else => return null,
+                    }
+                };
```
Techatrix, line 1482: I would suggest to use `resolveVarDeclAlias` instead of  `resolveTypeOfNodeInternal` because you don't search for a type but rather the initialization expression hidden behind some indirection/aliasing that you can undo with `resolveVarDeclAlias`. This also makes in unnecessary to set the node when resolving primitve types.

```suggestion
                var field_name_node: NodeWithHandle = .{ .node = params[1], .handle = handle };

                if (try analyser.resolveVarDeclAlias(field_name_node)) |decl_with_handle| {
                    if (decl_with_handle.decl == .ast_node) {
                        field_name_node = .{
                            .node = decl_with_handle.decl.ast_node,
                            .handle = decl_with_handle.handle,
                        };
                    }
                }
                const string_literal_node = switch (field_name_node.handle.tree.nodes.items(.tag)[field_name_node.node]) {
                    .string_literal => field_name_node.node,
                    .global_var_decl,
                    .local_var_decl,
                    .aligned_var_decl,
                    .simple_var_decl,
                    => blk: {
                        const init_node = field_name_node.handle.tree.fullVarDecl(field_name_node.node).?.ast.init_node;
                        if (field_name_node.handle.tree.nodes.items(.tag)[init_node] != .string_literal) return null;
                        break :blk init_node;
                    },
                    else => return null,
                };

                const field_name_token = field_name_node.handle.tree.nodes.items(.main_token)[string_literal_node];
                const field_name = offsets.tokenToSlice(field_name_node.handle.tree, field_name_token);

```

tests/lsp_features/completion.zig
```diff
@@ -1393,6 +1393,56 @@ test "builtin fn `field`" {
...
+        \\          pub const chip_name: []const u8 = "chip1";
+        \\      };
+        \\    const chip = @field(chip_mod.devices, chips.chip_name);
+        \\    chip.<cursor>
+        \\}
+    , &.{
+        .{ .label = "peripherals", .kind = .Constant, .detail = "const peripherals = struct" },
+        .{ .label = "chip1fn1", .kind = .Function, .detail = "fn chip1fn1() void" },
+        .{ .label = "chip1fn2", .kind = .Function, .detail = "fn chip1fn2(_: u32) void" },
+    });
```
Techatrix, line 1445: I think these tests are unnecesserily long. How about doing this:

Note that in the third test, the name of `Inner` is not `@"Inner"`. See the other review comment for the reasoning.
```zig
    try testCompletion(
        \\const Outer = struct {
        \\    const Inner = struct {
        \\        fn foo() void {}
        \\    };
        \\};
        \\test {
        \\    const chip = @field(Outer, "Inner");
        \\    chip.<cursor>
        \\}
    , &.{
        .{ .label = "foo", .kind = .Function, .detail = "fn foo() void" },
    });
    try testCompletion(
        \\const Outer = struct {
        \\    const Inner = struct {
        \\        fn foo() void {}
        \\    };
        \\};
        \\test {
        \\    const chip_name = "Inner";
        \\    const chip = @field(Outer, chip_name);
        \\    chip.<cursor>
        \\}
    , &.{
        .{ .label = "foo", .kind = .Function, .detail = "fn foo() void" },
    });
    try testCompletion(
        \\const Outer = struct {
        \\    const Inner = struct {
        \\        fn foo() void {}
        \\    };
        \\};
        \\test {
        \\    const chips = struct {
        \\        pub const chip_name: []const u8 = "Inner";
        \\    };
        \\    const chip = @field(Outer, chips.chip_name);
        \\    chip.<cursor>
        \\}
    , &.{
        .{ .label = "foo", .kind = .Function, .detail = "fn foo() void" },
    });
```

Techatrix (changes requested):
Does this perhaps address #1218 or #1219?

nullptrdevs:
> Does this perhaps address #1218 or #1219?

Yes, basically I'm cherry picking, cleaning up and PRing commits from this branch https://github.com/zigtools/zls/tree/nullptrdevs/microzig .

src/analysis.zig
nullptrdevs: Yes, but the point was to make the `orelse` stand out :)

Do let me know if there's a better strat to trim `@"` from `name`.

src/analysis.zig
nullptrdevs: Done

tests/lsp_features/completion.zig
nullptrdevs: There should be at least one `@"chip-name"` test to cover eg `pub const @"ESP32-C3" = struct {...};`.

src/DocumentScope.zig
```diff
@@ -726,7 +726,10 @@ noinline fn walkContainerDecl(
                 const name_token = tree.fullVarDecl(decl).?.ast.mut_token + 1;
                 if (name_token >= tree.tokens.len) continue;
 
-                const name = offsets.tokenToSlice(tree, name_token);
+                var name = offsets.tokenToSlice(tree, name_token);
+                if (name.len > 1 and name[0] == '@' and name[1] == '"' and name[name.len - 1] == '"') {
+                    name = name[2 .. name.len - 1];
+                }
```
Techatrix, line 732: There are many more places where this check is necessary. See #1574

tests/lsp_features/completion.zig
Techatrix: You mean something like this?
```zig
    try testCompletion(
        \\const Outer = struct {
        \\    const Inner = struct {
        \\        fn foo() void {}
        \\    };
        \\};
        \\test {
        \\    const chips = struct {
        \\        pub const @"chip-name": []const u8 = "Inner";
        \\    };
        \\    const chip = @field(Outer, chips.@"chip-name");
        \\    chip.<cursor>
        \\}
    , &.{
        .{ .label = "foo", .kind = .Function, .detail = "fn foo() void" },
    });
```
This doesn't seem to be related to `@field` but rather resolving `chips.@"chip-name"` itself.  #1574 has test coverage for this. But more tests will never hurt so do as you wish.

Techatrix (changes requested):

tests/lsp_features/completion.zig
nullptrdevs: True, but those tests didn't exist at the time, would probably keep them as they are.

src/analysis.zig
```diff
@@ -1083,7 +1083,7 @@ fn resolveTypeOfNodeUncached(analyser: *Analyser, node_handle: NodeWithHandle) e
             if (resolvePrimitiveType(name)) |primitive| {
                 const is_type = analyser.ip.indexToKey(primitive).typeOf() == .type_type;
                 return TypeWithHandle{
-                    .type = .{ .data = .{ .ip_index = .{ .index = primitive } }, .is_type_val = is_type },
```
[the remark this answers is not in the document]
nullptrdevs: Will revert

src/DocumentScope.zig
nullptrdevs: Will rebase on top of it

Techatrix (approved):

PR #1570, opened 2023-11-01
Full build runners

nullptrdevs:
- Resolve mod imports in `build.zig`s
 - Show completions for mod imports in `build.zig`s

### DIFF
--- a/src/DocumentStore.zig
+++ b/src/DocumentStore.zig
@@ -460,7 +460,7 @@ pub fn getBuildFile(self: *DocumentStore, uri: Uri) ?BuildFile {
 
 /// invalidates any pointers into `DocumentStore.build_files`
 /// **Thread safe** takes an exclusive lock
-fn getOrLoadBuildFile(self: *DocumentStore, uri: Uri) ?std.StringArrayHashMapUnmanaged(BuildFile).Entry {
+pub fn getOrLoadBuildFile(self: *DocumentStore, uri: Uri) ?std.StringArrayHashMapUnmanaged(BuildFile).Entry {
     {
         self.lock.lockShared();
         defer self.lock.unlockShared();
@@ -1333,6 +1333,14 @@ pub fn uriFromImportStr(self: *DocumentStore, allocator: std.mem.Allocator, hand
             for (config.value.packages) |pkg| {
                 if (std.mem.eql(u8, import_str, pkg.name)) {
                     return try URI.fromPath(allocator, pkg.path);
+                }
+            }
+        } else if (isBuildFile(handle.uri)) {
+            const build_file_entry = self.getOrLoadBuildFile(handle.uri) orelse return null;
+            const config = build_file_entry.value_ptr.config orelse return null;
+            for (config.value.deps_build_roots) |dep_build_root| {
+                if (std.mem.eql(u8, import_str, dep_build_root.name)) {
+                    return try URI.fromPath(allocator, dep_build_root.path);
                 }
             }
         }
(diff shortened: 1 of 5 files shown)


src/DocumentStore.zig
```diff
@@ -1335,6 +1335,14 @@ pub fn uriFromImportStr(self: *DocumentStore, allocator: std.mem.Allocator, hand
                     return try URI.fromPath(allocator, pkg.path);
                 }
             }
+        } else if (isBuildFile(handle.uri)) {
+            const build_file_entry = self.getOrLoadBuildFile(handle.uri) orelse return null;
+            const config = build_file_entry.value_ptr.config orelse return null;
```
Techatrix, line 1340: Similar to the what I've written above, this code has a potential race condition since pointers to build files are not stable and may be invalidated immediately after `getOrLoadBuildFile` has returned unless you are holding a lock to the document store. I will need to do some refactoring to make it easier to write this in a safe way.

The code that I've written above has also a race condition because we are reading from `build_file.config` without holding a lock. This case is much easier to fix than your code (just re-aquire a lock after `getBuildFile`).

You can let me fix this as follow up.

src/features/completions.zig
```diff
@@ -888,6 +888,17 @@ fn completeFileSystemStringLiteral(
                     .detail = pkg.path,
                 }, {});
             }
+        } else if (std.mem.endsWith(u8, handle.uri, "/build.zig")) blk: {
```
Techatrix, line 891:
```suggestion
        } else if (DocumentStore.isBuildFile(handle.uri)) blk: {
```

Techatrix (commented):
What was the motivation to copy over the entire build runner code from zig? Doesn't seem to me that its actually being used for something.

nullptrdevs:
> What was the motivation to copy over the entire build runner code from zig? Doesn't seem to me that its actually being used for something.

It runs all the steps and, supposedly, does so in parallel ..also respects max_rss ? :)

Could probably be trimmed down though, time permitting and all

src/DocumentStore.zig
nullptrdevs: Ah, I should've probably used `getBuildFile`, as this matters only when it is loaded and being edited.

Techatrix:
I see, but this will run the entire install step. If it fails (maybe because your project has compile errors) then no build information is collected. And this step can take an arbitrarily large amount of time. What if it is a large project like Zig which takes relatively long to compile? This will also cause compilation artifacts to be installed just by opening a project with ZLS.

src/DocumentStore.zig
Techatrix: #1575 should solve this.

nullptrdevs:
> I see, but this will run the entire install step. If it fails (maybe because your project has compile errors) then no build information is collected. And this step can take an arbitrarily large amount of time. What if it is a large project like Zig which takes relatively long to compile? This will also cause compilation artifacts to be installed just by opening a project with ZLS.

Those are all very good points !

- Artifacts being installed - I had not considered that, will look into filtering those steps out.
- Errors - I did remove the summary and error reporting/exit conditions at the end of `fn runStepNames` and in a quick test, where I introduced a compile error, the compiler did report the error on stderr, but the brunner still collected the deps and wrote to stdout.
- Large amount of time - I see the errors of my way, will have to rethink it.

nullptrdevs:
@Techatrix as conservative as it can be: https://github.com/zigtools/zls/compare/master...nullptrdevs/make-genfile-step%2Bmicrozig

PR #1581, opened 2023-11-05
Minor build runners update

nullptrdevs:
0.11.0
- Fix `getPath() was called on a GeneratedFile that wasn't built yet.`
- Add missing `cache.hash.addBytes(builtin.zig_version_string);`

0.11.0, master:
- Resolve mod imports in `build.zig`s
- Show completions for mod imports in `build.zig`s

<details>
<summary>As already known, anything that depends on a generated path cannot be resolved.. unless:</summary>

```diff
diff --git a/src/build_runner/0.11.0.zig b/src/build_runner/0.11.0.zig
index 27ad44c..bf35faa 100644
--- a/src/build_runner/0.11.0.zig
+++ b/src/build_runner/0.11.0.zig
@@ -313,6 +313,18 @@ fn getPkgConfigIncludes(
     } else |err| return err;
 }
 
+fn reify(step: *Build.Step) anyerror!void {
+    var progress: std.Progress = .{};
+    const main_progress_node = progress.start("", 0);
+    defer main_progress_node.end();
+
+    for (step.dependencies.items) |unknown_step| {
+        try reify(unknown_step);
+    }
+
+    try step.make(main_progress_node);
+}
+
 // TODO: Having a copy of this is not very nice
 const copied_from_zig = struct {
     /// Copied from `std.Build.LazyPath.getPath2` and massaged a bit.
@@ -320,7 +332,19 @@ const copied_from_zig = struct {
         switch (path) {
             .path => |p| return builder.pathFromRoot(p),
             .cwd_relative => |p| return pathFromCwd(builder, p),
-            .generated => |gen| return builder.pathFromRoot(gen.path orelse return null),
+            .generated => |gen| {
+                if (gen.path) |gen_path|
+                    return builder.pathFromRoot(gen_path)
+                else {
+                    for (gen.step.owner.top_level_steps.values()) |tls| {
+                        reify(&tls.step) catch return null;
+                    }
+                    if (gen.path) |gen_path|
+                        return builder.pathFromRoot(gen_path)
+                    else
+                        return null;
+                }
+            },

```

</details>

<details>
<summary>And in the case of `microzig`, where `microzig` is a dep of an app, but imports `chip` which is a module of `app`</summary>

```diff
diff --git a/src/DocumentStore.zig b/src/DocumentStore.zig
index 5fc212a..50b03ea 100644
--- a/src/DocumentStore.zig
+++ b/src/DocumentStore.zig
@@ -1403,14 +1403,17 @@ pub fn uriFromImportStr(self: *DocumentStore, allocator: std.mem.Allocator, hand
         }
         return null;
     } else if (!std.mem.endsWith(u8, import_str, ".zig")) {
-        if (handle.associated_build_file) |build_file_uri| blk: {
-            const build_file = self.getBuildFile(build_file_uri).?;
-            const build_config = build_file.tryLockConfig() orelse break :blk;
-            defer build_file.unlockConfig();
-
-            for (build_config.packages) |pkg| {
-                if (std.mem.eql(u8, import_str, pkg.name)) {
-                    return try URI.fromPath(allocator, pkg.path);
+        // lock?
+        for (self.handles.values()) |h| {
+            if (h.associated_build_file) |build_file_uri| blk: {
+                const build_file = self.getBuildFile(build_file_uri).?;
+                const build_config = build_file.tryLockConfig() orelse break :blk;
+                defer build_file.unlockConfig();
+
+                for (build_config.packages) |pkg| {
+                    if (std.mem.eql(u8, import_str, pkg.name)) {
+                        return try URI.fromPath(allocator, pkg.path);
+                    }
                 }
             }
         } else if (isBuildFile(handle.uri)) blk: {

```

</details>

### DIFF
--- a/src/DocumentStore.zig
+++ b/src/DocumentStore.zig
@@ -875,7 +875,7 @@ pub fn loadBuildConfiguration(self: *DocumentStore, build_file_uri: Uri) !std.js
             .allocator = self.allocator,
             .argv = args,
             .cwd = std.fs.path.dirname(build_file_path).?,
-            .max_output_bytes = 1024 * 100,
+            .max_output_bytes = 1024 * 1024,
         });
     };
     defer self.allocator.free(zig_run_result.stdout);
@@ -1411,6 +1411,16 @@ pub fn uriFromImportStr(self: *DocumentStore, allocator: std.mem.Allocator, hand
             for (build_config.packages) |pkg| {
                 if (std.mem.eql(u8, import_str, pkg.name)) {
                     return try URI.fromPath(allocator, pkg.path);
+                }
+            }
+        } else if (isBuildFile(handle.uri)) blk: {
+            const build_file = self.getBuildFile(handle.uri).?;
+            const build_config = build_file.tryLockConfig() orelse break :blk;
+            defer build_file.unlockConfig();
+
+            for (build_config.deps_build_roots) |dep_build_root| {
+                if (std.mem.eql(u8, import_str, dep_build_root.name)) {
+                    return try URI.fromPath(allocator, dep_build_root.path);
                 }
             }
         }
--- a/src/build_runner/0.11.0.zig
+++ b/src/build_runner/0.11.0.zig
@@ -13,13 +13,15 @@ const Build = std.Build;
 ///! This is a modified build runner to extract information out of build.zig
 ///! Modified version of lib/build_runner.zig
 pub fn main() !void {
+    // Here we use an ArenaAllocator backed by a DirectAllocator because a build is a short-lived,
+    // one shot program. We don't need to waste time freeing memory and finding places to squish
+    // bytes into. So we free everything all at once at the very end.
     var arena = std.heap.ArenaAllocator.init(std.heap.page_allocator);
     defer arena.deinit();
 
     const allocator = arena.allocator();
 
     var args = try process.argsAlloc(allocator);
-    defer process.argsFree(allocator, args);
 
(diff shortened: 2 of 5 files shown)


Techatrix (approved):

Techatrix:
> As already known, anything that depends on a generated path cannot be resolved.. unless:

I think it would be good idea to run generated steps in certain cases, even though I am unsure why the code you showed is calling reify on every top level step of the build file. Isn't that even worse that running the install step, since it could also include other steps like tests?

nullptrdevs:
> even though I am unsure why the code you showed is calling reify on every top level step of the build file. Isn't that even worse that running the install step, since it could also include other steps like tests?

Hey, that's why I run these by you :)

The diffs were mostly showcase/proof of concept. I did try making just that step but it wasn't enough so I went top level.
The second example was also a broadstroke "solution", but it can be done in a stricter way (I think).
Most microzig apps follow this pattern
```zig
const microzig = @import("microzig");
const peripherals = microzig.chip.peripherals;
```
This, I believe, goes through `analysis.getFieldAccessType`, so if we store original (app's) handle`const orig_handle = handle;` at the beginning of it,
and then (a few iterations later) pass it to `analyser.store.uriFromImportStr` as a fallback handle or do a second call to it with `orig_handle` it should work.
The only "downside" I can think of is issues being filed because go to def won't work within `microzig.zig` at `const inner = @import("chip");`.

> I think it would be good idea to run generated steps in certain cases

I'll look into refining it

Issue #1020, opened 2023-02-22
Rethink how ZLS maintains backwards compatibility


Techatrix:
This has partially be brought up in #1000 and #1019

With the numerous breaking changes to the Zig language and its build system, it became harder to maintain compatibility with older Zig versions. This issue is only going to get worse as more versions are released and Zig's development accelerates.

I think its time to discuss ZLS's plan on how its going maintain compatibility with Zig.
Its harder to make a decision without knowing how many people are going to be affected by certain changes.
How many people use a pre 0.10.x release of Zig with ZLS master?
How many people use ZLS master compared a tagged release?
What version(s) do Zig projects tend to require?
How quickly do projects transition to new versions of Zig?

Despite that, here are some possibilities i could think of:
### Strict
ZLS master only maintains compatibility with Zig master.
A tagged release of ZLS is only compatible with the same tagged release of Zig.

### Relaxed
ZLS master can only be compiled with Zig master and is compatible with Zig master and the the latest tagged release of Zig.
A tagged release of ZLS must be compiled with and is only compatible with the same tagged release of Zig.

### Split branches
We take the Strict Proposal but also maintain separate branch of ZLS that backports changes while maintaining compatibility with the latest tagged release of Zig.

leecannon:
This is just me writing my thoughts down, anything im missing or you disaggre with let me know and I will update it.

# Strict
Pros:
- Simpler code
- Less chance for less bugs
- Reduced maintenance burden

Cons:
- People on tagged releases of Zig have to wait for a Zig release cycle to get new ZLS features.
- People how use different version of Zig for different project now also need to switch between ZLS version (can the extension be made smart enough to make this not painful?) 

# Relaxed
Pros:
- Single binary supports wide range of Zig versions (even if your stuck on a specific commit of master instead of a release, for 2 months last year I could not update Zig past a specific commit without hitting a blocking issue)

Cons:
- Code is complicated by being forced to deal with the differences between Zig versions
- More chance of bugs and regressions
- More work
- Upon tagging a new release the code would need to be cleaned up to remove the compatibility logic for the previous release (as we only want to support master and the latest release)

# Split Branches
Pros:
- The main branch is nice and clean
- The tagged release support branch could be cleaner than the same code under the relaxed model as in the relaxed model both master and release code is intertwined

Cons:
- More work
[...]

PR #1585, opened 2023-11-05
flake: add per data version derivations

polykernel:
This commit refactors the flake `packages` output to generate derivations serving different data versions. This allows downstream flake consumer to use zls with a versioned release of zig without effectively duplicating the zls derivation as per the status quo. Currently, zls packages serving the latest three tagged releases and the master branch of zig are produced.

I wasn't sure what the support range for the `data_version` build option is, so I naively opted for the latest three versioned releases + master. Please let me know if this should be changed.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -10,9 +10,6 @@
       gitignore.inputs.nixpkgs.follows = "nixpkgs";
 
       flake-utils.url = "github:numtide/flake-utils";
-
-      langref.url = "https://raw.githubusercontent.com/ziglang/zig/63bd2bff12992aef0ce23ae4b344e9cb5d65f05d/doc/langref.html.in";
-      langref.flake = false;
     };
 
   outputs = inputs:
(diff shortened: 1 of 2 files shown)


Techatrix:
This change would suggest that ZLS master is supporting tagged releases of Zig which is not planned as discussed in #1020.
If you are using a tagged release of Zig, you should also be using a tagged release of ZLS. This means that the should not be a data_version option to avoid this kind of confusion. Once further breaking changes happen to Zig, it also makes sense to remove ZLS's build runner for tagged releases of Zig (0.10.0 & 0.11.0) as discussed [here](https://github.com/zigtools/zls/issues/1020#issuecomment-1772738942).

I do have to admit that this is not the most user friendly decision, but Zig before 1.0 is improving/breaking rapidly and it is not always easy to keep up with and ensuring backwards compatibility would increase the required effort even more. Everyone is free to join the discussion in #1020.

NOTE: it is kind of possible to use ZLS master with Zig 0.11.0 right now, but this is not a guarantee and you should not depend on this being the case once further breaking changes to Zig happen.

polykernel:
> If you are using a tagged release of Zig, you should also be using a tagged release of ZLS. This means that the should not be a data_version option to avoid this kind of confusion. Once further breaking changes happen to Zig, it also makes sense to remove ZLS's build runner for tagged releases of Zig (0.10.0 & 0.11.0) as discussed https://github.com/zigtools/zls/issues/1020#issuecomment-1772738942.

Ah I see, I did noticed there are tagged releases of ZLS but seeing `data_version` build option and multiple build runners, I thought there wasn't a strict requirement for matching ZLS and Zig versions. Thanks for the clarification.

> I do have to admit that this is not the most user friendly decision, but Zig before 1.0 is improving/breaking rapidly and it is not always easy to keep up with and ensuring backwards compatibility would increase the required effort even more. Everyone is free to join the discussion in https://github.com/zigtools/zls/issues/1020.

I agree with employing the strict approach discussed https://github.com/zigtools/zls/issues/1020. The split branch approach seems nice but given the current unstability of Zig, it is probably more trouble than its worth. In this case, I will close the PR in this case as the change doesn't align with the proposed maintenance plan for ZLS.

PR #1589, opened 2023-11-06
show type of primitive instead of value in hover & inlay hint

Techatrix:


nullptrdevs:
<details>
<summary>Is this the way to revive this PR?</summary>

```diff
diff --git a/src/analysis.zig b/src/analysis.zig
index 7151384..4a8ee3c 100644
--- a/src/analysis.zig
+++ b/src/analysis.zig
@@ -925,6 +925,7 @@ fn allDigits(str: []const u8) bool {
 }
 
 const primitives = std.ComptimeStringMap(InternPool.Index, .{
+    // Types
     .{ "anyerror", .anyerror_type },
     .{ "anyframe", .anyframe_type },
     .{ "anyopaque", .anyopaque_type },
@@ -946,7 +947,6 @@ const primitives = std.ComptimeStringMap(InternPool.Index, .{
     .{ "f32", .f32_type },
     .{ "f64", .f64_type },
     .{ "f80", .f80_type },
-    .{ "false", .bool_false },
     .{ "i16", .i16_type },
     .{ "i32", .i32_type },
     .{ "i64", .i64_type },
@@ -954,8 +954,6 @@ const primitives = std.ComptimeStringMap(InternPool.Index, .{
     .{ "i8", .i8_type },
     .{ "isize", .isize_type },
     .{ "noreturn", .noreturn_type },
-    .{ "null", .null_value },
-    .{ "true", .bool_true },
     .{ "type", .type_type },
     .{ "u16", .u16_type },
     .{ "u29", .u29_type },
@@ -964,12 +962,16 @@ const primitives = std.ComptimeStringMap(InternPool.Index, .{
     .{ "u128", .u128_type },
     .{ "u1", .u1_type },
     .{ "u8", .u8_type },
-    .{ "undefined", .undefined_value },
     .{ "usize", .usize_type },
     .{ "void", .void_type },
+    // Values
+    .{ "true", .bool_true },
+    .{ "false", .bool_false },
+    .{ "null", .null_value },
+    .{ "undefined", .undefined_value },
 });
 
-pub fn resolvePrimitiveType(identifier_name: []const u8) ?InternPool.Index {
+pub fn resolvePrimitive(identifier_name: []const u8) ?InternPool.Index {
     if (primitives.get(identifier_name)) |primitive| return primitive;
 
     if (identifier_name.len < 2) return null;
@@ -1099,8 +1101,8 @@ fn resolveTypeOfNodeUncached(analyser: *Analyser, node_handle: NodeWithHandle) e
             const is_escaped_identifier = tree.source[tree.tokens.items(.start)[name_token]] == '@';
             if (!is_escaped_identifier) {
                 if (std.mem.eql(u8, name, "_")) return null;
-                if (resolvePrimitiveType(name)) |primitive| {
-                    const is_type = analyser.ip.indexToKey(primitive).typeOf() == .type_type;
+                if (resolvePrimitive(name)) |primitive| {
+                    const is_type = analyser.ip.indexToKey(primitive) != .simple_value;
                     return TypeWithHandle{
                         .type = .{ .data = .{ .ip_index = .{ .index = primitive } }, .is_type_val = is_type },
                         .handle = handle,
@@ -3984,7 +3986,8 @@ pub fn referencedTypes(
         .ip_index => |payload| {
             const allocator = collector.referenced_types.allocator;
             const ip = analyser.ip;
-            const index = if (resolved_type.type.is_type_val) ip.indexToKey(payload.index).typeOf() else payload.index;
+            const key = ip.indexToKey(payload.index);
+            const index = if (resolved_type.type.is_type_val or key == .simple_value) key.typeOf() else payload.index;
             resolved_type_str.* = try std.fmt.allocPrint(allocator, "{}", .{index.fmt(ip.*)});
         },
         else => {},
@@ -4232,7 +4235,7 @@ fn addReferencedTypes(
                 const name = offsets.identifierTokenToNameSlice(tree, name_token);
                 const is_escaped_identifier = tree.source[tree.tokens.items(.start)[name_token]] == '@';
                 if (is_escaped_identifier) return null;
-                const primitive = Analyser.resolvePrimitiveType(name) orelse return null;
+                const primitive = Analyser.resolvePrimitive(name) orelse return null;
                 return try std.fmt.allocPrint(allocator, "{}", .{primitive.fmt(analyser.ip.*)});
             },
 
diff --git a/src/features/semantic_tokens.zig b/src/features/semantic_tokens.zig
index 18b8e13..8c02543 100644
--- a/src/features/semantic_tokens.zig
+++ b/src/features/semantic_tokens.zig
@@ -128,7 +128,7 @@ const Builder = struct {
 
     fn addDirect(self: *Builder, param_token_type: TokenType, token_modifiers: TokenModifiers, loc: offsets.Loc) error{OutOfMemory}!void {
         std.debug.assert(loc.start <= loc.end);
-        std.debug.assert(self.previous_source_index <= self.source_index); 
+        std.debug.assert(self.previous_source_index <= self.source_index);
         if (loc.start < self.previous_source_index) return;
         if (loc.start < self.source_index) return;
         var token_type = param_token_type;
@@ -701,7 +701,7 @@ fn writeNodeTokens(builder: *Builder, node: Ast.Node.Index) error{OutOfMemory}!v
         .@"asm",
         .asm_simple,
         => {
-            const asm_node: Ast.full.Asm = ast.fullAsm(tree,node).?;
+            const asm_node: Ast.full.Asm = ast.fullAsm(tree, node).?;
 
             try writeToken(builder, main_token, .keyword);
             try writeToken(builder, asm_node.volatile_token, .keyword);
@@ -992,8 +992,8 @@ fn writeIdentifier(builder: *Builder, name_token: Ast.Node.Index) error{OutOfMem
 
     if (!is_escaped_identifier) {
         if (std.mem.eql(u8, name, "_")) return;
-        if (Analyser.resolvePrimitiveType(name)) |primitive| {
-            const is_type = builder.analyser.ip.indexToKey(primitive).typeOf() == .type_type;
+        if (Analyser.resolvePrimitive(name)) |primitive| {
+            const is_type = builder.analyser.ip.indexToKey(primitive) != .simple_value;
             return try writeToken(builder, name_token, if (is_type) .type else .keywordLiteral);
         }
     }
diff --git a/tests/lsp_features/inlay_hints.zig b/tests/lsp_features/inlay_hints.zig
index 2a71e48..e348d79 100644
--- a/tests/lsp_features/inlay_hints.zig
+++ b/tests/lsp_features/inlay_hints.zig
@@ -92,6 +92,12 @@ test "inlayhints - var decl" {
     try testInlayHints(
         \\const foo<comptime_int> = 5;
     , .Type);
+    try testInlayHints(
+        \\const foo<bool> = true;
+    , .Type);
+    try testInlayHints(
+        \\const foo<@TypeOf(undefined)> = undefined;
+    , .Type);
     try testInlayHints(
         \\const foo<**const [3:0]u8> = &"Bar";
     , .Type);
```

</details>

Techatrix:
> Is this the way to revive this PR?

It appears to fix the issue but I believe that the underlying problems are a bit more complex that they may appear.

Let me explain some basics of how I see the InternPool usage for primitives:
```zig
const foo = true;
const bar = bool;
```
In the above code snippet, what InternPool index and  `is_type_val` should `resolveTypeOfNode` return on `foo` and `bar`?

## Option 1: return the typed-value
```zig
// is_type_val is always equivalent to `ip.indexToKey(primitive).typeOf() == .type_type`
const foo = true; // .{ .index = .bool_true, .is_type_val = false }
const bar = bool; // .{ .index = .bool_type, .is_type_val = true }
```

## Option 2: return only the type
Here we only store the type and use `is_type_val` to differentiate between values of a type and the type itself.
This matches how `Analyser.Type` is currently being used.
```zig
const foo = true; // .{ .index = .bool_type, .is_type_val = false }
const bar = bool; // .{ .index = .bool_type, .is_type_val = true }
```

The issue in the codebase is that I've mixed up these two representations. This PR assumed the first option but as an example, this code [here](https://github.com/zigtools/zls/blob/aa9fc0c2959798a59dada05f5c099487eb12c662/src/analysis.zig#L1636-L1697) assumes option 2.

I would like to settle on option 1 because of how it is capable of dealing with values and it also matches how I represent types and values in my sema branch, i.e. everything can be represented with a single intern pool index. Zig internally does the same.

One issue that arises from this is the following:
```zig
// the ip_index of `foo` should be `ip.get(gpa, InternPool.Key{.unknown_value = .{.ty = .bool_type}})` i.e. an unknown value of type bool
// the `instanceTypeVal` function will need to be adjusted to return an unknown value of the given type instead of just changing `is_type_val`.
const foo: bool = some_expression_that_cant_be_resolved;
```

Keep in mind that the `get` function of the InternPool is not thread safe so the intern pool must be "warmed up" by inserting all unknown value representation of all primitives so that we can use the `contains` function instead.

---

I do think that you can still go ahead with your suggested changes since it still fixes the issue I've had. Feel free to PR.

One think that I didn't understand was this change though. Aren't these expressions equivalent (for primitives)?
```diff
- const is_type = analyser.ip.indexToKey(primitive).typeOf() == .type_type;
+ const is_type = analyser.ip.indexToKey(primitive) != .simple_value;
```

Issue #1543, opened 2023-10-23
Unused variables need two saves to get formatted properly


TheFunctionalGuy:
### Zig Version

0.12.0-dev.1200+5f92b070b

### Zig Language Server Version

0.12.0-dev.185+645f728

### Steps to Reproduce

1. Implement this kind of code:
```rust
pub fn format(self: Self, comptime fmt: []const u8, options: std.fmt.FormatOptions, writer: anytype) !void {}
```
2. Press save so formatting kicks in

_Note:_ I would kindly provide more information when you need. Just tell me whether you need it.

### Expected Behavior

Getting this code after **ONE** save:
```rust
pub fn format(self: Self, comptime fmt: []const u8, options: std.fmt.FormatOptions, writer: anytype) !void {
    _ = writer;
    _ = options;
    _ = fmt;
    _ = self;
}
```

### Actual Behavior

Getting the following code after the **first** save:
```rust
pub fn format(self: Self, comptime fmt: []const u8, options: std.fmt.FormatOptions, writer: anytype) !void {
    _ = writer;
    _ = options;
    _ = fmt;
    _ = self;}
```
And after a **second** save:
```rust
pub fn format(self: Self, comptime fmt: []const u8, options: std.fmt.FormatOptions, writer: anytype) !void {
    _ = writer;
    _ = options;
    _ = fmt;
    _ = self;
}
```

Koranir:
Relevant: #1490

PR #1597, opened 2023-11-11
BUGFIX | Fix `autofix` misplacing `}` when resolving unused function parameters, and captures.

alexatcanva:
# What is this ? 

Hi All :wave: 

This is my first PR contributing to `zls`, and it's also my first Zig PR (ever), so if at any point during this PR process I'm a little slow to pick things up please forgive me :pray:

This PR is aimed at resolving issue https://github.com/zigtools/zls/issues/1543 

The issue occurs when you use the `autofix` functionality of `zls` to automatically provide discarded variable usages inside your function block, or inside your capture block. 

An example can be seen like so: 

```zig 

pub fn format(a: u8, b: u8, c: u8) !void {
}

// Then after you save:


pub fn format(a: u8, b: u8, c: u8) !void {
    _ = c;
    _ = b;
    _ = a;}

```

This PR addresses this issue so the output of the `autofix` should now be like: 

```zig

pub fn format(a: u8, b: u8, c: u8) !void {
    _ = a;
    _ = b;
    _ = c;
}
```

This behaviour can also be noted for the capture discards: 

```zig

// old

pub fn format(a: u8, b: u8, c: u8) !void {
    _ = c;
    _ = b;
    _ = a;

    for (0..10) |ix| {
        _ = ix;}
}

// new

pub fn format(a: u8, b: u8, c: u8) !void {
    _ = a;
    _ = b;
    _ = c;

    for (0..10) |ix| {
        _ = ix;
    }
}

```

Let me know if I've missed anything here! Thanks!

### DIFF
--- a/src/features/code_actions.zig
+++ b/src/features/code_actions.zig
@@ -8,6 +8,8 @@ const types = @import("../lsp.zig");
 const offsets = @import("../offsets.zig");
 const tracy = @import("../tracy.zig");
 
+const log = std.log.scoped(.zls_server);
+
 pub const Builder = struct {
     arena: std.mem.Allocator,
     analyser: *Analyser,
@@ -18,15 +20,18 @@ pub const Builder = struct {
         builder: *Builder,
         diagnostic: types.Diagnostic,
         actions: *std.ArrayListUnmanaged(types.CodeAction),
+        action_index: usize,
         remove_capture_actions: *std.AutoHashMapUnmanaged(types.Range, void),
     ) error{OutOfMemory}!void {
         const kind = DiagnosticKind.parse(diagnostic.message) orelse return;
 
         const loc = offsets.rangeToLoc(builder.handle.tree.source, diagnostic.range, builder.offset_encoding);
 
+        log.debug("code action for diagnostic: {s}", .{diagnostic.message});
+
         switch (kind) {
             .unused => |id| switch (id) {
-                .@"function parameter" => try handleUnusedFunctionParameter(builder, actions, loc),
+                .@"function parameter" => try handleUnusedFunctionParameter(builder, actions, action_index, loc),
                 .@"local constant" => try handleUnusedVariableOrConstant(builder, actions, loc),
                 .@"local variable" => try handleUnusedVariableOrConstant(builder, actions, loc),
                 .@"switch tag capture", .capture => try handleUnusedCapture(builder, actions, loc, remove_capture_actions),
@@ -81,7 +86,12 @@ fn handleNonCamelcaseFunction(builder: *Builder, actions: *std.ArrayListUnmanage
     try actions.append(builder.arena, action1);
 }
 
-fn handleUnusedFunctionParameter(builder: *Builder, actions: *std.ArrayListUnmanaged(types.CodeAction), loc: offsets.Loc) !void {
+fn handleUnusedFunctionParameter(
+    builder: *Builder,
+    actions: *std.ArrayListUnmanaged(types.CodeAction),
+    action_index: usize,
+    loc: offsets.Loc,
+) !void {
     const tracy_zone = tracy.trace(@src());
     defer tracy_zone.end();
 
@@ -109,7 +119,11 @@ fn handleUnusedFunctionParameter(builder: *Builder, actions: *std.ArrayListUnman
 
     const block = node_datas[payload.func].rhs;
 
-    const new_text = try createDiscardText(builder, identifier_name, token_starts[node_tokens[payload.func]], true);
+    const new_text = try if (action_index == actions.items.len)
+        // If our fix is on the last item, we need to append an additional newline.
+        createDiscardText(builder, identifier_name, token_starts[node_tokens[payload.func]], true, true)
+    else
+        createDiscardText(builder, identifier_name, token_starts[node_tokens[payload.func]], true, false);
 
     const index = token_starts[node_tokens[block]] + 1;
 
@@ -159,7 +173,7 @@ fn handleUnusedVariableOrConstant(builder: *Builder, actions: *std.ArrayListUnma
     if (last_token >= tree.tokens.len) return;
     if (token_tags[last_token] != .semicolon) return;
 
-    const new_text = try createDiscardText(builder, identifier_name, token_starts[first_token], false);
+    const new_text = try createDiscardText(builder, identifier_name, token_starts[first_token], false, false);
 
     const index = token_starts[last_token] + 1;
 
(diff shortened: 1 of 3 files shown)


src/features/code_actions.zig
```diff
@@ -509,6 +558,16 @@ fn getCaptureLoc(text: []const u8, loc: offsets.Loc) ?offsets.Loc {
...
 }
 
+fn prependSlice(list: *std.ArrayListUnmanaged(types.CodeAction), arena: std.mem.Allocator, values: []const types.CodeAction) !void {
+    const old_len = list.items.len;
+    try list.resize(arena, list.items.len + values.len);
+    var i = old_len + values.len - 1;
+    while (i >= values.len) : (i -= 1) {
+        list.items[i] = list.items[i - values.len];
+    }
+    std.mem.copy(types.CodeAction, list.items[0..values.len], values);
```
Techatrix, line 568: The `insertSlice` function should provide what you need. I would suggest to remove the `prependSlice` function all together and just call `insertSlice` directly.
```zig
try list.insertSlice(arena, 0, values);
```

src/features/code_actions.zig
```diff
@@ -109,7 +109,35 @@ fn handleUnusedFunctionParameter(builder: *Builder, actions: *std.ArrayListUnman
...
+    // To do this, you may have a function like:
+    // fn(a: i32, b: i32, c: i32) void { _ = a; _ = b; _ = c; }
+    // or
+    // fn(
+    //     a: i32,
+    //     b: i32,
+    //     c: i32,
+    // ) void { ... }
+    // We have to be able to detect both cases.
+    var param_end = offsets.tokenToLoc(tree, ast.paramLastToken(tree, payload.get(tree).?)).end;
```
Techatrix, line 123: Further towards the end of this function `payload.get(tree)` is being called again. I would suggest to do it once and store into a new variable.

src/features/code_actions.zig
```diff
@@ -109,7 +109,35 @@ fn handleUnusedFunctionParameter(builder: *Builder, actions: *std.ArrayListUnman
...
+                found_comma = true;
+                continue;
+            } else {
+                param_end += 1;
+                break;
+            },
+            ')' => break,
+            else => break,
+        }
+    }
```
Techatrix, line 138: Looks like you copied this code from the `getParamRemovalRange` function. You do not use `param_end` after iterating so there is no need to continue after encountering a comma.

This means that we can write the following:
```zig
var found_comma: bool = false;
while (param_end < tree.source.len) : (param_end += 1) {
    switch (tree.source[param_end]) {
        ' ', '\n' => continue,
        ',' => {
            found_comma = true;
            break;
        },
        ')' => break,
        else => break,
    }
}
```

If we use some of the functions from the `std.mem` we can also do this:
```zig
const found_comma = std.mem.startsWith(u8, std.mem.trimLeft(u8, tree.source[param_end..], " \n"), ",");
```

src/features/code_actions.zig
```diff
@@ -310,7 +352,7 @@ fn createCamelcaseText(allocator: std.mem.Allocator, identifier: []const u8) ![]
 }
 
 // returns a discard string `\n{indent}_ = identifier_name;`
-fn createDiscardText(builder: *Builder, identifier_name: []const u8, declaration_start: usize, add_block_indentation: bool) ![]const u8 {
+fn createDiscardText(builder: *Builder, identifier_name: []const u8, declaration_start: usize, add_block_indentation: bool, add_suffix_newline: bool) ![]const u8 {
```
Techatrix, line 355: Could you add a comma after the last parameter and then run `zig fmt` so the parameters gets spread out over multiple lines?

src/features/code_actions.zig
```diff
@@ -321,16 +363,23 @@ fn createDiscardText(builder: *Builder, identifier_name: []const u8, declaration
...
 
     new_text.appendAssumeCapacity('\n');
     new_text.appendSliceAssumeCapacity(indent);
     new_text.appendSliceAssumeCapacity(additional_indent);
     new_text.appendSliceAssumeCapacity("_ = ");
     new_text.appendSliceAssumeCapacity(identifier_name);
     new_text.appendAssumeCapacity(';');
+    // if we are adding a suffix newline, we need to add the additional indentation again.
+    new_text.appendSliceAssumeCapacity(additional_newline);
+    if (add_suffix_newline) new_text.appendSliceAssumeCapacity(indent);
```
Techatrix, line 382:
```suggestion
    if (add_suffix_newline) {
        new_text.appendAssumeCapacity('\n');
        new_text.appendSliceAssumeCapacity(indent);
    }
```

src/features/code_actions.zig
```diff
@@ -321,16 +363,23 @@ fn createDiscardText(builder: *Builder, identifier_name: []const u8, declaration
...
     const additional_indent = if (add_block_indentation) detectIndentation(builder.handle.tree.source) else "";
+    const additional_newline = if (add_suffix_newline) "\n" else "";
 
-    const new_text_len = 1 + indent.len + additional_indent.len + "_ = ;".len + identifier_name.len;
-    var new_text = try std.ArrayListUnmanaged(u8).initCapacity(builder.arena, new_text_len);
+    var new_text_len = 1 + indent.len + additional_indent.len + additional_newline.len + "_ = ;".len + identifier_name.len;
+    // if we have an additional newline, then we will need an additional indent too.
+    if (add_suffix_newline) new_text_len += indent.len;
+    const post_new_len = new_text_len;
+    var new_text = try std.ArrayListUnmanaged(u8).initCapacity(builder.arena, post_new_len);
```
Techatrix, line 372:
```suggestion
    const new_text_len =
        1 +
        indent.len +
        additional_indent.len +
        "_ = ;".len +
        identifier_name.len +
        if (add_suffix_newline) 1 + indent.len else 0;
    var new_text = try std.ArrayListUnmanaged(u8).initCapacity(builder.arena, new_text_len);
```

Techatrix (changes requested):
Hi @alexatcanva  👋

I've had some comment but those are mostly code simplification or formatting and nothing critical. Overall you did a good job.

The most important thing is the missing test coverage. The current code action test were formatting the source code before comparing so there was no way to check for correctly formatted output.

The following patch should remove the formatting pass (and adjust one test case).
Most code action tests fail after this patch but not after your PR which shows that you produce code that is formatted just like `zig fmt`.

```patch
diff --git a/tests/lsp_features/code_actions.zig b/tests/lsp_features/code_actions.zig
index 5f9ae04..8d9796c 100644
--- a/tests/lsp_features/code_actions.zig
+++ b/tests/lsp_features/code_actions.zig
@@ -85,10 +85,11 @@ test "code actions - discard captures" {
 test "code actions - discard capture with comment" {
     try testAutofix(
         \\test {
-        \\  if (1 == 1) |a|
-        \\      //a
-        \\      {}
+        \\    if (1 == 1) |a|
+        \\    //a
+        \\    {}
         \\}
+        \\
     ,
         \\test {
         \\    if (1 == 1) |a|
@@ -174,9 +175,5 @@ fn testAutofix(before: []const u8, after: []const u8) !void {
     defer allocator.free(actual);
     try ctx.server.document_store.refreshDocument(uri, try allocator.dupeZ(u8, actual));
 
-    try std.testing.expect(handle.tree.errors.len == 0);
-    const formatted_actual = try handle.tree.render(allocator);
-    defer allocator.free(formatted_actual);
-
-    try std.testing.expectEqualStrings(after, formatted_actual);
+    try std.testing.expectEqualStrings(after, handle.tree.source);
 }
```

alexatcanva:
Wow! @Techatrix thank you so much for taking the time to provide such an in-depth review! 🙏 

I really appreciate the effort you have spent here to clearly outline and explain the improvements that I could have made in here. This type of feedback is invaluable while I am trying to learn the ins-and-outs of Zig and zls! 

I have made the relevant adjustments to this PR as per your review comments, please let me know if there's anything else I can do to assist further! 

Thanks again! 🚀

Techatrix (approved):
:+1:

PR #1604, opened 2023-11-16
Support load http proxy environment variables

jimying:
zig project (https://github.com/ziglang/zig/pull/17407) add client.loadDefaultProxies to loads proxy environment variables (http_proxy, HTTP_PROXY, https_proxy, HTTPS_PROXY, all_proxy, ALL_PROXY).


nullptrdevs (approved):
Neat

PR #1605, opened 2023-11-16
Zig AstGen changes, unnecessary use of 'var'

nullptrdevs:
- https://github.com/ziglang/zig/commit/941090d94f6b6c7c5207f06917743e79ff06426f
- https://github.com/ziglang/zig/pull/18017

This can be merged as-is, or we can wait and see if I missed changing a `var` to `const` somewhere.

### DIFF
--- a/src/analyser/encoding.zig
+++ b/src/analyser/encoding.zig
@@ -146,8 +146,8 @@ pub fn decode(extra: *[]const u8, comptime T: type) T {
         },
         .Array => |info| blk: {
             var array: T = undefined;
-            const i: usize = 0;
-            while (i < info.len) {
+            var i: usize = 0;
+            while (i < info.len) : (i += 1) {
                 array[i] = decode(extra, info.child);
             }
             break :blk array;
--- a/src/analyser/encoding.zig
+++ b/src/analyser/encoding.zig
@@ -146,7 +146,7 @@ pub fn decode(extra: *[]const u8, comptime T: type) T {
         },
         .Array => |info| blk: {
             var array: T = undefined;
-            var i: usize = 0;
+            const i: usize = 0;
             while (i < info.len) {
                 array[i] = decode(extra, info.child);
             }
--- a/src/binned_allocator.zig
+++ b/src/binned_allocator.zig
@@ -137,13 +137,13 @@ pub fn BinnedAllocator(comptime config: Config) type {
                 fn init() @This() {
                     return .{ .count = std.atomic.Atomic(usize).init(0) };
                 }
-                fn load(self: @This()) usize {
+                fn load(self: *const @This()) usize {
                     return self.count.load(.Acquire);
                 }
-                fn increment(self: @This()) void {
+                fn increment(self: *@This()) void {
                     _ = self.count.fetchAdd(1, .AcqRel);
                 }
-                fn decrement(self: @This()) void {
+                fn decrement(self: *@This()) void {
                     _ = self.count.fetchSub(1, .AcqRel);
                 }
             }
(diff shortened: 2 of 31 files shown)


src/analyser/encoding.zig
```diff
@@ -146,7 +146,7 @@ pub fn decode(extra: *[]const u8, comptime T: type) T {
         },
         .Array => |info| blk: {
             var array: T = undefined;
-            var i: usize = 0;
+            const i: usize = 0;
             while (i < info.len) {
                 array[i] = decode(extra, info.child);
             }
```
Techatrix, line 152: interesting :eyes:

nullptrdevs: Would you like me to fix it in this PR or leave it to you to handle in the "intern pool rewrite" PR?

Techatrix: My IP work will take some time so I would suggest to fix it in this PR. Should be quite obvious what to do.

Techatrix:
Our dependencies will also need to be updated:
- https://github.com/ziglibs/diffz/pull/16
- https://github.com/ziglibs/known-folders/pull/41
- [binned_allocator.zig](https://gist.github.com/antlilja/8372900fcc09e38d7b0b6bbaddad3904) How am I supposed to edit/fix/PR a gist?

nullptrdevs:
>     * [binned_allocator.zig](https://gist.github.com/antlilja/8372900fcc09e38d7b0b6bbaddad3904) How am I supposed to edit/fix/PR a gist?

That gist has changed hands as many times as it's been updated, you have the choice of converting in to a repo.

Techatrix:
@SuperAuguste Can I create this as new repository in zigtools?

SuperAuguste:
sure @Techatrix! tbh I don't believe there's a lot of external interest in using the slab allocator so you could also just plop it in a file in the zls repo :P

Techatrix:
I guess that would also work. I though there were other project that also used the binned allocator but sourcegraph search couldn't find any.

@nullptrdevs Would you mind If I added the binned allocator in your PR directly?

nullptrdevs:
> @nullptrdevs Would you mind If I added the binned allocator in your PR directly?

Go ahead, and always feel free to make changes/push commits to any of my PRs !

src/binned_allocator.zig
```diff
@@ -0,0 +1,527 @@
...
+                fn load(self: @This()) usize {
+                    return self.count;
+                }
+                fn increment(self: @This()) void {
+                    self.count += 1;
+                }
+                fn decrement(self: @This()) void {
+                    self.count -= 1;
+                }
+            };
```
ianprime0509, line 165:
```suggestion
        const Counter = if (config.thread_safe)
            struct {
                count: std.atomic.Atomic(usize),
                fn init() @This() {
                    return .{ .count = std.atomic.Atomic(usize).init(0) };
                }
                fn load(self: *const @This()) usize {
                    return self.count.load(.Acquire);
                }
                fn increment(self: *@This()) void {
                    _ = self.count.fetchAdd(1, .AcqRel);
                }
                fn decrement(self: *@This()) void {
                    _ = self.count.fetchSub(1, .AcqRel);
                }
            }
        else
            struct {
                count: usize,
                fn init() @This() {
                    return .{ .count = 0 };
                }
                fn load(self: @This()) usize {
                    return self.count;
                }
                fn increment(self: *@This()) void {
                    self.count += 1;
                }
                fn decrement(self: *@This()) void {
                    self.count -= 1;
                }
            };
```
I believe the `*const @This()` is necessary for the atomic `load` function to ensure correct semantics (avoid a non-atomic copy of `count` when passing the argument).

SuperAuguste (approved):
LGTM; took it for a spin on ZLS for a bit and didn't encounter any crashes or anything :) (every PR I merge without fuzzing reminds me that I should really finish that code :/)

PR #1606, opened 2023-11-16
build_runners: Attempt to make GeneratedFile steps

nullptrdevs:
+ Use `original handle` when trying to resolve field access, ie the `microzig` case, where a dep requests the app to import one of it's own app modules - `app_dep.app_mod.some`.

### DIFF
--- a/src/analysis.zig
+++ b/src/analysis.zig
@@ -25,15 +25,18 @@ resolved_callsites: std.AutoHashMapUnmanaged(Declaration.Param, ?TypeWithHandle)
 resolved_nodes: std.HashMapUnmanaged(NodeWithUri, ?TypeWithHandle, NodeWithUri.Context, std.hash_map.default_max_load_percentage) = .{},
 /// used to detect recursion
 use_trail: NodeSet = .{},
+/// handle of the doc where the request originated
+root_handle: ?*DocumentStore.Handle = undefined,
 
 const NodeSet = std.HashMapUnmanaged(NodeWithUri, void, NodeWithUri.Context, std.hash_map.default_max_load_percentage);
 
-pub fn init(gpa: std.mem.Allocator, store: *DocumentStore, ip: *InternPool) Analyser {
+pub fn init(gpa: std.mem.Allocator, store: *DocumentStore, ip: *InternPool, root_handle: ?*DocumentStore.Handle) Analyser {
     return .{
         .gpa = gpa,
         .arena = std.heap.ArenaAllocator.init(gpa),
         .store = store,
         .ip = ip,
+        .root_handle = root_handle,
     };
 }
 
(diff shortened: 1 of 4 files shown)


Techatrix:
> * Use `original handle` when trying to resolve field access, ie the `microzig` case, where a dep requests the app to import one of it's own app modules - `app_dep.app_mod.some`.

I am not familiar with how `microzig` works as a dependency. Could you give me a usage example of `microzig` that would showcase the issue that this PR addresses so I can figure out if this new `root_handle` is a good solution?

nullptrdevs:
> > * Use `original handle` when trying to resolve field access, ie the `microzig` case, where a dep requests the app to import one of it's own app modules - `app_dep.app_mod.some`.
> 
> I am not familiar with how `microzig` works as a dependency. Could you give me a usage example of `microzig` that would showcase the issue that this PR addresses so I can figure out if this new `root_handle` is a good solution?

Yes, certainly -

An example microzig app https://github.com/ZigEmbeddedGroup/microzig-examples/tree/main/espressif-esp
common pattern https://github.com/ZigEmbeddedGroup/microzig-examples/blob/04f2c4a6ff5f4b20f2848175082f98409ecc0ede/espressif-esp/src/blinky.zig#L2-L3
these are actually app's modules https://github.com/ZigEmbeddedGroup/microzig/blob/f8a506ccc28447ff6f20bba13ad04e7ce8d26318/src/microzig.zig#L13-L26

So the app imports the `microzig` dep which uses the app's modules in the context of the app.

The app would also work if those [lines](https://github.com/ZigEmbeddedGroup/microzig/blob/f8a506ccc28447ff6f20bba13ad04e7ce8d26318/src/microzig.zig#L13-L26) were part of the app, but it's probably a convenience thing.

src/analysis.zig
```diff
@@ -25,15 +25,18 @@ resolved_callsites: std.AutoHashMapUnmanaged(Declaration.Param, ?TypeWithHandle)
 resolved_nodes: std.HashMapUnmanaged(NodeWithUri, ?TypeWithHandle, NodeWithUri.Context, std.hash_map.default_max_load_percentage) = .{},
 /// used to detect recursion
 use_trail: NodeSet = .{},
+/// handle of the doc where the request originated
+root_handle: *DocumentStore.Handle = undefined,
```
Techatrix, line 29: Could you make this an optional pointer. The analysis backend should not need to know about what a LSP request is.

Techatrix (changes requested):

src/analysis.zig
nullptrdevs: Maybe I've worded it poorly but it's simply the handle of the first node any lookup happens on, ie given a `some_import.dependency_var.other` it's the handle `some_import` node belongs to, so this can work:
`some_import` belongs to a handle (A)  with a build file, and
`dependency_var` belongs to a handle (B)  with it's own build file, yet it contains an `@import` which references a module of handle A.

Given that zig allows it, the real proper solution would be to keep a chain of handles when doing lookup and walk it up in such cases.

Anyhow, you have the big picture of the analyzer, so please make any changes you deem necessary.

Techatrix: I do understand the reasoning behind the root handle.
I just don't think that initializing an `Analyser` should force specifying a root handle because there may not be a root handle in certain situations. One example would be when handling a `workspace/symbol` request where no root handle exists.

nullptrdevs: Got it ! .. and as always, open to suggestions when it comes to naming variables :face_exhaling:

Techatrix (approved):

PR #1607, opened 2023-11-16
add autofix for `local variable is never mutated` error

Techatrix:
This will add autofix support for the new error that will be produced by https://github.com/ziglang/zig/pull/18017

Must be a hassle to fix all these errors without autofix...


mlugg:
> Must be a hassle to fix all these errors without autofix...

yes
image

SuperAuguste (approved):
common techatrix w

PR #1612, opened 2023-11-21
Update `deps.nix`

sagehane:
A recent commit (probably https://github.com/zigtools/zls/commit/e8951fc535f3750a417b15cf0aaca2076949f2e1) broke the build for Nix flakes.

The `deps.nix` file was manually edited as `zon2nix` currently doesn't support the new ZON format. https://github.com/nix-community/zon2nix/issues/6 seems to be the upstream issue.

### DIFF
--- a/deps.nix
+++ b/deps.nix
@@ -4,17 +4,17 @@
 
 linkFarm "zig-packages" [
   {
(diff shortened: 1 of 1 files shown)


SuperAuguste (approved):
Thanks!

PR #1615, opened 2023-11-22
fix error in binned_allocator when report_leaks is disabled

xdBronch:
CI didnt catch this because `report_leaks` is enabled by default


nektro (approved):

SuperAuguste (approved):
thanks!

PR #1619, opened 2023-11-23
call `zig init` instead of `zig init-exe` build_runner.yml

Techatrix:
see https://github.com/ziglang/zig/pull/18055


SuperAuguste (approved):
:)

PR #1627, opened 2023-11-24
goto test coverage + improvements

Techatrix:
These changes allow testing all 3 goto requests individually:
- textDocument/declaration
- textDocument/definition
- textDocument/typeDefinition

Here is one example of all three in action:
```zig
test {
    // - use `<>` to indicate the cursor position
    // - use `<decl>content</decl>` to set the expected range of the declaration
    // - use `<def>content</def>` to set the expected range of the definition
    // - use `<tdef>content</tdef>` to set the expected range of the type definition
    try testDefinition(
        \\const <def>foo</def> = 5;
        \\const <decl>bar</decl>: <tdef>u32</tdef> = foo;
        \\comptime {
        \\    _ = <>bar;
        \\}
    );
}
```

I've also made `textDocument/typeDefinition` give the type declaration instead of type definition because otherwise goto on primitive types would be completely broken.

### DIFF
--- a/tests/lsp_features/definition.zig
+++ b/tests/lsp_features/definition.zig
@@ -63,10 +63,16 @@ fn testDefinition(source: []const u8) !void {
     var phr = try helper.collectClearPlaceholders(allocator, source);
     defer phr.deinit(allocator);
 
+    var ctx = try Context.init();
+    defer ctx.deinit();
+
+    const test_uri = try ctx.addDocument(phr.new_source);
+
     var error_builder = ErrorBuilder.init(allocator);
     defer error_builder.deinit();
     errdefer error_builder.writeDebug();
 
+    try error_builder.addFile(test_uri, phr.new_source);
     try error_builder.addFile("old_source", source);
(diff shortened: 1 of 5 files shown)


tests/lsp_features/definition.zig
```diff
@@ -12,24 +11,139 @@ const offsets = zls.offsets;
...
+        \\    var maybe: ?S = 5;
+        \\    if (maybe) |<>some| {}
+        \\}
+    );
+    if (true) return error.SkipZigTest; // TODO
+    // primitives like `u32` are represented as a InternPool.Index so they
+    // don't have a Ast.Node.Index that gives them a source location
+    try testDefinition(
+        \\test {
+        \\    var maybe: <tdef>?u32<&tdef> = 5;
```
nullptrdevs, line 104: Is the `<&tdef>` a typo or special syntax?

Techatrix: just a typo. The `&` and `/` are next to each other on a german keyboard.

nullptrdevs (approved):

PR #1631, opened 2023-11-26
extend CompletionItemKind and DiagnosticTag enums

aoime:
supports forward and backward compatibility according to LSP specification (also allows VS 2022 clients to use zls)

### DIFF
--- a/src/lsp.zig
+++ b/src/lsp.zig
@@ -684,6 +684,7 @@ pub const CompletionItemKind = enum(u32) {
     Event = 23,
     Operator = 24,
     TypeParameter = 25,
+    _,
     pub usingnamespace EnumStringifyAsInt(@This());
 };
 
@@ -930,7 +931,7 @@ pub const DiagnosticSeverity = enum(u32) {
 /// The diagnostic tags.
 ///
 /// @since 3.15.0
-pub const DiagnosticTag = enum(u32) {
+pub const DiagnosticTag = enum(i32) {
     /// Unused or unnecessary code.
     ///
     /// Clients are allowed to render diagnostics with this tag faded out instead of having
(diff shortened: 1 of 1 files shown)


aoime:
Ah, sorry about that. I totally missed the code-gen notice. `//! generated by zig-lsp-codegen`

The client is Visual Studio 2022 Community (not VS Code).  The initialization handshake from the client includes extended CompletionItemKind and DiagnosticTag values that cause the current ZLS to fail the handshake.

The way I read the latest LSP spec is that the section under [enumerations](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#enumerations) specifically allows for all enumerations to be extended (including signed int values < 0) for back and future compatibility.  The VS 2022 client (v17.8.1) sends these values in the initialize method:
`"completionItemKind":{"valueSet":[0,1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20,21,22,23,24,25,118115,118116,118117,118118,118119,118120,118121,118122,118123,118124,118125,118126]}`
`"publishDiagnostics":{"tagSupport":{"valueSet":[1,2,-1,-2]}}`

If I can get past the handshake the rest of ZLS works wonderfully in VS 2022.

So obviously this PR should be killed but I feel the code-gen should be updated to supported this. If you agree with my interpretation of the spec I'd like to try again to make this work.

Techatrix:
> I'm following up with the metaModel.json owners on the potential conflict between the metaModel.json and the text of the specification as regards to unknown enumeration values and how would they advise the metaModel.json be leveraged in a code generation process that aligns with the spec text.

[#1847](https://github.com/microsoft/language-server-protocol/issues/1847)
Once the metaModel.json gets updated, we can then regenerate ZLS's LSP type definitions.

PR #1636, opened 2023-11-28
fix invalid unwrap when failing to create build file

Techatrix:
`getBuildFile` may return null when creating the build file failed.


nullptrdevs (commented):
I know there are a few more call sites, are they covered?

Techatrix:
> I know there are a few more call sites, are they covered?

I've looked again and realized that I've missed one call and one of my changes was unnecessary which also applies to all the other callsites of `getBuildFile` which I didn't edit because of the following guarantee:
https://github.com/zigtools/zls/blob/a3145f64a8d4dfbb2e9f5b906405aa00a5c17602/src/DocumentStore.zig#L160-L162

nullptrdevs (approved):
Got it

Issue #1617, opened 2023-11-22
completions for anonymous struct parameters only happen if the struct type and function are defined in a different file


xdBronch:
### Zig Version

0.12.0-dev.1680+be6f76655

### Zig Language Server Version

0.12.0-dev.248+62f17ab

### Steps to Reproduce

```zig
// main.zig
pub fn main() void {
    foo(.{<cursor>});
}

const Config = struct {
    conf: void,
};

fn foo(_: Config) void {}
```
request completions, none appear. move `Config` and `foo` into a separate file
```zig
// import.zig
pub const Config = struct {
    conf: void,
};

pub fn foo(_: Config) void {}
/////////////////////////////
// main.zig
const import = @import("import.zig");

pub fn main() void {
    import.foo(.{<cursor>});
}
```
request completions and they correctly appear

### Expected Behavior

completions should always be given

### Actual Behavior

inconsistent

nullptrdevs:
Parser limitation - try it with #1536 or `1.` instead of just `.`.

It has to do with syntax errors and the type node being after the current node.

xdBronch:
i have another issue to report where completions for configs like this are given instead of for the struct fields in cases where the function returns a type. is this related? i can open a new issue if not

nullptrdevs:
It's the `std.heap.GeneralPurposeAllocator(.{}){.<cursor>}` issue isn't it.. :sweat_smile:
I keep forgetting about it so an issue as a reminder wouldn't be bad.

xdBronch:
yeah thats it lol, will open an issue

xdBronch:
#1618 i also noticed that `@import("std").heap.GeneralPurposeAllocator(.{}){}` doesnt give any completions so i had to split the import into a const, is that the same issue as this?

nullptrdevs:
`collectFieldAccessContainerNodes` expects the first symbol to be something it can look up to determine if it's a `T` or `instance_of_T`
https://github.com/zigtools/zls/blob/71d2c3302acd4df716d39adb5d6797c2075bd946/src/features/completions.zig#L1375-L1382

A `if (field_access_slice[0] == '@') break :blk 1; // assume @import("..").some.Other{.}` could work.

PR #1638, opened 2023-11-29
Some `.enum_literal` completions fixes

### LINKED ISSUE #1618: type functions that accept struct arguments give completions for the argument struct instead of the returned struct
### Zig Version

0.12.0-dev.1680+be6f76655

0.12.0-dev.248+62f17ab

### Steps to Reproduce

```zig
const std = @import("std");

pub fn main() void {
   _ = std.heap.GeneralPurposeAllocator(.{}){<cursor>};
}
```
request completions

### Expected Behavior

list of completions should be of the fields from the GPA type

### Actual Behavior

fields from the config param are given instead

nullptrdevs:
Fixes #1618 and https://github.com/zigtools/zls/issues/1617#issuecomment-1823583412

### DIFF
--- a/src/features/completions.zig
+++ b/src/features/completions.zig
@@ -1040,6 +1040,7 @@ const EnumLiteralContext = struct {
     likely: Likely,
     identifier_token_index: Ast.TokenIndex = 0,
     fn_arg_index: usize = 0,
+    need_ret_type: bool = false,
 };
 
 fn getEnumLiteralContext(
@@ -1066,7 +1067,7 @@ fn getEnumLiteralContext(
             dot_context.identifier_token_index = token_index;
         },
         .l_brace, .comma, .l_paren => {
-            dot_context = getStructInitContext(tree, dot_token_index) orelse getSwitchContext(tree, dot_token_index) orelse return null;
+            dot_context = getSwitchOrStructInitContext(tree, dot_token_index) orelse return null;
         },
         else => return null,
     }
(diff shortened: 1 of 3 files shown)


src/features/completions.zig
```diff
@@ -1366,13 +1335,27 @@ fn collectFieldAccessContainerNodes(
...
+                const body = node_type.handle.tree.nodes.items(.data)[node_type.type.data.other].rhs;
+                node_type = try analyser.resolveReturnType(full_fn_proto, node_type.handle, if (has_body) body else null) orelse return;
+                // `E!T`
+                if (node_type.type.data == .other and node_type.handle.tree.nodes.items(.tag)[node_type.type.data.other] == .error_union) {
+                    const node_data = node_type.handle.tree.nodes.items(.data)[node_type.type.data.other];
+                    if (node_data.rhs == 0) return;
+                    node_type = try analyser.resolveTypeOfNode(.{ .node = node_data.rhs, .handle = node_type.handle }) orelse return;
+                }
+                try node_type.getAllTypesWithHandlesArrayList(arena, types_with_handles);
+                return;
```
Techatrix, line 1349: This code snippet appears three times in your dot completion code. Perhaps its better move this logic into analysis.
The code duplication seems to have caused a mistake because you are doing `return` instead of `continue` on failure (you are in a for loop here).

I guess you are often looking for a function that does the following conversion:
```rust
match(T) {
    ?P -> P,
    E!P -> P,
    ?E!P -> P,
    else -> T,
}
```
Its like the reverse of type coercion.

Techatrix (changes requested):

Techatrix (approved):

PR #1644, opened 2023-12-01
callsite references related optimizations

### LINKED ISSUE #1635: ZLS timeout when using (huge?) cimport
### Zig Version

0.12.0-dev.1746+19af8aac8

0.12.0-dev.261+f16bfd3

### OS
Windows 10

### Steps to Reproduce

1. Create a new project
2. Add the following import
```
const d3dx9 = @cImport({
    @cInclude("d3dx9.h");
});
```
3. type `d3dx9.`

### Expected Behavior

zls show the d3dx9 components.

### Actual Behavior

- zls process, that usually use less than 15 MB, in 3 seconds jumps to 100MB
- it keep processing using 100% of a cpu core
- after 10~ minutes it dies (or nvim or windows kills it)
  - nvim shows the message `Client 1 quit with exit code 3221225477 and signal 0`

### Additional Details
The `cimport.zig`  with d3dx9.h content generated by zig has more then 7 MB [cimport.zip](https://github.com/zigtools/zls/files/13485429/cimport.zip)
lsp log https://pastebin.com/Mv01eFDe

Techatrix:
fixes #1635
~~I will have to look if I can produce a simple reproduction to convert into a test case. Maybe I will then find a better solution for this problem.~~
I may not know what exactly is going on, but resolving recursively resolving callsite references is bad. So just stop doing that.

# Before 
Screenshot from 2023-12-01 19-53-50

# After

This graph is without disabling callsite references completely in `cimport.zig` files. So only resolving callsite references while resolving callsite references is prevented.

Screenshot from 2023-12-01 20-43-31

---

The given `cimport.zig` in #1635 is also a good showcase of the parallelization of ZLS

Screenshot from 2023-12-01 20-42-33-with-notes

### DIFF
--- a/src/analysis.zig
+++ b/src/analysis.zig
@@ -25,8 +25,9 @@ resolved_callsites: std.AutoHashMapUnmanaged(Declaration.Param, ?TypeWithHandle)
 resolved_nodes: std.HashMapUnmanaged(NodeWithUri, ?TypeWithHandle, NodeWithUri.Context, std.hash_map.default_max_load_percentage) = .{},
 /// used to detect recursion
 use_trail: NodeSet = .{},
+collect_callsite_references: bool,
 /// handle of the doc where the request originated
-root_handle: ?*DocumentStore.Handle = undefined,
+root_handle: ?*DocumentStore.Handle,
 
 const NodeSet = std.HashMapUnmanaged(NodeWithUri, void, NodeWithUri.Context, std.hash_map.default_max_load_percentage);
 
@@ -36,6 +37,7 @@ pub fn init(gpa: std.mem.Allocator, store: *DocumentStore, ip: *InternPool, root
         .arena = std.heap.ArenaAllocator.init(gpa),
         .store = store,
         .ip = ip,
+        .collect_callsite_references = true,
         .root_handle = root_handle,
     };
 }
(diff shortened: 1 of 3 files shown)


nullptrdevs:
Impressive ! `53542ms` with a non-llvm debug build .

But what if the file is named [everything.zig](https://github.com/marlersoft/zigwin32/blob/main/win32/everything.zig)?

Techatrix:
> But what if the file is named [everything.zig](https://github.com/marlersoft/zigwin32/blob/main/win32/everything.zig)?

The win32 files are quite different from files that generated from translate-c. There are zero anytype parameters in everything.zig so callsite references are never computed.

nullptrdevs:
Yeah, I concede. Was simply entertaining the idea of having limits based on items collected or time elapsed(yeah, host cpu dependent..), as a safeguard at least until `cancelRequest` is implemented.

src/analysis.zig
```diff
@@ -3051,14 +3069,21 @@ pub const DeclWithHandle = struct {
...
-                            // maybe it'd be a perf issue and this is fine?
-                            // you figure it out future contributor <3
-                            .node = call.ast.params[real_param_idx],
-                            .handle = handle,
-                        }) orelse continue;
+                        const ty = blk: {
+                            // don't resolve callsite references while resolving callsite references
+                            const old_collect_callsite_references = analyser.collect_callsite_references;
+                            defer analyser.collect_callsite_references = false;
+                            analyser.collect_callsite_references = old_collect_callsite_references;
```
nullptrdevs, line 3076: :eyes:

Techatrix: 🤦

nullptrdevs (approved):

Issue #1629, opened 2023-11-25
module in build.zig is not considered as a BuildConfig packages


evopen:
### Zig Version

0.12.0-dev.1684+ea4a07701

### Zig Language Server Version

38a199813a1018305d4e85233950c2e6c8ca30a1

### Steps to Reproduce

I'm using nested structure in a monorepo which contains multiple Zig packages (multiple build.zig.zon and build.zig) and there is a build.zig in the root directory. 
A package A in the subdirectory contains a test target with a associated test step and a bare module without a step. Another package B reference A in its zon file and build.zig. Zig builds package B fine.

First I thought I was the build.zig discovery mechanism introduced in #688. The root build.zig does not reference package A. Package A's build.zig does but the build runner fails to find them, because
1. The package is referenced by the `addTest` of package A which is named `root` in build runner and `uriAssociatedWithBuild` consider uri unmatched.
2. A bare module added with `b.addModule` is not deemed as a package in build runner.

ZLS in this case choose the root build.zig as the associated build file for package A and everything falls apart.

A simple workaround in my case would be to change root build.zig to a normal zig file and run with `zig run`.

But the root cause is the module handling I think. Did I miss something?

### Expected Behavior

ZLS finds the correct module when parsing package B.

### Actual Behavior

ZLS cannot parse the `@import` statement resulting in an unknown type.

nullptrdevs:
Hi,

Is there a repository I can checkout? I'd like to see the build.zig files and where the expected `@import` is.

For reference (not saying it's the same) :
here's a package that has no steps and exports a module
https://github.com/ziglibs/known-folders/blob/master/build.zig
zls is able to resolve the import
https://github.com/zigtools/zls/blob/35351a6ce059b6b0213b98c84793a94685f25441/src/Server.zig#L20

evopen:
Hi, I've created a simple reproduce repo. https://github.com/evopen/zls-issue-1629.
The gist is, in `/myapp/src/main.zig`,  `const mylib` is unknown type. 
If you mv `/build.zig` to `/meow.zig`, ZLS resolve `mylib` in `myapp` correctly. 

My thought is 
1. ZLS did not recognize `/mylib/build.zig` as a valid build.zig to `mylib` module, because the module is named `root` in the build runner which doesn't match the name `mylib`.
2. And ZLS resolve `myapp`'s dependency `mylib` to use `/build.zig`, because build.zig high in filesystem tree takes precedence.

[...]

PR #1647, opened 2023-12-03
Delay finding and associating a build file with a (handle) uri

nullptrdevs:
Current logic requires the modules available in a build.zig file to be known in order to correctly associate the file with an uri.

See #1629 

Take N2

### DIFF
--- a/src/DocumentStore.zig
+++ b/src/DocumentStore.zig
@@ -35,6 +35,11 @@ pub fn computeHash(bytes: []const u8) Hash {
     return hash;
 }
 
+pub const RwLock = if (builtin.single_threaded)
+    std.Thread.RwLock.SingleThreadedRwLock
+else
+    std.Thread.RwLock.DefaultRwLock;
+
 pub const BuildFile = struct {
     uri: Uri,
     /// this build file may have an explicitly specified path to builtin.zig
@@ -42,7 +47,7 @@ pub const BuildFile = struct {
     /// config options extracted from zls.build.json
     build_associated_config: ?std.json.Parsed(BuildAssociatedConfig) = null,
     impl: struct {
-        mutex: std.Thread.Mutex = .{},
+        lock: RwLock = .{},
         /// contains information extracted from running build.zig with a custom build runner
         /// e.g. include paths & packages
         /// TODO this field should not be nullable, callsites should await the build config to be resolved
@@ -50,16 +55,25 @@ pub const BuildFile = struct {
         config: ?std.json.Parsed(BuildConfig) = null,
     } = .{},
 
-    pub fn tryLockConfig(self: *BuildFile) ?BuildConfig {
-        self.impl.mutex.lock();
-        return if (self.impl.config) |cfg| cfg.value else {
-            self.impl.mutex.unlock();
-            return null;
-        };
+    pub fn hasConfig(self: *BuildFile) bool {
+        if (!self.impl.lock.tryLockShared()) return false;
+        defer self.impl.lock.unlockShared();
+        return self.impl.config != null;
     }
 
-    pub fn unlockConfig(self: *BuildFile) void {
-        self.impl.mutex.unlock();
+    /// Non-blocking - tries to obtain a shared lock,
+    /// holds a 'shared' lock on success
+    /// release with `build_file.unlockShared();`
+    pub fn getConfig(self: *BuildFile) ?BuildConfig {
+        if (!self.impl.lock.tryLockShared()) return null;
+        if (self.impl.config) |cfg| return cfg.value;
+        self.impl.lock.unlockShared();
+        return null;
+    }
+
+    /// Releases a 'shared' lock
+    pub fn unlockShared(self: *BuildFile) void {
+        self.impl.lock.unlockShared();
     }
 
     /// Usage example:
@@ -79,8 +93,8 @@ pub const BuildFile = struct {
         const tracy_zone = tracy.trace(@src());
         defer tracy_zone.end();
 
-        const build_config = self.tryLockConfig() orelse return false;
-        defer self.unlockConfig();
+        const build_config = self.getConfig() orelse return false;
+        defer self.unlockShared();
 
         try package_uris.ensureUnusedCapacity(allocator, build_config.packages.len);
         for (build_config.packages) |package| {
@@ -106,8 +120,8 @@ pub const BuildFile = struct {
         const tracy_zone = tracy.trace(@src());
         defer tracy_zone.end();
 
-        const build_config = self.tryLockConfig() orelse return false;
-        defer self.unlockConfig();
+        const build_config = self.getConfig() orelse return false;
+        defer self.unlockShared();
 
         try include_paths.ensureUnusedCapacity(allocator, build_config.include_dirs.len);
         for (build_config.include_dirs) |include_path| {
@@ -129,8 +143,8 @@ pub const BuildFile = struct {
         const tracy_zone = tracy.trace(@src());
         defer tracy_zone.end();
 
-        self.impl.mutex.lock();
-        defer self.impl.mutex.unlock();
+        self.impl.lock.lock();
+        defer self.impl.lock.unlock();
 
         if (self.impl.config) |*old_config| {
             old_config.deinit();
@@ -157,10 +171,6 @@ pub const Handle = struct {
     /// error messages from comptime_interpreter or astgen_analyser
     analysis_errors: std.ArrayListUnmanaged(ErrorMessage) = .{},
 
-    /// `DocumentStore.build_files` is guaranteed to contain this uri
-    /// uri memory managed by its build_file
-    associated_build_file: ?Uri = null,
-
     /// private field
     impl: struct {
         /// @bitCast from/to `Status`
@@ -168,12 +178,26 @@ pub const Handle = struct {
         /// TODO can we avoid storing one allocator per Handle?
         allocator: std.mem.Allocator,
 
-        lock: std.Thread.Mutex = .{},
-        condition: std.Thread.Condition = .{},
+        lock: RwLock = .{},
 
         document_scope: DocumentScope = undefined,
         zir: Zir = undefined,
         comptime_interpreter: *ComptimeInterpreter = undefined,
+
+        /// `DocumentStore.build_files` is guaranteed to contain this uri
+        /// uri memory managed by its build_file
+        associated_build_file: union(enum) {
+            /// The Handle has no associated build file (build.zig).
+            none,
+            /// The associated build file (build.zig) has not been resolved yet.
+            /// Stores a list of potential build file uris.
+            pending: struct {
+                potential_uris: []const Uri,
+                last_index_checked: usize,
+            },
+            /// The associated build file (build.zig) has been successfully resolved.
+            uri: Uri,
+        } = .none,
     },
 
     const Status = packed struct(u32) {
@@ -181,20 +205,15 @@ pub const Handle = struct {
         /// `false` indicates the document only exists because it is a dependency of another document
         /// or has been closed with `textDocument/didClose` and is awaiting cleanup through `garbageCollection`
         open: bool = false,
-        /// true if a thread has acquired the permission to compute the `DocumentScope`
-        /// all other threads will wait until the given thread has computed the `DocumentScope` before reading it.
-        has_document_scope_lock: bool = false,
         /// true if `handle.impl.document_scope` has been set
         has_document_scope: bool = false,
-        /// true if a thread has acquired the permission to compute the `ZIR`
-        has_zir_lock: bool = false,
         /// all other threads will wait until the given thread has computed the `ZIR` before reading it.
         /// true if `handle.impl.zir` has been set
         has_zir: bool = false,
         zir_outdated: bool = undefined,
         /// true if `handle.impl.comptime_interpreter` has been set
         has_comptime_interpreter: bool = false,
-        _: u25 = undefined,
+        _: u27 = undefined,
     };
 
     pub const ZirStatus = enum {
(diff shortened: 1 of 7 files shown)


src/DocumentStore.zig
```diff
@@ -1087,6 +1083,24 @@ fn createDocument(self: *DocumentStore, uri: Uri, text: [:0]const u8, open: bool
...
+        });
+    }
+
+    return handle;
+}
+
+pub fn findBuildFileForUri(self: *DocumentStore, uri: Uri) error{OutOfMemory}!void {
+    const handle = self.getHandle(uri) orelse return;
     if (isBuildFile(handle.uri) and !isInStd(handle.uri)) {
         _ = self.getOrLoadBuildFile(handle.uri);
```
Techatrix, line 1105: This is unrelated to finding the `associated_build_file` so why does this happen here?

src/DocumentStore.zig
```diff
@@ -1087,6 +1083,24 @@ fn createDocument(self: *DocumentStore, uri: Uri, text: [:0]const u8, open: bool
...
+        const server = @fieldParentPtr(Server, "document_store", self);
+
+        server.job_queue_lock.lock();
+        defer server.job_queue_lock.unlock();
+
+        try server.job_queue.ensureUnusedCapacity(1);
+        server.job_queue.writeItemAssumeCapacity(.{
+            .find_build_file_for_uri = try server.allocator.dupe(u8, uri),
+        });
+    }
```
Techatrix, line 1097: You have also delayed the collection imports and cimports which can cause wrong results until `findBuildFileForUri` is done. 

The code that prints the `Opened document {s} with build file {s}` message should also be removed because there is no guarantee that at that point in time, the associated build file has been found.

src/DocumentStore.zig
```diff
@@ -1087,6 +1083,24 @@ fn createDocument(self: *DocumentStore, uri: Uri, text: [:0]const u8, open: bool
...
+
+    return handle;
+}
+
+pub fn findBuildFileForUri(self: *DocumentStore, uri: Uri) error{OutOfMemory}!void {
```
Techatrix, line 1102: This function potentially writes to `associated_build_file`, `import_uris` and `cimports` of `Handle` while other threads may read from them -> Data Race

Techatrix (changes requested):

Techatrix:
I just realized that you had converted your PR into a Draft, sorry for doing the review too early.

nullptrdevs:
> I just realized that you had converted your PR into a Draft, sorry for doing the review too early.

It's actually appreciated — thank you.

I'm scrapping (the approach on) this one and could use ideas on how to solve it.

Haven't tested but I thought about refactoring out the find build file logic into a fn (similar to the one in this PR) and if there are no matches here
https://github.com/zigtools/zls/blob/b3285004f11b0a5f632a778685dcc8c1fcf50cf9/src/DocumentStore.zig#L1420-L1424
call it again? Con: extra file system operations

or 

Keep a `handle.potential_build_zig_files: []Uri` and check those.

src/DocumentStore.zig
```diff
@@ -50,6 +50,12 @@ pub const BuildFile = struct {
         config: ?std.json.Parsed(BuildConfig) = null,
     } = .{},
 
+    pub fn hasConfig(self: *BuildFile) bool {
+        self.impl.mutex.lock();
+        defer self.impl.mutex.unlock();
+        return if (self.impl.config) |_| true else false;
```
Techatrix, line 56:
```suggestion
        return self.impl.config != null;
```

src/DocumentStore.zig
```diff
@@ -588,6 +600,32 @@ fn getOrLoadBuildFile(self: *DocumentStore, uri: Uri) ?*BuildFile {
...
+            handle.associated_build_file = handle.potential_build_file_uris.items[0];
+            if (count == 1) break :blk;
+            const ok: bool = ok: for (handle.potential_build_file_uris.items[1..], 1..) |build_file_uri, index| {
+                const build_file = self.getOrLoadBuildFile(build_file_uri) orelse break :ok false;
+                if (!build_file.hasConfig()) break :ok false;
+                if (try self.uriAssociatedWithBuild(build_file, handle.uri)) {
+                    handle.associated_build_file = handle.potential_build_file_uris.items[index];
+                    break :ok true;
+                }
+            } else unreachable;
```
Techatrix, line 619: Why is the else `unreachable`?

src/DocumentStore.zig
```diff
@@ -588,6 +600,32 @@ fn getOrLoadBuildFile(self: *DocumentStore, uri: Uri) ?*BuildFile {
...
+pub fn doubleCheckPotentialBuildFileUri(self: *DocumentStore, handle: *Handle) error{OutOfMemory}!?Uri {
+    handle.impl.lock.lock();
+    defer handle.impl.lock.unlock();
+    switch (handle.potential_build_file_uris.items.len) {
+        0 => {},
+        else => |count| blk: {
+            _ = self.getOrLoadBuildFile(handle.potential_build_file_uris.items[0]) orelse break :blk;
+            handle.associated_build_file = handle.potential_build_file_uris.items[0];
+            if (count == 1) break :blk;
+            const ok: bool = ok: for (handle.potential_build_file_uris.items[1..], 1..) |build_file_uri, index| {
```
Techatrix, line 612: How about have a loop with the following structure?
`handle.associated_build_file = for (handle.potential_build_file_uris.items) { ... };`

You can have a special check before for `handle.potential_build_file_uris.items.len == 1`. This should avoid having the boolean to reset the associated_build_file after the loop.

src/DocumentStore.zig
```diff
@@ -588,6 +600,32 @@ fn getOrLoadBuildFile(self: *DocumentStore, uri: Uri) ?*BuildFile {
     return gop.value_ptr.*;
 }
 
+pub fn doubleCheckPotentialBuildFileUri(self: *DocumentStore, handle: *Handle) error{OutOfMemory}!?Uri {
```
Techatrix, line 603: If we have a handle with multiple potential build files but none of them are associated to the handle, then this functions returns `null`. Every subsequent call will then do its entire work again and then still return `null`. There should be some kind of caching. This should also help with reducing contention on the handle mutex.

Techatrix (changes requested):

src/DocumentStore.zig
```diff
@@ -1063,7 +1101,7 @@ fn uriInImports(
     };
     gop.key_ptr.* = handle.uri;
 
-    if (handle.associated_build_file) |associated_build_file_uri| {
+    if (handle.associated_build_file orelse try self.doubleCheckPotentialBuildFileUri(handle)) |associated_build_file_uri| {
```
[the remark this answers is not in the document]
nullptrdevs: Within that fn, is a `return self.associated_build_file;` guaranteed to make a copy or should the lock be held outside?

src/DocumentStore.zig
nullptrdevs: I'd say the issue is when there are none, otherwise if they all've been ran the first one sticks.

I'm thinking maybe passing an optional handle uri to getOrLoadBuildFile to do the association after the file has been ran (for every build file other than the first one). It'd be easier to pass a *Handle, but having an uri and calling getHandle on it ensures the handle is still valid.

src/DocumentStore.zig
nullptrdevs: The `for` is guaranteed at least one entry/iteration.

Techatrix: How is the number of iterations related to whether or the not the else path is reached?

You code is equivalent to:
```zig
for (handle.potential_build_file_uris.items[1..], 1..) |build_file_uri, index| {
    const build_file = self.getOrLoadBuildFile(build_file_uri) orelse break :ok false;
    if (!build_file.hasConfig()) break :ok false;
    if (try self.uriAssociatedWithBuild(build_file, handle.uri)) {
        handle.associated_build_file = handle.potential_build_file_uris.items[index];
        break :ok true;
    }
    // if this code path here is reached on the last iteration the you hit unreachable.
    // I don't see why it shouldn't be possible to reach on the last iteration.
    // The handle may not be associated with any of the potential build files.
} else unreachable
```

src/DocumentStore.zig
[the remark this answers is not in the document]
Techatrix: Within the function, you are holding a lock so the read from `self.associated_build_file` is protected but outside the function, you are not holding a lock which makes accessing `self.associated_build_file` a potential data race.
In short, always hold the lock of the handle when accessing `self.associated_build_file`.

Does that make it more clear what the problem is?

src/DocumentStore.zig
Techatrix: Isn't it possible to return null when searching through multiple potential build files because of the `if (!ok) handle.associated_build_file = null;` line? If `!ok` then we return `null` and any subsequent call would also return `null`.

src/DocumentStore.zig
nullptrdevs: TIL; I thought it'd be only if the condition failed.

src/DocumentStore.zig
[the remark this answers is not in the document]
nullptrdevs: I get the data race, my question is if  the `handle.getAssociatedBuildFile(store)` is
```zig
// lock defer unlock here
return self.associated_build_file;
```
is it guaranteed to make a copy which can be safely read outside the fn or should it be similar to`BuildFile.tryLockConfig`

src/DocumentStore.zig
nullptrdevs: Equipped with the knowledge of how `for() {} else` works I think correcting the else to `else true;` should handle this condition;

nullptrdevs:
Did the 'pass an optional handle uri to getOrLoadBuildFile to do the association after the file has been ran', but it had a blind spot where if you had a 'build.zig' file open it would be created without the "callback", would get loaded, but have no config when checked via the "callback", so a recheck in `handle.getAssociatedBuildFileUri` was needed.

Should we keep the "callback"  complexity or have the "on demand" only logic?

Techatrix (changes requested):
> Should we keep the "callback" complexity or have the "on demand" only logic?

I would consider your current "callback" approach to be acceptable (but a bit complex). I you want to do the "on demand" approach then this implementation of `getAssociatedBuildFileUri` could be feasible.

Keep in mind that I have not tested this code and that TODO should definitely be resolved.
```zig
pub fn getAssociatedBuildFileUri(self: *Handle, store: *DocumentStore) ?Uri {
    self.impl.lock.lock();
    defer self.impl.lock.unlock();
    switch(associated_build_file) {
        .none => return null,
        .unresolved => |potential_build_file_uris| {
            // should there be special logic if `potential_build_file_uris.len == 1`?
            
            var has_missing_build_config = false;
            for (potential_build_file_uris) |build_file_uri| {
                const build_file = self.getOrLoadBuildFile(build_file_uri, handle.uri) orelse continue; // continue or break?
                if (!build_file.hasConfig()) {
                    has_missing_build_config = true;
                    continue;
                }
                if (!try self.uriAssociatedWithBuild(build_file, handle.uri)) {
                    // TODO
                    // the given `build_file_uri` should be removed from `potential_build_file_uris`
                    // so that future calls don't redo all this work on every call.
                    continue;
                }
                /// the associated build file has been resolved so there is no need to store the potential build files.
                handle.potential_build_file_uris.clearAndFree(handle.allocator);
    
                associateBuildFileUri = .{ .resolved = build_file.uri };
                return build_file.uri;
            };
    
            if (!has_missing_build_config) {
                /// there is no associated build file so there is no need to store the potential build files.
                handle.potential_build_file_uris.clearAndFree(handle.allocator);
                associated_build_file = .none;
            }
            // when build configs are missing we keep the state at .unresolved so that
            // future calls will retry until all build config are resolved.
            // Then will have a conclusive result on whether or not there is a associated build file.
            return null;
        },
        .resolved => |uri| return uri,
    }
}
```
```zig
associated_build_file: union(enum) {
    /// The Handle has no associated build file (build.zig).
    none,
    /// The associated build file (build.zig) has not been resolved yet.
    /// Stores a list of potential build files.
    unresolved: []const Uri,
    /// The associated build file (build.zig) has been successfully resolved.
    resolved: Uri,
} = .unresolved,
```

src/DocumentStore.zig
```diff
@@ -1072,10 +1124,6 @@ fn uriInImports(
     };
     gop.key_ptr.* = handle.uri;
 
-    if (handle.associated_build_file) |associated_build_file_uri| {
-        return std.mem.eql(u8, associated_build_file_uri, build_file_uri);
-    }
```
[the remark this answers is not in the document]
nullptrdevs: Noice - a clear objective! ..also, I put those back in :sweat_smile:, which restored same timings as `main`.
..but it was still slow (on my machine/fs), ..so I took it further :zany_face:

src/DocumentStore.zig
```diff
@@ -360,6 +363,26 @@ pub const Handle = struct {
         return @bitCast(self.impl.status.load(.Acquire));
     }
 
+    /// Non-blocking - tries to obtain a shared lock,
+    /// holds a `shared` lock on success
+    /// Release with `handle.unlockShared();`
+    pub fn getAssociatedBuildFileUri(self: *Handle, store: *DocumentStore, associate_if_pending: bool) ?Uri {
+        if (!self.impl.lock.tryLockShared()) return null;
```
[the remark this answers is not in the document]
nullptrdevs: Most of the calls are for completions, so I didn't consider it necessary.

src/DocumentStore.zig
```diff
@@ -483,9 +509,10 @@ allocator: std.mem.Allocator,
 config: *const Config,
 /// the DocumentStore assumes that `runtime_zig_version` is not modified while calling one of its functions.
 runtime_zig_version: *const ?ZigVersionWrapper,
-lock: std.Thread.RwLock = .{},
+lock: RwLock = .{},
 handles: std.StringArrayHashMapUnmanaged(*Handle) = .{},
 build_files: std.StringArrayHashMapUnmanaged(*BuildFile) = .{},
+build_files_lock: RwLock = .{},
```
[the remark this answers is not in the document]
nullptrdevs: I favor granularity, but can't think of a scenario where contention will be too high, .. still

src/DocumentStore.zig
```diff
@@ -594,6 +622,23 @@ fn getOrLoadBuildFile(self: *DocumentStore, uri: Uri) ?*BuildFile {
...
 
+pub fn associateBuildFileUri(self: *DocumentStore, handle: *Handle) error{OutOfMemory}!void {
+    handle.impl.lock.lock();
+    defer handle.impl.lock.unlock();
+
+    if (handle.impl.associated_build_file != .pending) return;
+
+    handle.impl.associated_build_file = blk: for (handle.potential_build_file_uris.items, 0..) |build_file_uri, index| {
+        const build_file = self.getOrLoadBuildFile(build_file_uri, handle.uri) orelse break :blk .pending;
+        if (!build_file.hasConfig()) break :blk .pending;
```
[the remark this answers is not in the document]
nullptrdevs: But then it would reach the `else => .none`, unless I come around and implement your suggestion

src/DocumentStore.zig
```diff
@@ -594,6 +622,23 @@ fn getOrLoadBuildFile(self: *DocumentStore, uri: Uri) ?*BuildFile {
...
+        if (try self.uriAssociatedWithBuild(build_file, handle.uri)) break :blk .{ .uri = handle.potential_build_file_uris.items[index] };
+    } else .none;
```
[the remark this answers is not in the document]
nullptrdevs: I'm still to implement the `TODO` but I'm thinking
```zig
pending: struct {
	potential: []const Uri,
    last_checked_index: usize,
};
```
to avoid extra allocations?

src/DocumentStore.zig
```diff
@@ -35,31 +35,45 @@ pub fn computeHash(bytes: []const u8) Hash {
...
     } = .{},
 
-    pub fn tryLockConfig(self: *BuildFile) ?BuildConfig {
-        self.impl.mutex.lock();
-        return if (self.impl.config) |cfg| cfg.value else {
-            self.impl.mutex.unlock();
-            return null;
-        };
+    pub fn hasConfig(self: *BuildFile) bool {
+        if (!self.impl.lock.tryLockShared()) return false;
```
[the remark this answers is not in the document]
nullptrdevs: I'd like to keep them - technically, they'd succeed 99.9% of the time and more an of indication that it isn't critical.

PR #1652, opened 2023-12-05
disable autofix for var never mutated

xdBronch:
the fix is still available as a code action, just manual now

### DIFF
--- a/tests/lsp_features/code_actions.zig
+++ b/tests/lsp_features/code_actions.zig
@@ -24,7 +24,7 @@ test "code actions - discard value" {
         \\    _ = bar;
         \\}
         \\
-    , &.{});
+    );
 }
 
 test "code actions - discard function parameter" {
@@ -38,7 +38,7 @@ test "code actions - discard function parameter" {
         \\    _ = c;
         \\}
         \\
-    , &.{});
+    );
 }
 
 test "code actions - discard captures" {
@@ -79,7 +79,7 @@ test "code actions - discard captures" {
         \\    };
         \\}
         \\
-    , &.{});
+    );
 }
 
 test "code actions - discard capture with comment" {
@@ -99,7 +99,7 @@ test "code actions - discard capture with comment" {
         \\    }
         \\}
         \\
-    , &.{});
+    );
 }
 
 test "code actions - remove pointless discard" {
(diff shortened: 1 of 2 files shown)


tests/lsp_features/code_actions.zig
```diff
@@ -183,6 +188,9 @@ fn testAutofixOptions(before: []const u8, after: []const u8, want_zir: bool) !vo
 
     for (response) |action| {
         const code_action = action.CodeAction;
+        for (ignore_messages) |message| {
+            if (std.mem.eql(u8, code_action.title, message)) return;
```
Techatrix, line 192: This line here should be a `continue` instead of `return` which explains why the `"code actions - correct unnecessary uses of var"` test is unexpectedly passing. Note that code action tests only test autofix actions.

This diff should do all the necessary work:
```diff
diff --git a/src/features/code_actions.zig b/src/features/code_actions.zig
index 87037fd..337501d 100644
--- a/src/features/code_actions.zig
+++ b/src/features/code_actions.zig
@@ -304,7 +304,7 @@ fn handleVariableNeverMutated(builder: *Builder, actions: *std.ArrayListUnmanage
 
     try actions.append(builder.arena, .{
         .title = "use 'const'",
-        .kind = .@"source.fixAll",
+        .kind = .quickfix,
         .isPreferred = true,
         .edit = try builder.createWorkspaceEdit(&.{
             builder.createTextEditLoc(var_keyword_loc, "const"),
diff --git a/tests/lsp_features/code_actions.zig b/tests/lsp_features/code_actions.zig
index f6c0382..8d17679 100644
--- a/tests/lsp_features/code_actions.zig
+++ b/tests/lsp_features/code_actions.zig
@@ -130,22 +130,6 @@ test "code actions - remove pointless discard" {
     );
 }
 
-test "code actions - correct unnecessary uses of var" {
-    try testAutofix(
-        \\test {
-        \\    var foo = 5;
-        \\    _ = foo;
-        \\}
-        \\
-    ,
-        \\test {
-        \\    const foo = 5;
-        \\    _ = foo;
-        \\}
-        \\
-    );
-}
-
 fn testAutofix(before: []const u8, after: []const u8) !void {
     try testAutofixOptions(before, after, true); // diagnostics come from our AstGen fork
     try testAutofixOptions(before, after, false); // diagnostics come from calling zig ast-check
```

Techatrix (changes requested):

tests/lsp_features/code_actions.zig
xdBronch: i mightve misunderstood what he meant but i spoke to auguste and he said to leave the test and do some other things for this. just removing it was my original plan

Techatrix: Yes, ideally we would want to keep the test but this would require to adjust the `testAutofix` functions to allow testing non autofix code actions where you can select which actions should be applied and detect missing actions. I personally wouldn't mind losing the test, given that we also don't have tests for other non autofix code actions.

xdBronch: i'd like to get this merged, should i just take out my changes to the testing stuff and remove the test?

SuperAuguste (approved):

PR #1654, opened 2023-12-06
code_actions: autofix unused capture in while loop with continue stat…

travisstaloch:
previously this statement wasn't autofixed
```zig
...
while (lines.next()) |line| : (linei += 1) {}
```
and and now it becomes
```zig
...
while (lines.next()) |line| : (linei += 1) {
    _ = line;
}
```

### DIFF
--- a/src/features/code_actions.zig
+++ b/src/features/code_actions.zig
@@ -225,27 +225,29 @@ fn handleUnusedCapture(
 
     // handle while loop continue statements such as `while(foo) |bar| : (x += 1) {}`
     if (source[block_start] == ':') {
-        block_start += 1;
-        var depth: usize = 1;
-        var state = enum { ws, continue_stmt, ws2 }.ws;
-        while (block_start < source.len) : (block_start += 1) {
-            const c = source[block_start];
-            switch (state) {
-                .ws => if (c == '(') {
-                    state = .continue_stmt;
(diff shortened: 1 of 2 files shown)


travisstaloch:
I'm not sure what to do in response to the code coverage report. I do see one line that I added in the report (`depth += 1;` on line 241.) but I'm not sure what to do about that.

travisstaloch:
going to try adding another test where there are nested parens in the continue statement...

Techatrix (changes requested):
searching for the closing parenthesis by looping through the source can be a bit tricky:
```zig
const std = @import("std");
test {
    var lines = std.mem.tokenizeSequence(u8, "", "\n");
    var linei: usize = 0;
    while (lines.next()) |line| : (linei += ")))".len) {}
}
```
I would suggest to use [`sourceIndexToTokenIndex`](https://github.com/zigtools/zls/blob/b45f7a8ae1c257f5298ae3356cf5cb83ff511ea7/src/offsets.zig#L71C21-L71C21) and then iterate through the tokens.
You may use the [`findMatchingRBrace`](https://github.com/zigtools/zls/blob/e2da11dbc2ad9a0dfd75fd1ae9bd495b1494d92d/src/ast.zig#L70) function for reference.

Techatrix (approved):

PR #1656, opened 2023-12-07
completions for re-exported functions

### LINKED ISSUE #423: Feature Request: Add completions for re-exported functions
A common pattern in bindings to C code goes like this:

```zig
const Handle = opaque {
    pub const setValue = lib_Handle_setValue;
    pub fn setArrayValues(self: *Handle, values: []u32) void {
        lib_Handle_setArrayValues(self, values.ptr, values.len);
    }
    
    extern fn lib_Handle_setValue(self: *Handle, val: u32) void;
    extern fn lib_Handle_setArrayValues(self: *Handle, values: ?[*]u32, len: usize) void;
};
```

Here two C functions have been bound to the `Handle` type for more convenient use.  `setArrayValues` is a proper wrapper, in order to use a slice parameter.  But `setValue` does not need any signature modification, so it has been exported directly under a new name.  Unfortunately, ZLS cannot recognize this pattern.  Autocomplete on a handle pointer will suggest `setArrayValues` but not `setValue`.  It would be very convenient if ZLS could recognize these sorts of pass-through member functions and add them to the autocomplete list.

nolanderc:
closes #423

I encountered this bug while using https://github.com/ziglibs/zgl, which has many types that look like the following:

```zig
pub const VertexArray = enum(UInt) {
    invalid = 0,
    _,

    pub const create = gl.createVertexArray;
    pub const delete = gl.deleteVertexArray;
    pub const gen = gl.genVertexArray;
    pub const bind = gl.bindVertexArray;
    pub const enableVertexAttribute = gl.enableVertexArrayAttrib;
    pub const disableVertexAttribute = gl.disableVertexArrayAttrib;

    pub const attribFormat = gl.vertexArrayAttribFormat;
    pub const attribIFormat = gl.vertexArrayAttribIFormat;
    pub const attribLFormat = gl.vertexArrayAttribLFormat;

    pub const attribBinding = gl.vertexArrayAttribBinding;

    pub const vertexBuffer = gl.vertexArrayVertexBuffer;
    pub const elementBuffer = gl.vertexArrayElementBuffer;
};
```

Previously, there were no completions for any of the re-exported functions. But with this patch we yield completions for all re-exported functions which accept the container (`VertexArray` in the case above) as their first argument.

### DIFF
--- a/src/analysis.zig
+++ b/src/analysis.zig
@@ -232,6 +232,28 @@ pub fn isInstanceCall(
 }
 
 pub fn hasSelfParam(analyser: *Analyser, handle: *DocumentStore.Handle, func: Ast.full.FnProto) error{OutOfMemory}!bool {
+    const token_starts = handle.tree.tokens.items(.start);
+    const in_container = try innermostContainer(handle, token_starts[func.ast.fn_token]);
+    return analyser.firstParamIs(handle, func, in_container);
+}
+
+pub fn isSelfFunction(analyser: *Analyser, func_type: TypeWithHandle, container: TypeWithHandle) error{OutOfMemory}!bool {
+    if (!func_type.isFunc()) return false;
+    const func_decl = try analyser.resolveFuncProtoOfCallable(func_type) orelse return false;
+    if (func_decl.type.is_type_val) return false;
+
+    var buf: [1]Ast.Node.Index = undefined;
+    const func = func_decl.handle.tree.fullFnProto(&buf, func_decl.type.data.other) orelse return false;
+
+    return analyser.firstParamIs(func_decl.handle, func, container);
+}
+
+pub fn firstParamIs(
+    analyser: *Analyser,
+    handle: *DocumentStore.Handle,
+    func: Ast.full.FnProto,
+    expected: TypeWithHandle,
+) error{OutOfMemory}!bool {
     // Non-decl prototypes cannot have a self parameter.
     if (func.name_token == null) return false;
     if (func.ast.params.len == 0) return false;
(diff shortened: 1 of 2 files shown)


tests/lsp_features/completion.zig
```diff
@@ -1672,6 +1672,24 @@ test "completion - enum completion on out of bound parameter index" {
     , &.{});
 }
 
+test "completion - delegated declaration" {
```
Techatrix, line 1675: Could you move this test case to some of the other tests that deal with functions like `completion - function call`

Techatrix (changes requested):

src/analysis.zig
```diff
@@ -3236,7 +3259,18 @@ fn iterateSymbolsContainerInternal(
...
+                        // allow declarations which evaluate to functions where
+                        // the first parameter has the type of the container:
+                        const alias = try analyser.resolveVarDeclAlias(
+                            NodeWithHandle{ .node = node, .handle = handle },
+                        ) orelse continue;
+
+                        const alias_type = try alias.resolveType(analyser) orelse continue;
+
+                        const container_type = TypeWithHandle.typeVal(container_handle);
+                        if (!try analyser.isSelfFunction(alias_type, container_type)) continue;
```
Techatrix, line 3272: I didn't notice this previously, but `resolveVarDeclAlias` will only resolve things like imports, symbol lookups and field accesses but the expression could be more complex than that which can be handled by `resolveTypeOfNode`.
```suggestion
                        const alias = try analyser.resolveTypeOfNode(
                            NodeWithHandle{ .node = node, .handle = handle },
                        ) orelse continue;

                        const container_type = TypeWithHandle.typeVal(container_handle);
                        if (!try analyser.isSelfFunction(alias, container_type)) continue;
```

Techatrix (changes requested):

Techatrix (approved):

PR #1657, opened 2023-12-08
Disable autofix by default

### LINKED ISSUE #1623: Should autofix be disabled by default?
I've been seeing a larger number of people disliking autofix being enabled by default as it mutates their code unexpectedly.

I'm undecided.

SuperAuguste:
Closes #1623

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ The following options are currently available.
 | `enable_argument_placeholders` | `bool` | `true` | Whether to enable function argument placeholder completions |
 | `enable_ast_check_diagnostics` | `bool` | `true` | Whether to enable ast-check diagnostics |
 | `enable_build_on_save` | `bool` | `false` | Whether to enable build-on-save diagnostics |
-| `enable_autofix` | `bool` | `true` | Whether to automatically fix errors on save. Currently supports adding and removing discards. |
+| `enable_autofix` | `bool` | `false` | Whether to automatically fix errors on save. Currently supports adding and removing discards. |
 | `semantic_tokens` | `enum` | `.full` | Set level of semantic tokens. Partial only includes information that requires semantic analysis. |
 | `enable_inlay_hints` | `bool` | `true` | Enables inlay hint support when the client also supports it |
 | `inlay_hints_show_variable_type_hints` | `bool` | `true` | Enable inlay hints for variable types |
--- a/schema.json
+++ b/schema.json
@@ -27,7 +27,7 @@
         "enable_autofix": {
             "description": "Whether to automatically fix errors on save. Currently supports adding and removing discards.",
             "type": "boolean",
-            "default": "true"
+            "default": "false"
         },
         "semantic_tokens": {
             "description": "Set level of semantic tokens. Partial only includes information that requires semantic analysis.",
--- a/src/Config.zig
+++ b/src/Config.zig
@@ -17,7 +17,7 @@ enable_ast_check_diagnostics: bool = true,
 enable_build_on_save: bool = false,
 
 /// Whether to automatically fix errors on save. Currently supports adding and removing discards.
-enable_autofix: bool = true,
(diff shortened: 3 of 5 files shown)


xdBronch:
i feel that most of the dislike of autofix was because of the var -> const one (which is already disabled now). i dont think ive seen many if any people complain about the discarding unused variables which ime helps a lot of people get over the unused variable error

SuperAuguste:
I've seen people complain about that unfortunately. I've just had enough of those comments so I'm going to merge this.

Techatrix:
> I've just had enough of those comments so I'm going to merge this.

Have you tried ignoring people that have a passion for complaining? 	=^_^=

SuperAuguste:
It's hard to ignore them when even your friends start saying the same stuff (albeit in a much nicer way). And they do have a point: editing people's code by default is a little jarring, I've experienced (and hated) this myself with the Go LSP, so I do get where they're coming from.

Techatrix:
> And they do have a point: editing people's code by default is a little jarring

That is a valid argument. That is something I would have liked to discuss in #1623.

Let me make a counterargument:
Autofix is not the only that is editing people's code.
The vscode-zig has format on save enabled by default which beyond formatting, upgrades your code by applying fixups.
Of course, autofix and fixups are not the same, but they both editing people's code without permission, so is the editing of people's code that is the problem or is there an underlying issue? Perhaps they perceive the actions of autofix to be just undesired, non-self-explaining or just confusing?

GrayHatter:
> Let me make a counterargument:
> Autofix is not the only that is editing people's code. The [code that's not zls] has [...]

Just because other projects are doing something is hardly related to if zls should or shouldn't do it. Useful to study and figure out sure. But definitely not something to emulate! Good software will default having fewer opinions whenever possible. Formatting/style is opinionated by definition, and IIRC Zig formatting used to be required for the compiler to function. Consistent formatting is useful for source control features, and, the point of installing tooling. If I wanted to format my code myself, I wouldn't install a code formatter. But I didn't install ZLS because I wanted it to fix or format my code, I installed it to get better linting hints, and better syntax highlighting. I never wanted it to change my code.

The problem for me isn't which fixup's it's making, the problem is that it's making decisions about my code without asking first. I think autofix is a great feature, and one zls should have. But it's not one I'd ever use, never for any of my code. It's not just jarring, it also breaks how I interact with code, unused variable warnings are compile errors for a reason, and I appreciate them! The author needs to make a decision, use the variable, or intentionally suppress it. Having my LSP do it for me, especially when I didn't want it, and when I didn't ask for it, is problematic. If autofix is happy to suppress these, errors why wouldn't it add `@constCast` for me? 

Autofix shouldn't be enabled by default, because I don't trust autofix to act reasonably, if I want to take on the risk that something changed my code and I missed a bug because of it, that's my decision to make, but I should have to make the decision to enable it. 

I'm currently compiling from source myself because this is important to me, it's also mentioned as a warning every time I mention or suggest zls. 

Unsolicited rant aside, I only found out about this pull because I noticed the branch was added when I pulled to check for updates. So I just came by to say thank you @SuperAuguste, I really appreciate this change!

nullptrdevs (approved):

Narsil:
@SuperAuguste 

Random comment of sympathy about the autofix feature.

Having it enabled by default was super nice for me as a beginner (when iterating on code really quick it's quite annoying when something doesn't build because I quickly uncommented part of the logic).

It's ok if it's not the default (I can sort of get the rationale), I enabled it for me since I highly value the speedup to the devtime when changing stuff around a lot.

Thanks a lot for the feature, and thanks for making it the default for a bit, I wouldn't have discovered it without it, and it definitely eases the pain of zig complaining about unused variables (which again, I appreciate, but it's annoying when you're not trying to clean up your code, just iterating fast on some parts of your code while investigating something)

PR #1594, opened 2023-11-10
another intern pool rewrite

Techatrix:
some work I did on the InternPool from my sema branch.

I decided to do some more changes.
- [x] don't store pointers in the InternPool
- [x] degibberish function for types (bonus)
- [x] make the InternPool thread safe
- [x] remove/replace encoding.zig
~- [ ] improve storage of container types~ nah, I will do this another time

I am gonna have so much fun, rebasing my sema branch. 	(︶︹︶)

PR #1663, opened 2023-12-10
Fix type resolution of various primitives

### LINKED ISSUE #1600: types of boolean literals are resolved as their literal value instead of `bool`
### Zig Version

0.12.0-dev.1591+3fc6a2f11

0.12.0-dev.228+aa9fc0c

### Steps to Reproduce

image

i believe this is somewhat recent, i dont remember it happening before

### Expected Behavior

both should be resolved as `bool` type

### Actual Behavior

resolved as `true` and `false`

FnControlOption:
```zig
const a = false; // bool
const b = true; // bool
const c = null; // @TypeOf(null)
const d = undefined; // @TypeOf(undefined)
const e: i7 = 42; // i7
const f: u7 = 42; // u7
```

- Added `.int` variant to `Type.data` for integer types not covered by `InternPool.Index`
- `primitives.get("true")` returns `.bool_true` but the type of `true` should be `.bool_type`. Let's use the return value of `InternPool.Key.typeOf` if the primitive is a value and not a type. (Closes #1600)

@Techatrix: What are the following lines for? https://github.com/zigtools/zls/blob/a4c568d/src/analysis.zig#L4071-L4079 At first glance, they seem redundant since `addReferencedTypes` already handles `.ip_index`

### DIFF
--- a/src/analysis.zig
+++ b/src/analysis.zig
@@ -977,19 +977,35 @@ const primitives = std.ComptimeStringMap(InternPool.Index, .{
     .{ "void", .void_type },
 });
 
-pub fn resolvePrimitiveType(identifier_name: []const u8) ?InternPool.Index {
-    if (primitives.get(identifier_name)) |primitive| return primitive;
+pub fn resolvePrimitiveType(analyser: *Analyser, handle: *DocumentStore.Handle, identifier_name: []const u8) ?TypeWithHandle {
+    if (primitives.get(identifier_name)) |primitive| {
+        const primitive_type = analyser.ip.indexToKey(primitive).typeOf();
+        const is_type = primitive_type == .type_type;
+        const ip_index = if (is_type) primitive else primitive_type;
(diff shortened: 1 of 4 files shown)


xdBronch:
closes #1600

Techatrix (commented):
The InternPool is capable of representing integers of any bit size. It just need to mutate the InternPool which is not thread safe yet.
I have outlined how I envisioned the fix to #1600 in #1589 which is different from what is being done in this PR.
If this were to be approved/merged, I would have to revert most of these changes after I finish #1594.

> What are the following lines for?

I've always carried over this special handling for ip indicies. At some point, `addReferencedTypes` started supporting ip indices itself but I've never removed those lines from `referencedTypes`. Tests appear to pass without them so doesn't look like its needed anymore.

Techatrix:
Most of these fixes have been implemented by #1684.

PR #1666, opened 2023-12-12
handle `offsets.positionToIndex` on more characters than line length

Techatrix:
The LSP specification says the following about the `character` field of `Position`:
> If the character value is greater than the line length it defaults back to the line length.


nullptrdevs (approved):
It does say it.

PR #1668, opened 2023-12-12
give filesystem completions on string literals in build files

Techatrix:
Many string literals in `build.zig` files refer to file system paths so we can optimistically provide completions there. If the string literal is not a valid file path then no completions will appear.


nullptrdevs (approved):
:rocket:

PR #1669, opened 2023-12-12
type erase `iterateChildren` and `iterateChildrenTypeErased`

Techatrix:
This reduced binary size by about 32kb when building ReleaseSmall. Not really sure if this small improvement is worth it but I guess why not.


nullptrdevs (approved):
Worth

PR #1672, opened 2023-12-12
Surface build runner errors

SuperAuguste:
Adds a cute lil popup message + a showDocument containing the error log! I wrote this late but I think the impl is sound and the UX is *fine*.

### DIFF
--- a/src/Server.zig
+++ b/src/Server.zig
@@ -19,6 +19,7 @@ const ZigVersionWrapper = @import("ZigVersionWrapper.zig");
 const Transport = @import("Transport.zig");
 const known_folders = @import("known-folders");
 const BuildRunnerVersion = @import("build_runner/BuildRunnerVersion.zig").BuildRunnerVersion;
+const URI = @import("uri.zig");
 
 const signature_help = @import("features/signature_help.zig");
 const references = @import("features/references.zig");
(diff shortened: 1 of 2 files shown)


src/Server.zig
```diff
@@ -755,6 +758,52 @@ fn handleConfiguration(server: *Server, json: std.json.Value) error{OutOfMemory}
...
+    if (res.value != null) {
+        const request_id = try std.fmt.allocPrint(
+            server.allocator,
+            "show_document-{d}",
+            .{millis},
+        );
+
+        const cache = std.fs.cwd().makeOpenPath("zig-cache", .{}) catch return;
+        cache.writeFile2(.{
+            .sub_path = "build_runner_failure.txt",
```
SuperAuguste, line 780: Thus hurts me lol; iirc `rust-analyzer` does this via per-editor integration which means they can use virtual documents rather than real ones. I put this in `zig-cache` rather than `zig-cache/tmp` because VSCode opens the target directory if it's in-workspace. The system tmp directory could be more fruitful, but I didn't have the time to implement that. Lemme know and I'll happily do it :)

nullptrdevs (approved):
Feels sledge hammer-ish, but beats none and we all know it is needed.

vscode :white_check_mark: 
nvim :white_check_mark: 
Kate :x:

PR #1684, opened 2023-12-24
fix analysis issues related to primitive types

### LINKED ISSUE #1600: types of boolean literals are resolved as their literal value instead of `bool`

Techatrix:
I've outlined how I use the InternPool to store types and values here: https://github.com/zigtools/zls/pull/1589#issuecomment-1802450090

fixes #1600


nullptrdevs (approved):
Nice

PR #1696, opened 2024-01-02
Added config setting to toggle display of function definition in label details

diocletiann:
Helps with visibility of return types in some editors (sublime, vscode). `self` check is not ideal but it will work for the vast majority of methods and fallback to `(...)`.

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,7 @@ The following options are currently available.
 | `global_cache_path` | `?[]const u8` | `null` | Path to a directory that will be used as zig's cache. null is equivalent to `${KnownFolders.Cache}/zls` |
 | `build_runner_global_cache_path` | `?[]const u8` | `null` | Path to a directory that will be used as the global cache path when executing a projects build.zig. null is equivalent to the path shown by `zig env` |
 | `completions_with_replace` | `bool` | `true` | Completions confirm behavior. If 'true', replace the text after the cursor |
+| `completion_label_details` | `bool` | `true` | When false, the function signature of completion results is hidden. Improves readability in some editors |
 <!-- DO NOT EDIT -->
 
 ### Per-build Configuration Options
--- a/schema.json
+++ b/schema.json
@@ -148,6 +148,11 @@
             "description": "Completions confirm behavior. If 'true', replace the text after the cursor",
             "type": "boolean",
             "default": true
+        },
+        "completion_label_details": {
+            "description": "When false, the function signature of completion results is hidden. Improves readability in some editors",
+            "type": "boolean",
+            "default": true
         }
     }
 }
--- a/src/Config.zig
+++ b/src/Config.zig
@@ -92,4 +92,7 @@ build_runner_global_cache_path: ?[]const u8 = null,
 /// Completions confirm behavior. If 'true', replace the text after the cursor
 completions_with_replace: bool = true,
 
+/// When false, the function signature of completion results is hidden. Improves readability in some editors
+completion_label_details: bool = true,
+
 // DO NOT EDIT
--- a/src/analysis.zig
+++ b/src/analysis.zig
@@ -160,9 +160,14 @@ pub const FormatFunctionOptions = struct {
     /// only included if available
     include_name: bool,
     skip_first_param: bool = false,
-    include_parameter_modifiers: bool,
-    include_parameter_names: bool,
-    include_parameter_types: bool,
+    parameters: union(enum) {
+        collapse,
+        show: struct {
+            include_modifiers: bool,
+            include_names: bool,
+            include_types: bool,
+        },
+    },
     include_return_type: bool,
     snippet_placeholders: bool,
 };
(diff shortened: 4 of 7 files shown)


src/Server.zig
```diff
@@ -376,6 +376,7 @@ fn initializeHandler(server: *Server, _: std.mem.Allocator, request: types.Initi
         log.info("client is '{s}-{s}'", .{ clientInfo.name, clientInfo.version orelse "<no version>" });
 
         if (std.mem.eql(u8, clientInfo.name, "Sublime Text LSP")) blk: {
+            server.config.completion_label_details = false;
```
Techatrix, line 379: Any editor that supports label details will struggle to display long label details with a small enough editor window, not just Sublime Text.
What is the motivation for having this option be `false` only in Sublime Text?

The default value of every config option is documented. Secretly overriding the default value on a specific editor/client should be clearly documented so that the user is not being misled.

Overriding the default value on a specific editor/client also means that we are interfering with how the client chooses to display the information we provide them. Ideally, the server/ZLS should behave identical on all clients (workarounds are an exception).

diocletiann: >  What is the motivation for having this option be false only in Sublime Text?

Apologies, I had a comment in the option docstring stating it would be set to `false` in Sublime by default but I forgot to include that part when I resubmitted the PR. 

The motivation for defaulting to `false` in Sublime is due to how the completion results are rendered and the majority of return types aren't visible. I believe it will be a better experience for Sublime users. That's just my opinion, it was annoying enough to me to implement this change. That said, happy to change the PR to either a `true` default for all editors or improve the documentation to explain why `false` is the default for Sublime.

diocletiann: I also now see there is no way to override `server.config.completion_label_details = false;` in `zls.json`. My original intention was to default to `false` but allow the user to change it back to `true` if desired.

Techatrix: I do see a reason for having this disabled on Sublime Text, but the same still goes for VSCode when you use a small enough window. Though, I've observed that VSCode was better at using the limited window space.

I would suggest to change it to `true` for all editors since this option tries to resolve an issue on the editor side which 
ideally shouldn't be the responsibility of the server/ZLS to work around. Perhaps, how the ST LSP plugin (or other clients) deals with long completion label details could be brought to their attention to discuss further improvements.

diocletiann: Sounds good. I removed the line forcing `false` for Sublime Text. I've also mentioned this to ST LSP maintainers. Thanks.

rwols: From my cursory glance, is this a server setting? If so, you can set server settings via the client configuration json in ST LSP.

diocletiann: > From my cursory glance, is this a server setting? If so, you can set server settings via the client configuration json in ST LSP.

Yes this is a server setting, since it has value in vscode and at in Neovim (at least in the Lazy.nvim config I use). Good to know it can be set on the client, thanks.

diocletiann (commented):
Changes made.

diocletiann:
I like the results in neovim, at least in this configuration (Lazy.nvim).

image

Techatrix:
@diocletiann
My apologies for letting this PR unanswered until now. My recent work that refactored how the completion label detail is being generated has conflicted with your work so I have edited your PR to resolve the conflicts.
Could you verify that everything works as it did previously?

diocletiann:
> @diocletiann My apologies for keeping this PR unanswered. My recent work that refactored how the completion label detail is being generated has conflicted with your work so I have edited your PR to resolve the conflicts. Could you verify that everything works as it did previously?

No worries at all. I checked the new behavior and it has diverged a bit. My initial goal was to to ensure the function name and the return type are for as many results as possible. In the new version, the full signature is displayed after the shortened function name (instead of the return type) and for many results the return type gets cut off:

image

diocletiann:
Also, I omitted the elipsis when `self` was the only argument but in this version every member function has one:

image

diocletiann:
Return type is now missing in VS Code:

image

Techatrix:
> My initial goal was to to ensure the function name and the return type are visible for as many results as possible. In the new version, the full signature is displayed after the shortened function name (instead of the return type) and for many results the return type gets cut off:

I did not realize that the [`description`](https://github.com/zigtools/zls/blob/88a352a46ce7fbf81aa009dec6f03bd7d08af2b1/src/lsp.zig#L4776-L4777) was being set to the return type. I bit odd since the `description` should be used for fully qualified names or file path according to the LSP documentation :shrug: . This should be fixed now.

> Also, I omitted the ellipsis when `self` (hacky I know) was the only argument but in this version every member function has one:

This is because you are requesting completion on the `ArrayList` **type** and not a value of it.

diocletiann:
> This is because you are requesting completion on the `ArrayList` **type** and not a value of it. Here is an example:
Right, thanks for pointing that out.

Looks good in Sublime and VS Code, thank you!

image

image

Techatrix (approved):

diocletiann:
@Techatrix I just noticed, there seems to much more whitespace between the function name and return type than before. Not a big deal, just fyi.

Techatrix:
> I just noticed, there seems to much more whitespace between the function name and return type than before. Not a big deal, just fyi.

Could you give an example where more whitespace occurs?

diocletiann:
> > I just noticed, there seems to much more whitespace between the function name and return type than before. Not a big deal, just fyi.
> 
> Could you give an example where more whitespace occurs?

The gap between the function name and the return type:

before the refactor
image

after (same pane width, I don't have an old VSC screenshot though I'm pretty sure the gap was narrower)
image

Issue #1396, opened 2023-08-10
zls is broken in bun's repo


Jarred-Sumner:
### Zig Version

0.11.0-dev.4006+bf827d0b5

### Zig Language Server Version

HEAD (latest published to VSCode)

### Steps to Reproduce

1. `gh clone oven-sh/bun`
2. Open a .zig file which uses the `bun` identifier (almost all of them)
3. No completions appear.

Looks like zls is failing to load build.zig in bun's repository. This started happening yesterday or today. I don't think we made any relevant changes.

Any identifiers which reference an import no longer have completions
<img width="175" alt="image" src="https://github.com/zigtools/zls/assets/709451/ef82cb3f-3f3b-4bcd-ba10-31a34510a748">

Note that `bun` here is `const bun = @import("root").bun`. The `@import("root")` seems to not resolve, most likely due to failing to load build.zig.

Relevant log (note the invalid unicode character):
```js
error: (store ): Failed to execute build runner to collect build configuration, command:
info : (server): Setting configuration...
```

Complete log:
```js
info : ( main ): Starting ZLS 0.12.0-dev.2+9532f4a @ '/Users/jarred/Library/Application Support/Code/User/globalStorage/ziglang.vscode-zig/zls_install/zls'
info : (config): Using zig executable '/Users/jarred/zig/master/files/zig'
info : (config): Using zig lib path '/Users/jarred/zig/0.11.0-dev.4006+bf827d0b5/files/lib'
info : (config): Using build runner global cache path '/Users/jarred/.cache/zig'
info : (config): Detected runtime zig version: '0.11.0-dev.4006+bf827d0b5'
info : (server): client is 'Visual Studio Code-1.81.0'
info : (server): zls initializing
info : (server): Server.ClientCapabilities{ .supports_snippets = true, .supports_apply_edits = true, .supports_will_save = true, .supports_will_save_wait_until = true, .supports_publish_diagnostics = true, .supports_code_action_fixall = true, .hover_supports_md = true, .completion_doc_supports_md = true, .label_details_support = true, .supports_configuration = true, .supports_workspace_did_change_configuration_dynamic_registration = true, .supports_textDocument_definition_linkSupport = true }
info : (server): Using offset encoding: utf-16
warning: (server): Zig `0.11.0-dev.4006+bf827d0b5` is older than ZLS `0.12.0-dev.2+9532f4a`. Update Zig to avoid unexpected behavior.
info : (server): Setting configuration...
[...]

PR #1698, opened 2024-01-03
update to new module system

xdBronch:
breaking changes from https://github.com/ziglang/zig/pull/18160
made corresponding prs in [known-folders](https://github.com/ziglibs/known-folders/pull/43) and [diffz](https://github.com/ziglibs/diffz/pull/19), iiuc the hashes i used in the zon will still work once the prs are merged, github magic or whatever :p

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -7,15 +7,16 @@ const zls_version = std.SemanticVersion{ .major = 0, .minor = 12, .patch = 0 };
 /// Remove all usages of `std.mem.copy` and remove `std.mem.set` (#18143)
 const min_zig_string = "0.12.0-dev.1767+1e42a3de89";
 
-pub fn build(b: *std.build.Builder) !void {
-    comptime {
-        const current_zig = builtin.zig_version;
-        const min_zig = std.SemanticVersion.parse(min_zig_string) catch unreachable;
-        if (current_zig.order(min_zig) == .lt) {
-            @compileError(std.fmt.comptimePrint("Your Zig version v{} does not meet the minimum build requirement of v{}", .{ current_zig, min_zig }));
-        }
+const Build = blk: {
+    const current_zig = builtin.zig_version;
+    const min_zig = std.SemanticVersion.parse(min_zig_string) catch unreachable;
+    if (current_zig.order(min_zig) == .lt) {
+        @compileError(std.fmt.comptimePrint("Your Zig version v{} does not meet the minimum build requirement of v{}", .{ current_zig, min_zig }));
     }
+    break :blk std.Build;
+};
 
+pub fn build(b: *Build) !void {
     const target = b.standardTargetOptions(.{});
     const optimize = b.standardOptimizeOption(.{});
 
@@ -110,15 +111,15 @@ pub fn build(b: *std.build.Builder) !void {
     exe.pie = pie;
     b.installArtifact(exe);
 
-    exe.addModule("build_options", exe_options_module);
-    exe.addModule("known-folders", known_folders_module);
-    exe.addModule("diffz", diffz_module);
+    exe.root_module.addImport("build_options", exe_options_module);
+    exe.root_module.addImport("known-folders", known_folders_module);
+    exe.root_module.addImport("diffz", diffz_module);
 
     if (enable_tracy) {
         const client_cpp = "src/tracy/public/TracyClient.cpp";
 
         // On mingw, we need to opt into windows 7+ to get some features required by tracy.
-        const tracy_c_flags: []const []const u8 = if (target.isWindows() and target.getAbi() == .gnu)
+        const tracy_c_flags: []const []const u8 = if (target.result.isMinGW())
             &[_][]const u8{ "-DTRACY_ENABLE=1", "-fno-sanitize=undefined", "-D_WIN32_WINNT=0x601" }
         else
             &[_][]const u8{ "-DTRACY_ENABLE=1", "-fno-sanitize=undefined" };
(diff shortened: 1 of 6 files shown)


xdBronch:
forgot to update the min zig version of course lol

xdBronch:
also there seems to be a regression relating to build runners being cached so zls cant really use its build runner correctly meaning module imports are broken

build.zig
```diff
@@ -4,10 +4,10 @@ const builtin = @import("builtin");
 const zls_version = std.SemanticVersion{ .major = 0, .minor = 12, .patch = 0 };
 
 /// document the latest breaking change that caused a change to the string below:
-/// Remove all usages of `std.mem.copy` and remove `std.mem.set` (#18143)
-const min_zig_string = "0.12.0-dev.1767+1e42a3de89";
+/// Move many settings from being per-Compilation to being per-Module (#18160)
+const min_zig_string = "0.12.0-dev.2015+60958d135";
 
-pub fn build(b: *std.build.Builder) !void {
+pub fn build(b: *std.Build) !void {
```
Techatrix, line 10: The `*std.Build` expression is analyzed before the version check which means that  you will get a `container 'std' has no member called 'Build'` error when building with old zig versions like 0.9.0. The following changes will ensure that the version check is always analyzed first.

```suggestion
const Build = blk: {
    const current_zig = builtin.zig_version;
    const min_zig = std.SemanticVersion.parse(min_zig_string) catch unreachable;
    if (current_zig.order(min_zig) == .lt) {
        @compileError(std.fmt.comptimePrint("Your Zig version v{} does not meet the minimum build requirement of v{}", .{ current_zig, min_zig }));
    }
    break :blk std.Build;
};

pub fn build(b: *Build) !void {
```

build.zig
```diff
@@ -140,6 +140,7 @@ pub fn build(b: *std.build.Builder) !void {
     const gen_exe = b.addExecutable(.{
         .name = "zls_gen",
         .root_source_file = .{ .path = "src/config_gen/config_gen.zig" },
+        .target = target,
```
Techatrix, line 143: This breaks cross compilation because the `zls_gen` executable need to always target the host machine.
```suggestion
        .target = b.host,
```

Techatrix (changes requested):

src/build_runner/master.zig
```diff
@@ -69,15 +69,15 @@ pub fn main() !void {
     cache.addPrefix(global_cache_directory);
     cache.hash.addBytes(builtin.zig_version_string);
 
-    const host = try std.zig.system.NativeTargetInfo.detect(.{});
+    const host = try std.zig.system.resolveTargetQuery(.{});
```
Techatrix, line 72: This doesn't match what I have found in zig's [build_runner.zig](https://github.com/ziglang/zig/blob/04ac028a2cd20ea430dd6a05b956363ec182b6ff/lib/build_runner.zig#L49-L52). `query` appears to store the parameter that has been passed to `resolveTargetQuery`.
```suggestion
    const host: std.Build.ResolvedTarget = .{
        .query = .{},
        .result = try std.zig.system.resolveTargetQuery(.{}),
    };
```
The call to `std.Build.create` will also need to be adjusted.

Techatrix (changes requested):

xdBronch:
oh thats funny, the [changes](https://github.com/ziglang/zig/pull/17431) to the expectEqual function meant to make them better to work with are actually causing CI to fail

nullptrdevs:
Past experience tells me it might be a good idea to rename the current build runner and use it with zig versions < 2030, some/certain projects only update in large increments.

Techatrix:
> Past experience tells me it might be a good idea to rename the current build runner and use it with zig versions < 2030, some/certain projects only update in large increments.

Any projects that doesn't target the latest Zig shouldn't use the latest ZLS as well.
I would rather have the multi build runner system removed at some point instead of extending it with more versions.
Previous Discussion on this: https://github.com/zigtools/zls/issues/1020#issuecomment-1772738942

nullptrdevs:
> > Past experience tells me it might be a good idea to rename the current build runner and use it with zig versions < 2030, some/certain projects only update in large increments.
> 
> Any projects that doesn't target the latest Zig shouldn't use the latest ZLS as well. I would rather have the multi build runner system removed at some point instead of extending it with more versions. Previous Discussion on this: [#1020 (comment)](https://github.com/zigtools/zls/issues/1020#issuecomment-1772738942)

Yeah, well #1396

xdBronch:
realistically it doesnt matter since the 2 commits are like a day apart but updating the minimum version to the expectEqual change isnt required afaict. i tested earlier and the updated tests work with the new and old expectEqualDeep

Techatrix:
FYI, I have edited your PR and done the following:
- updated the minimum build version again
- squashed some of your commits
- updated `build.zig.zon` so that the commit hashes refer to the master/main branch
- updated everything nix related

Techatrix:
> realistically it doesnt matter since the 2 commits are like a day apart but updating the minimum version to the expectEqual change isnt required afaict. i tested earlier and the updated tests work with the new and old expectEqualDeep

The minimum build version had to be updated anyway because of [ziglang/#18438](https://github.com/ziglang/zig/issues/18438), so I did it just in case.

xdBronch:
ah i actually did something a little sneaky before :smile:
i made the minimum version refer to the commit that fixed that issue but the comment refer to the big module pr. in hindsight it probably didnt make much sense to do so but i was being overly pedantic, the comment above says `document the latest breaking change that caused a change to the string below:` but it was a bug fix not a breaking change :p 
either way it doesnt matter because now CI is failing because of yet another breaking change haha

Techatrix (approved):
Everything looks to be in order finally.
The only thing missing are the nix flakes which I will update in a follow up PR once https://github.com/mitchellh/zig-overlay has updated again.

PR #1699, opened 2024-01-06
Update Nix flake inputs

hqnna:
Just updating the flake inputs so it builds on nix again.


SuperAuguste (approved):
Works on your machine :P

PR #1713, opened 2024-01-14
remove build options documentation from README

Techatrix:
Only the `-Ddata_version` option is mentioned.
That option may also give new users the false impression that ZLS is compatible with any Zig version so its better not to show this at the top of the README to avoid confusion.


SuperAuguste (approved):
Good call.

PR #1719, opened 2024-01-17
Fix/improve `union(enum)` completions

### LINKED ISSUE #1707: Autocomplete union(enum) variant
Right now, there is autocomplete for `union(enum)` variants.

Here's a screenshot of such a scenario. The cursor is right after the dot (`.`).

image

My hope is that zls inserts `{ .content_length = <insert position> }` after I press <kbd>Enter</kbd>.

nullptrdevs:
Resolves #1707 

It'd require some rework but should this be extended to anon struct inits?
Should completions include ` = ` for struct fields?

### DIFF
--- a/src/features/completions.zig
+++ b/src/features/completions.zig
@@ -701,10 +701,7 @@ fn completeDot(builder: *Builder, loc: offsets.Loc) error{OutOfMemory}!void {
             dot_context,
         );
         for (containers) |container| {
-            if (dot_context.likely == .enum_arg and !container.isEnumType()) continue;
-            if (dot_context.likely != .struct_field)
-                if (!container.isEnumType() and !container.isUnionType()) continue;
-            try collectContainerFields(builder, container);
+            try collectContainerFields(builder, dot_context.likely, container);
         }
     }
 
@@ -917,6 +914,7 @@ pub fn completionAtIndex(
                 else
                     .{ .TextEdit = .{ .newText = item.insertText orelse item.label, .range = insert_range } };
             }
+            item.insertText = null;
         }
     }
 
@@ -1041,11 +1039,12 @@ fn globalSetCompletions(builder: *Builder, kind: enum { error_set, enum_set }) e
 
 const EnumLiteralContext = struct {
     const Likely = enum { // TODO: better name, tagged union?
-        /// `mye: Enum = .`, `abc.field = .` or `f(.{.field = .` if typeof(field) is enumlike)
-        /// `== .`, `!= .`, switch case
+        /// `mye: Enum = .`, `abc.field = .`, `f(.{.field = .`, `switch` case
         enum_literal,
         /// Same as above, but`f() = .` or `identifier.f() = .` are ignored, ie lhs of `=` is a fn call
         enum_assignment,
+        // `==`, `!=`
+        enum_comparison,
         /// the enum is a fn arg, eg `f(.`
         enum_arg,
         /// `S{.`, `var s:S = .{.`, `f(.{.` or `a.f(.{.`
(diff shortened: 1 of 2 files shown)


Techatrix:
> It'd require some rework but should this be extended to anon struct inits?

Could give an explanation on how completions would look like with anon struct inits?

> Should completions include `=` for struct fields?

I would say yes otherwise it may be a bit unclear on what to type next.

src/features/completions.zig
```diff
@@ -1345,7 +1354,15 @@ pub fn collectContainerFields(
     for (container_decl.ast.members) |member| {
         const field = handle.tree.fullContainerField(member) orelse continue;
         const name = handle.tree.tokenSlice(field.ast.main_token);
-        try completions.append(arena, .{
+        if (likely == .enum_assignment and !field.ast.tuple_like and container.isTaggedUnion()) {
+            try completions.append(arena, .{
+                .label = name,
+                .kind = if (field.ast.tuple_like) .EnumMember else .Field,
+                .detail = Analyser.getContainerFieldSignature(handle.tree, field),
+                .insertText = if (snippets_ok) try std.fmt.allocPrint(arena, "{{ .{s} = $1 }}$0", .{name}) else try std.fmt.allocPrint(arena, "{{ .{s} = ", .{name}),
```
Techatrix, line 1362: I am a bit unsure about this but should this feature be enabled even snippets are disabled? I would have expected to retain the old behavior when snippets are disabled.

Techatrix (changes requested):
This is missing test coverage.

nullptrdevs:
> > It'd require some rework but should this be extended to anon struct inits?
> 
> Could give an explanation on how completions would look like with anon struct inits?

I'm thinking it'd be identical, ie`b.addExecutable(.` would list the fields, just as it does with `b.addExecutable(.{.`, but insert text would be `{ .field_name =`.. how jarring would it be for `.name` to be `{ .field_name = `?

src/features/completions.zig
nullptrdevs: For me snippets is the `$1 $2 $0` functionality, otherwise it's a valid insertText.

xdBronch:
im messing around with this branch a bit and noticed if i tab thru the list of completions it pastes the text including the snippet placeholders 
image
this doesnt happen with any other types of snippets afaict, is this a bug? editor issue? ideally it would just fill the name of the variant when im tabbing then do the snippet on enter

xdBronch:
also why is this only done for tagged unions? regular unions are initialized the same way

nullptrdevs:
> im messing around with this branch a bit and noticed if i tab thru the list of completions it pastes the text including the snippet placeholders image this doesnt happen with any other types of snippets afaict, is this a bug? editor issue? ideally it would just fill the name of the variant when im tabbing then do the snippet on enter

VSCode just selects whichever is highlighted :shrug:

Techatrix (approved):

nullptrdevs:
Hmm, hmm, hmm ... still undecided if the label should be `name` or `{ .name =`
label

nektro:
`name` imo, and then `.{ .` should also complete

nullptrdevs:
> ... `.{ .` should also complete

it does, as long as it is after the `const exe = b.addExecutable(...`

label2


The amount of times I've spent back and forth debugging those  situations $#^#^@^@#

The regression you made me aware is `analyser.resolveOptionalChildType` not working as before, eg `exe.libc_file = .` assigning to optional should unwrap the type.

Techatrix:
@nullptrdevs I could take care of rebasing this PR if you'd like.

Techatrix:
I've removed the second copy of every test that tested with and without snippets and instead used the new `testCompletionTextEdit` function to test completion is performing the expected text edit.
FYI `insertText` gets converted into `textEdit` before sending the response.

nullptrdevs:
> FYI `insertText` gets converted into `textEdit` before sending the response.

It was worth finding out, :+1:

nullptrdevs:
Does your approval stand on this one, merge?

Techatrix:
I have found an issue. In the following example I am doing a comparison which should only complete to `alpha` but completes to `{ .alpha = <cursor>}`.
```zig
const Birdie = enum { canary };
const U = union(enum) { alpha: []const u8 };
const u: U = undefined;
const boolean = u == .<cursor>
```

Can be tested with:
```zig
try testCompletionTextEdit(.{
    .source =
    \\const U = union(enum) { alpha: []const u8 };
    \\const u: U = undefined;
    \\const boolean = u == .<cursor>
    ,
    .label = "alpha",
    .expected_insert_line = "const foo: U = .alphh",
    .expected_replace_line = "const foo: U = .alphh",
});
```

Techatrix (approved):

PR #1738, opened 2024-01-30
find associated build file in case of missing build config data

### LINKED ISSUE #1629: module in build.zig is not considered as a BuildConfig packages

Techatrix:
The associated build file resolution mechanism may fail to find the build file because the necessary build config from running the build.zig may not have finished yet.

These changes should ensure that the associated build file is kept unresolved until all necessary information is available.

fixes #1629

### DIFF
--- a/src/ComptimeInterpreter.zig
+++ b/src/ComptimeInterpreter.zig
@@ -892,7 +892,7 @@ pub fn interpret(
                     } };
                 }
 
-                const import_uri = (try interpreter.document_store.uriFromImportStr(interpreter.allocator, interpreter.getHandle().*, import_str[1 .. import_str.len - 1])) orelse return error.ImportFailure;
+                const import_uri = (try interpreter.document_store.uriFromImportStr(interpreter.allocator, interpreter.getHandle(), import_str[1 .. import_str.len - 1])) orelse return error.ImportFailure;
                 defer interpreter.allocator.free(import_uri);
 
                 const import_handle = interpreter.document_store.getOrLoadHandle(import_uri) orelse return error.ImportFailure;
--- a/src/DocumentStore.zig
+++ b/src/DocumentStore.zig
@@ -156,10 +156,6 @@ pub const Handle = struct {
     /// error messages from comptime_interpreter or astgen_analyser
     analysis_errors: std.ArrayListUnmanaged(ErrorMessage) = .{},
 
-    /// `DocumentStore.build_files` is guaranteed to contain this uri
-    /// uri memory managed by its build_file
-    associated_build_file: ?Uri = null,
-
     /// private field
     impl: struct {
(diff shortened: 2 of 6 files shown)


nullptrdevs (approved):
The only reason it doesn't resolve 1629 is because it stops iterating at the first `is_associated`.

Techatrix:
> The only reason it doesn't resolve 1629 is because it stops iterating at the first `is_associated`.

Resolving the build configs in reverse order appears to resolve it.

```patch
diff --git a/src/DocumentStore.zig b/src/DocumentStore.zig
index 765c2de0..fbff1a33 100644
--- a/src/DocumentStore.zig
+++ b/src/DocumentStore.zig
@@ -304,7 +304,10 @@ pub const Handle = struct {
 
         var has_missing_build_config = false;
 
-        var it = unresolved.has_been_checked.iterator(.{ .kind = .unset });
+        var it = unresolved.has_been_checked.iterator(.{
+            .kind = .unset,
+            .direction = .reverse,
+        });
         while (it.next()) |i| {
             const build_file_uri = unresolved.potential_build_files[i];
             const build_file = document_store.getOrLoadBuildFile(build_file_uri) orelse continue;
```

nullptrdevs:
> Resolving the build configs in reverse order appears to resolve it.

Naturally, but how would that affect the original use-case #688?

In testing this I simply iterated over all of the potential build files, storing the index of the last `is_associated` and assigning it afterwards.

Techatrix:
Isn't "iterate over all potential build files and use the last associated" the same as "iterating over all potential build files in reverse and use the first associated"?

Issue #1739, opened 2024-01-31
not compatible with zig 0.11.0


subh-cs:
### Zig Version

0.11.0

### Zig Language Server Version

N/A

### Steps to Reproduce and Observed Behavior

I'm using ubuntu 22.04 and I downloaded zig from snapstore which installed 0.11.0 LTS version. 
And now I need to install ZLS but when I follow installation guide, it requires latest version. 
```
Your Zig version v0.11.0 does not meet the minimum build requirement of v0.12.0-dev.2138+828d23956
```

### Expected Behavior

1. There should be an installation guide on how to download a specific version of ZLS based on ZIg version.
2. maybe just install ZLS which will take a look at system for existing ZIG version, if ZIG is not installed, it will ask first to install it and then install the specific ZLS accordingly, in this case everything is in abstraction and UX will be awesome.

nektro:
there are in the releases https://github.com/zigtools/zls/releases/tag/0.11.0

PR #1740, opened 2024-02-04
provide a more helpful error when the minimum build version is not met

Techatrix:
The new error message should hopefully inform the user on how to resolve this issue.
See #1739


nullptrdevs (approved):
Neat

PR #1754, opened 2024-02-09
Fix fallback logic in `analysis.getPositionContext`

llogick:
Most lines are indented, ie `line[0] == ' '`(a space).

Tokenizing `line` as the sole buffer returns location offsets relative/within to that line, but are expected to be indexes within `handle.tree.source` .

### DIFF
--- a/src/analysis.zig
+++ b/src/analysis.zig
@@ -2947,10 +2947,10 @@ pub fn getPositionContext(
     var stack = try std.ArrayListUnmanaged(StackState).initCapacity(allocator, 8);
     defer stack.deinit(allocator);
 
-    {
-        var held_line = try allocator.dupeZ(u8, text[0..line_loc.end]);
-        defer allocator.free(held_line);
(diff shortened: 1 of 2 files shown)


Techatrix (changes requested):
This looks like something that could be covered by a test case in [position_context.zig](https://github.com/zigtools/zls/blob/master/tests/utility/position_context.zig).

nullptrdevs:
Having done this makes me realize having that logic is very dicey and
https://github.com/zigtools/zls/blob/05ae41e3f4af684628c97a921b85e07d921f38e2/src/features/completions.zig#L1338
should just use `tok_loc.end` or 
https://github.com/zigtools/zls/blob/05ae41e3f4af684628c97a921b85e07d921f38e2/src/features/completions.zig#L1197
look_ahead = true here (haven't tested for side effects).

Techatrix (approved):

llogick:
This is a low priority PR, the test changes are worth having, but in my opinion the fallback logic shouldn't exist (and obviously nothing depends on it currently).

PR #1763, opened 2024-02-11
improve completions on functions

Techatrix:
- bring back the label details (accidentally removed in #1746)
- remove the function name from the detail
- differentiate between functions and methods
- add empty parentheses when there are no arguments even if snippets are disabled


nullptrdevs (approved):
Fancy work

PR #1764, opened 2024-02-12
highlight symbols that call `@compileError` as deprecated

Techatrix:
Screenshot from 2024-02-12 01-10-32


llogick (approved):

tcka4enckoalexei2014-commits (commented):
Go

PR #1793, opened 2024-02-24
update http code of config_gen

Techatrix:
See https://github.com/ziglang/zig/pull/18955


nullptrdevs (approved):
Up to you, the regression only affects the website linux build of zig and those that don't have the deps yet.

Techatrix:
How about keeping support for the older Zig versions and not bumping the minimum build version

PR #1803, opened 2024-02-29
Destructure hover

gabydd:
This implements type resolution for assign destructure using the right hand side and the index, it works for the uses I have tried but if there is a different way zls wants to approach this thats totally fine to me. I am also quite new to zig so this might not be the most idiomatic way to do this.

### DIFF
--- a/src/analysis.zig
+++ b/src/analysis.zig
@@ -3660,7 +3660,26 @@ pub const DeclWithHandle = struct {
                 })) orelse return null,
                 .Single,
             ),
-            .assign_destructure => null, // TODO
+            .assign_destructure => |pay| {
+                const data = tree.nodes.items(.data);
+                const rh = data[pay.node].rhs;
+                const node = try analyser.resolveTypeOfNode(.{ .node = rh, .handle = self.handle });
+                if (node) |right| {
+                    return switch (right.data) {
+                        .array => blk: {
+                            var buffer: [2]Ast.Node.Index = undefined;
+                            break :blk try analyser.resolveTypeOfNode(.{
+                                .node = tree.fullArrayInit(&buffer, rh).?.ast.elements[pay.index],
+                                .handle = self.handle,
+                            });
+                        },
+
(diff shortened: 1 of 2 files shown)


src/analysis.zig
```diff
@@ -3660,7 +3660,25 @@ pub const DeclWithHandle = struct {
...
+                }) orelse return null;
+                return switch (node.data) {
+                    .array => blk: {
+                        var buffer: [2]Ast.Node.Index = undefined;
+                        const array = tree.fullArrayInit(&buffer, rhs) orelse return null;
+                        break :blk try analyser.resolveTypeOfNode(.{
+                            .node = array.ast.elements[pay.index],
+                            .handle = self.handle,
+                        });
+                    },
```
Techatrix, line 3677: The child type of the array has already been resolved.
```suggestion
                    .array => |array_info| try array_info.elem_ty.instanceTypeVal(analyser),
```

src/analysis.zig
```diff
@@ -3660,7 +3660,25 @@ pub const DeclWithHandle = struct {
...
+                    .other => try analyser.resolveTupleFieldType(node, pay.index),
+                    else => null,
+                };
+            }, // TODO
```
Techatrix, line 3681: TODO solved.
```suggestion
            },
```

Techatrix (changes requested):

src/analysis.zig
Techatrix: I should have also mentioned that this makes the `rhs` variable above unused.

gabydd: yeah just fixing that manually and testing will push in a second

Techatrix:
While testing this, I realized that adding an explicit type specifier to the variable was not being used to resolve the type.
Here is an example:
```zig
test {
    const bar, const baz: u32 = undefined;
}
```
Hovering over `baz` was not showing `u32`. I have fixed this and pushed the change to your PR.

Techatrix (approved):
Thank you very much.

PR #1804, opened 2024-03-01
inlay hints for destructured variables

gabydd:
hacked on this tonight should be good to go, one thing I have noticed is:
```zig
const a = .{1, @as(u32, 2)};
```
does not give a tuple type(it just gives an array of the first element type) and therefore doesn't destructure properly in inlay hints or hover, but that seems like it was an issue before this so I'll make a seperate issue for tracking it.

### DIFF
--- a/src/features/inlay_hints.zig
+++ b/src/features/inlay_hints.zig
@@ -354,6 +354,18 @@ fn writeNodeInlayHint(
                 try typeStrOfNode(builder, node) orelse return,
             );
         },
+        .assign_destructure => {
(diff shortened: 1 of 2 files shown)


Techatrix:
> ```zig
> const a = .{1, @as(u32, 2)};
> ```

Relevant code:
https://github.com/zigtools/zls/blob/0f12bbac5ee4e5384996e97a951b1df955d8911a/src/analysis.zig#L1554-L1556

There currently is not representation for tuple types in ZLS which should be added to resolve this issue.

Techatrix (approved):

PR #1822, opened 2024-03-07
truncate type hints on anonymous container decls

Techatrix:
### Before
Screenshot from 2024-03-07 14-47-58
### After
Screenshot from 2024-03-07 14-48-13


Techatrix:
> Is the hint even necessary here? it is duplicate information, therefore noise

I guess this could be considered another example where https://github.com/zigtools/zls/issues/1526 could apply.

But also, my example is only there to illustrate the changes of this PR. 
No one in their right mind would write code like this. I hope...

PR #1827, opened 2024-03-13
Improved `positionLessThan()`, `rangeLessThan()` and `extractString()`

ThisAccountHasBeenSuspended:
I looked at the source code and when I saw these functions I really wanted to change them.

I changed the `extractString()` function because multiple underlying functions are called for a single character which are not necessary and reduce performance.

I found a lot of things that can be changed to make the source code more readable and increase performance. Maybe it would be appropriate to focus on that.

> [!NOTE]
> These small changes also reduced the size of the compiled file from `3.198 KB` to `3.197 KB`.

### DIFF
--- a/src/offsets.zig
+++ b/src/offsets.zig
@@ -834,21 +834,10 @@ fn testGetNCodeUnitByteCount(text: []const u8, n: [3]usize) !void {
     try std.testing.expectEqual(n[0], getNCodeUnitByteCount(text, n[2], .@"utf-32"));
 }
 
-pub fn rangeLessThan(a: types.Range, b: types.Range) bool {
+pub inline fn rangeLessThan(a: types.Range, b: types.Range) bool {
     return positionLessThan(a.start, b.start) or positionLessThan(a.end, b.end);
 }
 
-pub fn positionLessThan(a: types.Position, b: types.Position) bool {
-    if (a.line < b.line) {
-        return true;
-    }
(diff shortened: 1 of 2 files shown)


Techatrix:
The `inline` keyword on function is not a hint to the optimizer to inline the function. It causes semantic analysis to inline the function body at the callsite before reaching the code backend. It should ideally only be used when it is needed for the code to become semantically valid. A hint to the optimizer is instead possible through `callconv(.Inline)` or `@call(.always_inline, ...)`. It also best to leave decision making on inlining to the optimizer unless a performance benchmark has shown a significant performance improvement.

The change to `positionLessThan` is only stylistic. It doesn't make the code more or less readable in my opinion.
I also don't see how the change to `extractString` makes it more readable. The old version reads like a simple sentence: "if the string starts with a quotation mark and ends with a quotation mark then ...".

If the goal is to improve performance then I would suggest to focus where the hotspots in the ZLS codebase are located because functions like `extractString` are not one of them. ZLS's performance can always be improved so any contribution here is very much welcome.

If you remove the usage of the `inline` keyword then I can still approve this.

xdBronch:
> A hint to the optimizer is instead possible through `callconv(.Inline)` or `@call(.always_inline, ...)`

both of these have the same semantic effects as `inline fn` afaict. in fact zig fmt will turn `callconv(.Inline)` into `inline fn`

PR #1834, opened 2024-03-20
Fix zig std api breaking (std.os.iovec_const -> std.posix.iovec_const)

jimying:
Detail:
extract std.posix from std.os
https://github.com/ziglang/zig/pull/19354


SuperAuguste:
Will merge https://github.com/zigtools/zls/pull/1836 once Zig's download page updates. Thanks for making this PR. :)

PR #1839, opened 2024-03-26
Fix build for latest Zig master

ianprime0509:
As of https://github.com/ziglang/zig/pull/19414, a small workaround is necessary to avoid global variables containing references to comptime vars.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -6,8 +6,8 @@ const zls_version = std.SemanticVersion{ .major = 0, .minor = 12, .patch = 0 };
 const zls_version_is_tagged: bool = false;
 
 /// document the latest breaking change that caused a change to the string below:
-/// std.os: extract and separate std.posix
-const min_zig_string = "0.12.0-dev.3385+3a836b480";
+/// compiler: implement analysis-local comptime-mutable memory
+const min_zig_string = "0.12.0-dev.3451+405502286";
 
 pub fn build(b: *Build) !void {
     const target = b.standardTargetOptions(.{});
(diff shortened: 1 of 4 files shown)


nektro:
this patch feels like a workaround for a regression. this is the intended behavior?

ianprime0509:
It's the same workaround applied in several files of the PR that caused this change, e.g. https://github.com/ziglang/zig/pull/19414/files#diff-810d7757edd41ea303795cbc7b7e2a6f1a9ab93fbd91db1495796e791997811e

Based on that, my assumption is that it's intentional.

nektro:
perhaps there should be a check that the reference being taken is the last line of the block to allow it

llogick:
It's just a verbose/pedantic way of making a comptime value accessible at runtime, ie placing it in `.data`.

mochalins:
I'd just like to comment that I arrived at a different change to fix compilation/make tests pass:

In `uri.zig`, `&escapes` can adjusted to `escapes`
In `lsp.zig`, `kvs_array[0..]` can adjusted to `kvs_array`

Might you have any insight as to whether these changes would, in effect, be identical to pre-PR behavior? @ianprime0509

ianprime0509:
@mochalins that seems reasonable, thanks for the suggestion! Neither occurrence has anything relying on the type being a slice or pointer to an array, so just using the array by itself seems to be totally fine here. I'll update this change with your suggestion.

Techatrix (approved):
nice!

Issue #1853, opened 2024-04-11
config_gen compile error


weskoerber:
### Zig Version

0.12.0-dev.3632+7fb5a0b18

### Zig Language Server Version

4e01c08f558ea07462aaa7b71d2a24f86f47a855

### Client / Code Editor / Extensions

NVIM v0.10.0-dev-2869+g4459e0cee with nvim-lspconfig

### Steps to Reproduce and Observed Behavior

Update to `master` and build:
```console
git switch -d 4e01c08f558ea07462aaa7b71d2a24f86f47a855
zig build
```

Output:
```
❯ zig build
install
└─ install zls
   └─ zig build-exe zls Debug native
      └─ run zls_gen (version_data_master_19824.zig)
         └─ zig build-exe zls_gen Debug native 1 errors
src/config_gen/config_gen.zig:990:20: error: member function expected 0 argument(s), found 1
        try request.send(.{});
            ~~~~~~~^~~~~
/home/wes/.local/share/zvm/master/zig-linux-x86_64-0.12.0-dev.3632+7fb5a0b18/lib/std/http/Client.zig:824:9: note: function declared here
    pub fn send(req: *Request) SendError!void {
    ~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
referenced by:
    generateVersionDataFile: src/config_gen/config_gen.zig:828:30
    main: src/config_gen/config_gen.zig:1115:13
    remaining reference traces hidden; use '-freference-trace' to see all reference traces
error: the following command failed with 1 compilation errors:
/home/wes/.local/share/zvm/master/zig-linux-x86_64-0.12.0-dev.3632+7fb5a0b18/zig build-exe -fsingle-threaded -ODebug -Mroot=/home/wes/Documents/repos/zig/zls/src/config_gen/config_gen.zig --cache-dir /home/wes/Documents/repos/zig/zls/zig-cache --global-cache-dir /home/wes/.cache/zig --name zls_gen --listen=-
Build Summary: 3/8 steps succeeded; 1 failed (disable with --summary none)
install transitive failure
└─ install zls transitive failure
   └─ zig build-exe zls Debug native transitive failure
      └─ run zls_gen (version_data_master_19824.zig) transitive failure
         └─ zig build-exe zls_gen Debug native 1 errors
error: the following build command failed with exit code 1:
[...]

PR #1852, opened 2024-04-11
Function argument removed

ThisAccountHasBeenSuspended:
This function no longer contains any arguments.

```
install
└─ install zls
   └─ zig build-exe zls ReleaseSafe native
      └─ run zls_gen (version_data_master_19824.zig)
         └─ zig build-exe zls_gen Debug native 1 errors
src\config_gen\config_gen.zig:990:20: error: member function expected 0 argument(s), found 1
        try request.send(.{});
```

https://github.com/ziglang/zig/pull/19598/files#diff-ab7384310a623b1845d2013b68374070ab590b1cc82d786e188fadc3f70ccda5L805

PR #1854, opened 2024-04-11
fix: superfluous argument in http req. send method

weskoerber:
related: 845a9a01

closes: #1853


weskoerber:
Sorry... did not look at existing PRs before submitting.

Closing in favor of #1852.

PR #1876, opened 2024-04-28
Workspace Symbols

SuperAuguste:
**Note: I plan on editing this write-up and publishing it someplace once this PR is merged :)**

## Introduction

We're seeking to add workspace symbols, that is the ability to search for declarations by name across all the files in your editor's file tree, to ZLS. This incurs significant construction and access overhead which must be minimized to maintain interactivity, which is crucial in the editor environments where ZLS is run.

## N-grams and trigrams

An n-gram is a chunk of text of size `n`. It is produced by sliding a window of size `n` and stride 1 across a larger chunk of text, ending when a new window of size `n` cannot be created. A trigram is an n-gram where `n = 3`.

To obtain the trigrams of `agent`, for example, we obtain the first 3 characters, `age`, then shift our window by 1 to obtain `gen`, and again to obtain `ent`. Shifting our window by 1 again would not yield 3 characters, so we stop here. Thus, the trigrams of agent are `age`, `gen`, and `ent`.

N-grams are a nice way to execute approximate searches over a large corpus of text, allowing the consideration not only the entirety, prefix, or suffix of a search target, but also all of its constituent parts. Trigrams also enable efficient large-scale regular expression searches (see [Zoekt](https://github.com/sourcegraph/zoekt) from my ex-employer Sourcegraph), but that's out of scope for this article.

## Indexing

We need to index the name of every single global constant, variable, and function declaration. This is easily doable with the now-refactored [DocumentScope](https://github.com/zigtools/zls/blob/34ab587a9e8c180a1a808e7fd52007b67b35663f/src/DocumentScope.zig), which lists all declarations in a single contiguous list.

Note that we could perform trigram indexing immediately during the construction of the `DocumentScope`, but that would incur overhead on every edit that we'd rather split into a separate task in our multithreaded setup to keep ZLS fast and responsive.

We can begin by attaching a flag, `should_be_indexed_for_trigrams`, during the construction of the `DocumentScope` to each declaration identifying whether it's one of our search targets, thus preventing locals and symbols with names shorter than three characters long from being indexed.

During indexing, we iterate over the declarations for each document and find the trigrams for their names. We then create an inverse mapping from each trigram in the declaration's name to the declaration.

So for example, if declaration `Declaration.Index(1)` has name `agent`, our inverse mapping would look like this:
```
age -> [all other declarations containing trigram age ..., Declaration.Index(1)]
gen -> [all other declarations containing trigram gen ..., Declaration.Index(1)]
ent -> [all other declarations containing trigram ent ..., Declaration.Index(1)]
```

This inverse mapping is constructed per-document. After it is constructed, we also construct a Binary Fuse filter to quickly disqualify documents that do not contain certain trigrams at query time.

## Querying

We begin by obtaining the trigrams for our query. We then iterate through all our documents and check if each document contains each trigram in our query via the Binary Fuse filter, which cannot return false negatives but can return false positives, albeit with a very low false positive rate. This allows us to reduce our computation to only documents that likely have all our query trigrams, and is especially effective for longer queries.

Once we've gathered our candidate documents, querying is essentially just performing an intersection.

We use a "merge intersection," which is ripped out of merge sort, to intersect lists. [King](https://github.com/kprotty) tried beating this approach in a couple of purely hashmap-based ways, but with our setup which mostly involves many small inverse mappings (~10,000 trigrams with ~30 declarations each, for example), the merge intersection always won out.

I took a look at and partially implemented [Fast Set Intersection in Memory](https://arxiv.org/pdf/1103.2409), but it seems to be significant overkill for this sort of small intersection application. In Section 4, "Experimental Evaluation," they show that merge intersection performs rather well and sometimes comparably to the implementations shown in the paper for small intersection sizes, so that's what we're sticking with unless [someone can find a better solution](https://github.com/zigtools/trigram-bench).

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -68,6 +68,7 @@ pub fn build(b: *Build) !void {
 
     const known_folders_module = b.dependency("known_folders", .{}).module("known-folders");
     const diffz_module = b.dependency("diffz", .{}).module("diffz");
+    const fastfilter_module = b.dependency("fastfilter", .{}).module("fastfilter");
     const tracy_module = getTracyModule(b, .{
         .target = target,
         .optimize = optimize,
@@ -124,6 +125,7 @@ pub fn build(b: *Build) !void {
         .imports = &.{
             .{ .name = "known-folders", .module = known_folders_module },
             .{ .name = "diffz", .module = diffz_module },
+            .{ .name = "fastfilter", .module = fastfilter_module },
             .{ .name = "tracy", .module = tracy_module },
             .{ .name = "build_options", .module = build_options_module },
             .{ .name = "version_data", .module = version_data_module },
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -15,8 +15,8 @@
             .hash = "1220102cb2c669d82184fb1dc5380193d37d68b54e8d75b76b2d155b9af7d7e2e76d",
         },
         .fastfilter = .{
-            .url = "https://github.com/hexops/fastfilter/archive/b504ba1e636fb4847a7978c10ea32c64d457aaeb.tar.gz",
-            .hash = "1220e672d4980c183c6b29cf9b1200c3db23a7987968b0de7e6357f49b6b174c3276",
+            .url = "https://github.com/hexops/fastfilter/archive/b6e46f6d4811da0d8a0f8675ab85407172a207ba.tar.gz",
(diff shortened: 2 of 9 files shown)


Console-buche:
That's a feature I'd love to see merged in ZLS. I see last commit dates back to almost a year ago. Any plans to continue work on this @SuperAuguste?

PR #1877, opened 2024-04-29
Update `std.posix.iovec` field names for master branch

mochalins:
Updated for commit ziglang/zig@aecd9cc6d152443dc7c02dfe373be654d8adae64

PR #1887, opened 2024-05-05
fix: 0.13.0-dev.68+b86c4bde6 compatibility

ert78gb:
Summary:
- use std.fs.cwd().writeFile instead of deprecated std.fs.cwd().writeFile2
- rename the members of the new std.posix.iovec_const
- bump minimum_zig_version 0.13.0-dev.33+8af59d1f9 => 0.13.0-dev.68+b86c4bde6

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -7,10 +7,10 @@ const zls_version = std.SemanticVersion{ .major = 0, .minor = 13, .patch = 0 };
 const zls_version_is_tagged: bool = false;
 
 /// Specify the minimum Zig version that is required to compile and test ZLS:
-/// ComptimeStringMap: return a regular struct and optimize
+/// Rename Dir.writeFile2 -> Dir.writeFile and update all callsites
 ///
 /// Must match the `minimum_zig_version` in `build.zig.zon`.
-const minimum_zig_version = "0.13.0-dev.33+8af59d1f9";
+const minimum_zig_version = "0.13.0-dev.68+b86c4bde6";
 
 /// Specify the minimum Zig version that is required to run ZLS:
 /// Release 0.12.0
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -3,7 +3,7 @@
     // Must match the `zls_version` in `build.zig`
     .version = "0.13.0-dev",
     // Must match the `minimum_zig_version` in `build.zig`
-    .minimum_zig_version = "0.12.0",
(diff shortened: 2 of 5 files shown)


ert78gb:
if I see well one part of this PR covered by #1877

mochalins:
The minimum Zig version here is incorrect. New `iovec` field names were not yet introduced here.

ert78gb:
it was introduced in the master. I recognised the LLVM 18 CI issue after I opened the PR. If I see well they are actively working on LLVM.  I am keeping my eyes on the zig releases and I will update the PR when the new release will available.

build.zig
```diff
@@ -10,7 +10,7 @@ const zls_version_is_tagged: bool = false;
 /// ComptimeStringMap: return a regular struct and optimize
```
mochalins, line 10: This comment should be updated.

build.zig
```diff
@@ -10,7 +10,7 @@ const zls_version_is_tagged: bool = false;
...
 ///
 /// Must match the `minimum_zig_version` in `build.zig.zon`.
-const minimum_zig_version = "0.13.0-dev.33+8af59d1f9";
+const minimum_zig_version = "0.13.0-dev.46+3648d7df1";
```
mochalins, line 13: Minimum Zig version should be the commit that causes errors on deprecated function usage. Additionally, this version should also be specified in `build.zig.zon`.

mochalins (changes requested):

mochalins:
> it was introduced in the master. I recognised the LLVM 18 CI issue after I opened the PR. If I see well they are actively working on LLVM. I am keeping my eyes on the zig releases and I will update the PR when the new release will available.

Generally I believe it is best practice for the PR to actually contain the versions necessary for the patch, regardless of whether it is yet available on upstream download links. Regardless, newer Zig versions are now available to the CI.

build.zig
ert78gb: Thanks for your help. This is my 2nd day learning zig. Could you help me what does the number after the `dev.` mean? The last part is the commit hash but I don't know what is the 46 or 33 was earlier

ert78gb: I found the `zig version` command print it

ert78gb:
I updated the`minimum_zig_version` let me know if I should update something else.

Techatrix (approved):

PR #1888, opened 2024-05-07
introduce 0.13-dev master build runner

leecannon:
`writeFile2` was renamed to `writeFile`.

@Techatrix I'm a bit confused about the check in `BuildRunnerVersion.zig` line 32 regarding the `minimum_runtime_zig_version` as having a master build runner does not mean we can't support 0.12 at runtime, for now I have commented the assert that gets hit out and will leave this as a draft.

### DIFF
--- a/src/build_runner/0.12.0.zig
+++ b/src/build_runner/0.12.0.zig
@@ -295,7 +295,17 @@ pub fn main() !void {
         }
         const s = std.fs.path.sep_str;
         const tmp_sub_path = "tmp" ++ s ++ (output_tmp_nonce orelse fatal("missing -Z arg", .{}));
-        local_cache_directory.handle.writeFile2(.{
+
+        const writeFileFn = comptime blk: {
+            const writeFile2_removed_version =
+                std.SemanticVersion.parse("0.13.0-dev.68+b86c4bde6") catch unreachable;
+            break :blk if (builtin.zig_version.order(writeFile2_removed_version) == .lt)
+                std.fs.Dir.writeFile2
+            else
+                std.fs.Dir.writeFile;
+        };
+
+        writeFileFn(local_cache_directory.handle, .{
             .sub_path = tmp_sub_path,
             .data = buffer.items,
             .flags = .{ .exclusive = true },
@@ -304,6 +314,7 @@ pub fn main() !void {
                 local_cache_directory, tmp_sub_path, @errorName(err),
             });
         };
+
         process.exit(3); // Indicate configure phase failed with meaningful stdout.
     }
 


Techatrix (changes requested):
This is not well documented, but the `minimum_runtime_zig_version_string` specifies the latest version supported by the latest build runner. If the `master.zig` exists then the minimum runtime Zig version must be a development build. If there is no `master.zig` then it must be the latest tagged release. The assertion you have commented out is meant to force you to bump the `minimum_runtime_zig_version_string` when introducing a `master.zig`. Zig `0.12.0` should still be usable. We used to have build runners for `0.10.0` and `0.11.0` but the minimum Zig version was still `0.12.0-dev`.

With that being said, I was hoping to avoid reintroducing a copy of the build runner. Instead, `@import("builtin").zig_version` should be used to deal with breaking changes. This allows the `minimum_runtime_zig_version` to remain at `0.12.0`. If `0.12.0` and `0.13.0-dev` diverge too much, we can split them into separate files like we used to.

leecannon:
Okay removed master build runner.

Although I still don't understand the assertion, it means that if we have a master builder runner then the tagged release build runner can never be used without hitting the assertion, so why have the build runner selection at all?

Techatrix:
> Although I still don't understand the assertion, it means that if we have a master builder runner then the tagged release build runner can never be used without hitting the assertion, so why have the build runner selection at all?

Why would you hit the assertion?
Nothing is being asserted on the `runtime_zig_version`.

### Example `master.zig` exists and `minimum_runtime_zig_version_string` is `0.13.0-dev.68+b86c4bde6`
```zig
comptime std.debug.assert(minimum_runtime_zig_version_is_tagged == !has_master);
comptime std.debug.assert(minimum_runtime_zig_version_is_tagged == !@hasField(BuildRunnerVersion, "master"));
comptime std.debug.assert(minimum_runtime_zig_version_is_tagged == !true);
comptime std.debug.assert(minimum_runtime_zig_version_is_tagged == false);
```
```zig
const minimum_runtime_zig_version = comptime std.SemanticVersion.parse(build_options.minimum_runtime_zig_version_string) catch unreachable;
const minimum_runtime_zig_version = comptime std.SemanticVersion.parse("0.13.0-dev.68+b86c4bde6") catch unreachable;
const minimum_runtime_zig_version_is_tagged = minimum_runtime_zig_version.build == null and minimum_runtime_zig_version.pre == null;
const minimum_runtime_zig_version_is_tagged = false;
```

### Example `master.zig` doesn't exist and `minimum_runtime_zig_version_string` is `0.12.0`
```zig
comptime std.debug.assert(minimum_runtime_zig_version_is_tagged == !has_master);
comptime std.debug.assert(minimum_runtime_zig_version_is_tagged == !@hasField(BuildRunnerVersion, "master"));
comptime std.debug.assert(minimum_runtime_zig_version_is_tagged == !false);
comptime std.debug.assert(minimum_runtime_zig_version_is_tagged == true);
```
```zig
const minimum_runtime_zig_version = comptime std.SemanticVersion.parse(build_options.minimum_runtime_zig_version_string) catch unreachable;
const minimum_runtime_zig_version = comptime std.SemanticVersion.parse("0.12.0") catch unreachable;
const minimum_runtime_zig_version_is_tagged = minimum_runtime_zig_version.build == null and minimum_runtime_zig_version.pre == null;
const minimum_runtime_zig_version_is_tagged = true;
```

leecannon:
Guess i'm just confused by the name `minimum_runtime_zig_version` as in the case of having a master builder runner and setting the minimum runtime version to 0.13-dev, we still support a runtime zig version of 0.12.0.

Techatrix (approved):

PR #1892, opened 2024-05-10
scm: Update build to compile with latest master

mochalins:
Depends on https://github.com/ziglibs/known-folders/pull/48 and https://github.com/ziglibs/diffz/pull/20


Techatrix:
> Unsure if the build runner should be updated as well or if a new build runner for 0.13.0 should be added...

Ideally, the 0.12.0 runner should be updated to check to version so that 0.12 and master are supported by the same build runner. Just like in #1888

I can take a look at this.

mochalins:
> Ideally, the 0.12.0 runner should be updated to check to version so that 0.12 and master are supported by the same build runner. Just like in #1888
> 
> I can take a look at this.

Ah I see, thank you!

Techatrix (approved):

Issue #1728, opened 2024-01-26
zls shows wrong variable type hint when RHS expression has function returning optional and orelse is used to return error set


eshom:
### Zig Version

0.12.0-dev.2341+92211135f

### Zig Language Server Version

0.12.0-dev.365+a8a83b6

### Client / Code Editor / Extensions

emacs 29.1 (eglot)

### Steps to Reproduce and Observed Behavior

With variable type hints enabled (should be the default) paste to your editor:
```zig
const std = @import("std");

fn myNull() ?i32 {
    return null;
}

pub fn main() !void {
    const y = myNull();
    const x = myNull() orelse error.NullValue;
    std.debug.print("My value: {d}\n", .{try x});
    _ = y;
}
```

In my editor notice that type hint for `x` is `i32`:
![image](https://github.com/zigtools/zls/assets/48328872/5a2e5d04-866a-433d-a887-0377ec84bbac)

### Expected Behavior

Expected type hint for `x` is `error{NullValue}!i32`

eshom:
Not fixed as of `0.15.0-dev.53+2d56313`

![Image](https://github.com/user-attachments/assets/78ef6ec6-dad1-4d23-a69b-470292c27bbd)

Reduced code:
```zig
pub fn main() !void {
    const y = @as(?i32, null);
    const x = y orelse error.NullValue;
    _ = x;
}
```

PR #1896, opened 2024-05-16
Fix `orelse` type resolution

### LINKED ISSUE #1330: Wrong variable type info when orelse assigns variable with an optional type
### Zig Version

0.11.0-dev.4002+7dd1cf26f

8da21e159e3dec9bc6fe01320313cf8108ce4d0b

### Steps to Reproduce

Paste this code into an editor with zls running as lsp.
Send a textDocument/hover request of variable c.
```zig
pub fn opt_test(a: ?i32, b: ?i32) void {
    const c = a orelse b;
    std.debug.print("{s}\n", .{@typeName(@TypeOf(c))});
}
```
Running this code snippet prints ?i32 as type of c.
```zig
const std = @import("std");

pub fn main() !void {
    const a = null;
    const b = null;
    opt_test(a, b);
}

pub fn opt_test(a: ?i32, b: ?i32) void {
    const c = a orelse b;
    std.debug.print("{s}\n", .{@typeName(@TypeOf(c))});
}
```

### Expected Behavior

Type of c being ?i32 .

### Actual Behavior

zls reports the type of c being i32.

I took a stab at trying to come up with a fix by replacing [this switch case](https://github.com/zigtools/zls/blob/8da21e159e3dec9bc6fe01320313cf8108ce4d0b/src/analysis.zig#L1188) with the one below:
```zig
                .@"orelse" => orelse_type: {
                    const base_right = .{ .node = datas[node].rhs, .handle = handle };
                    const base_type_right = (try analyser.resolveTypeOfNodeInternal(base_right));
                    if (base_type_right) |type_right| {
                        const ok = switch (type_right.type.data) {
                            .other => |node_type| switch (node_tags[node_type]) {
                                .optional_type => true,
                                else => false,
                            },
                            else => false,
                        };
                        if (ok) {
                            break :orelse_type base_type;
                        }
                    }
                    break :orelse_type try analyser.resolveUnwrapOptionalType(base_type);
                },
```
This worked for the example above, but when trying to open [hover.zig](src/features/hover.zig) with this change zls crashes.

Sekky61:
Fixes #1330 

See the added tests, they make it pretty clear. One example:

```zig
const foo: ?i32 = 5;
const b<cursor>ar = foo orelse foo;
```

outputs

```zig
?i32
```

This is my first PR so I'd be happy for any pointers.

### DIFF
--- a/src/analysis.zig
+++ b/src/analysis.zig
@@ -799,6 +799,13 @@ pub fn resolveOptionalUnwrap(analyser: *Analyser, optional: Type) error{OutOfMem
     }
 }
 
+pub fn resolveOrelseType(analyser: *Analyser, lhs: Type, rhs: Type) error{OutOfMemory}!?Type {
+    return switch (rhs.data) {
+        .optional => rhs,
+        else => try analyser.resolveOptionalUnwrap(lhs),
+    };
+}
+
 /// Resolves the child type of an optional type
 pub fn resolveOptionalChildType(analyser: *Analyser, optional_type: Type) error{OutOfMemory}!?Type {
     _ = analyser;
@@ -1472,10 +1479,7 @@ fn resolveTypeOfNodeUncached(analyser: *Analyser, node_handle: NodeWithHandle) e
                 .array_access => try analyser.resolveBracketAccessType(base_type, .Single),
                 .@"orelse" => {
                     const type_right = try analyser.resolveTypeOfNodeInternal(.{ .node = datas[node].rhs, .handle = handle }) orelse return null;
-                    return switch (type_right.data) {
-                        .optional => type_right,
-                        else => try analyser.resolveOptionalUnwrap(base_type),
-                    };
+                    return try analyser.resolveOrelseType(base_type, type_right);
                 },
                 .@"catch" => try analyser.resolveUnwrapErrorUnionType(base_type, .payload),
                 .@"try" => try analyser.resolveUnwrapErrorUnionType(base_type, .payload),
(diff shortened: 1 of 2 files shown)


Sekky61:
I have just noticed #1728, so I will work on that before opening the PR.

Sekky61:
In the end I opted for a simpler solution. I noticed `orelse` has basically the same problem as `if` and others -- the unification of the types from both child nodes / branches.

I looked into `resolvePeerTypes` and how the `Either` type is used but that may be a discussion for another issue.

src/analysis.zig
```diff
@@ -1853,13 +1859,29 @@ fn resolveTypeOfNodeUncached(analyser: *Analyser, node_handle: NodeWithHandle) e
                 label_token: ?Ast.TokenIndex,
                 then_expr: Ast.Node.Index,
                 else_expr: Ast.Node.Index,
-            } = if (ast.fullWhile(tree, node)) |while_node|
-                .{
+            } = if (ast.fullWhile(tree, node)) |while_node| no_hang: {
+
+                // while (true) {} is a noreturn type
```
Techatrix, line 1864: For this to be correct, you will need to check that the loop has no `break`s. Something like `while (true) { break 5; };` resolves to `comptime_int` instead of `noreturn`. 
We do have the code for resolving the type of `for` or `while` loops based on all the break operand but it is very incomplete. You could work on improving it but that is not really related to the `orelse` operation. Just like you mentioned, an implementation of peer type resolution would be needed.

Techatrix (changes requested):

src/analysis.zig
Sekky61: Yeah, this was a bit premature. Let's wait for the proper implementation.

Techatrix (approved):

PR #1899, opened 2024-05-20
don't autofix discarded error capture

### LINKED ISSUE #1898: autofix with catch discard is broken
### Zig Version

0.13.0-dev.211+6a65561e3

0.13.0-dev.30+ab928b5

sublime text stable build 4169

The following code doesn't get autofixed correctly on save.
```zig
test {
    const x = @as(anyerror!u8, 12) catch |_| {};
}
```
The first issue is that each save produces a new incorrect autofix.
```zig
test {
    const x = @as(anyerror!u8, 12) catch |_| {
        _ = _; // autofix
        _ = _; // autofix
    };
}
```
The second issue is that a discard of x isn't created.

And third, maybe the discard capture should be removed?

related to #689

### Expected Behavior

```zig
test {
    const x = @as(anyerror!u8, 12) catch {};
    _ = x; // autofix
}



### Relevant log output

_No response_

Techatrix:
fixes #1898

### DIFF
--- a/src/features/code_actions.zig
+++ b/src/features/code_actions.zig
@@ -34,6 +34,7 @@ pub const Builder = struct {
             .non_camelcase_fn => try handleNonCamelcaseFunction(builder, actions, loc),
             .pointless_discard => try handlePointlessDiscard(builder, actions, loc),
             .omit_discard => |id| switch (id) {
+                .@"error capture; omit it instead" => {},
                 .@"error capture" => try handleUnusedCapture(builder, actions, loc, remove_capture_actions),
             },
             // the undeclared identifier may be a discard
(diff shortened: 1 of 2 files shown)


travisstaloch (approved):
Lgtm :+1:.

I test drove this and now realize that my idea to remove the discard is better left to the user.  And leaving the discard prevents the second autofix (`_ = x;`) from happening.

Techatrix:
Oh, I overlooked the thing about the missing `_ = x;`. This is because AstGen stops running after encountering this error (See
[AstGen.zig#L6021](https://github.com/ziglang/zig/blob/1550b5b16d4899cb4f7184ee3d392b4a6197ed42/lib/std/zig/AstGen.zig#L6021)) . This could be changed to continue as if no capture has been made.

Issue #1837, opened 2024-03-24
Rename doesn't rename struct initialization


johan0A:
### Zig Version

0.12.0-dev.3245+4f782d1e8

### Zig Language Server Version

0.12.0-dev.496+96eddd0

### Client / Code Editor / Extensions

vscode

### Steps to Reproduce and Observed Behavior

Rename field inside TestStruct, initialize struct using syntax below ⇾ the initialization does not get updated

```zig
const TestStruct = struct {
    test_var_renamed: i32, //variable was renamed here
};

test "test" {
    const test_struct: TestStruct = .{
        .test_var = 1, //does not get renamed
    };

    const test_struct2: TestStruct = TestStruct{
        .test_var = 1, //does not get renamed
    };

    std.debug.print("{}", .{test_struct});
    std.debug.print("{}", .{test_struct2});
}

```

### Expected Behavior

```zig
const TestStruct = struct {
    test_var_renamed: i32, //variable was renamed here
};

test "test" {
    const test_struct: TestStruct = .{
        .test_var_renamed = 1, //gets renamed
    };

    const test_struct2: TestStruct = TestStruct{
        .test_var_renamed = 1, //gets renamed
    };

    std.debug.print("{}", .{test_struct});
    std.debug.print("{}", .{test_struct2});
}

```

#398 might be related to this bug

WillLillis:
I'm also running into this issue using `0.13.0-dev.29+bb19bee`. It looks like zls is ignoring the initialization accesses to the struct's fields when it searches for references as part of the renaming request. These references are also missing when using the server's `textDocument/reference` capability. I'm not super familiar with the language/ zls code base yet, but after a bit of poking around it looks like the references are missed here: 

https://github.com/zigtools/zls/blob/93b7bbd0d96ee7088c46d1373a2810e84f7d47be/src/features/references.zig#L95-L135

If I add a `.assign_destructure` arm to the switch block, it looks like that matches up with the missing references to the field initializations. Still working on how to properly add the appropriate information to the references builder, but hopefully this helps narrow down the issue :)

Related #1700

xdBronch:
[...]

Issue #1700, opened 2024-01-06
`textDocument/rename` does not consider anonymous struct literals in return expressions


J-MR-T:
### Zig Version

0.11.0

### Zig Language Server Version

0.11.0

### Client / Code Editor / Extensions

nvim v0.10.0-dev-6405fa4 with nvim-lsp, default capabilities

### Steps to Reproduce and Observed Behavior

Take this snippet:

```zig
fn f() struct{a:i32} {
    return .{.a = 1};
}

fn main() void {
    var x = f();
    x.a = 2;
}
```

Reproduce: Invoke a rename on the `a` in the struct that is returned from `f()`; rename it to `b`.
The line `x.a = 2;` changes to `x.b = 2;`, as expected, but `return .{.a = 1};` does not change.

### Expected Behavior

In the example above, I would expect `return .{.a = 1};` to change to `return .{.b = 1};`.

Trying to track all possible anonymous struct literals, that have been saved and later used as the return value of the function might prove difficult, but I would imagine the simple case of `return .{...};`, that also covers most use cases, should be relatively easy. I'm happy to try to implement it myself and submit a PR, if you agree this is an issue (I would suppose it would have to be implemented here: https://github.com/zigtools/zls/blob/3471da593bc9d979c73b35c516e902c06781310b/src/features/references.zig#L444-L464 ?).

xdBronch:
this works properly now

PR #1903, opened 2024-05-28
fix: references/rename for struct_init_one, struct_init_one_comma

WillLillis:
This fixes some cases where references to fields in some struct initializations aren't picked up, causing features like renaming and reference viewing to fail. Currently the fix only works *one* way, in that a rename invoked on the actual struct initialization will still fail, but *will* work when invoked on other locations such as the struct's definition or a field access. I took some time to try and fix the other direction, but I don't think I'm familiar enough with the codebase to do that yet. I added a test case for this but left it commented out as it currently fails. 

This fix only covers the cases of `.struct_init_one` and `.struct_init_one_comma`. I'd really like to look into fixing the other node cases (e.g. `.struct_init`, `.struct_init_comma`, etc. if I'm understanding things properly) but wanted to cover this first case and get some feedback/ a sanity check before proceeding on. Here's a quick demo showing both renaming and references with the improvements:

zls_rename

Related Issues: #1837, #1700

### DIFF
--- a/src/features/references.zig
+++ b/src/features/references.zig
@@ -100,6 +100,7 @@ const Builder = struct {
         const datas = tree.nodes.items(.data);
         const token_tags = tree.tokens.items(.tag);
         const starts = tree.tokens.items(.start);
+        const main_tokens = tree.nodes.items(.main_token);
 
         switch (node_tags[node]) {
             .identifier,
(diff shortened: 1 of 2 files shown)


WillLillis:
> It should be possible to handle all the other .struct_init_* cases by doing something like this:

Awesome, I'll take a pass at the other cases soon. Thank you! :)

WillLillis:
Thought I ran zig fmt before pushing last night, really sorry about that!

WillLillis:
Looks like it's failing due to the recent deprecations in https://github.com/ziglang/zig/commit/f97c2f28fdc3061bc7e30ccfcafaccbee77993b6, which is fixed in #1904?

Techatrix (approved):
I removed the initial switch case that just handled `.struct_init_one.*` because the other case is already capable of handling all variants. Otherwise, this looks great!

WillLillis:
> I removed the initial switch case that just handled `.struct_init_one.*` because the other case is already capable of handling all variants. Otherwise, this looks great!

Makes sense, thanks so much for the help!

PR #1904, opened 2024-05-29
Fix deprecated `getStderrMutex` usage

kcbanner:
This API was deprecated in https://github.com/ziglang/zig/commit/f97c2f28fdc3061bc7e30ccfcafaccbee77993b6


Techatrix:
The std.Progress rework also broke ZLS's custom build runner. This can be checked by CI when updating the minimum Zig version. 
I will still have to wait for [zig-overlay](https://github.com/mitchellh/zig-overlay) to update before merging this.

PR #1905, opened 2024-05-30
flake: replace dependency on flake-utils with lambda function

Nanotwerp:
Changes also include making the scoping more clear, via more `let` expressions. The flake was reformatted with `nixfmt-rfc-style`, which is the official, `RFC 166`-compliant, formatter for Nix. `nixpkgs-fmt` is replaced by `nixfmt-rfc-style` in this commit. Additionally, the branch of `nixpkgs` is changed from `23.11` to `nixos-unstable`.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -21,24 +21,6 @@
         "systems": "systems"
       },
       "locked": {
-        "lastModified": 1710146030,
-        "narHash": "sha256-SZ5L6eA7HJ/nmkzGG7/ISclqe6oZdOZTNoesiInkXPQ=",
-        "owner": "numtide",
-        "repo": "flake-utils",
-        "rev": "b1d9ab70662946ef0850d488da1c9019f3a9752a",
-        "type": "github"
-      },
-      "original": {
-        "owner": "numtide",
-        "repo": "flake-utils",
-        "type": "github"
-      }
-    },
-    "flake-utils_2": {
-      "inputs": {
-        "systems": "systems_2"
-      },
-      "locked": {
         "lastModified": 1705309234,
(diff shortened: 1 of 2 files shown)


Nanotwerp:
> What is the motivation for moving away from `flake-utils`? What is the motivation for switching from `23.11` to `nixos-unstable`? AFAIK `nixfmt-rfc-style` has not been declared stable. I would rather just wait for its full release as `nixfmt`.

Removing `flake-utils` would reduce the amount of inputs with the light cost of a few lines of boilerplate code.`flake-utils` was only used to create a package attribute set for each system. The benefit of removing an extra input would be better flake lock update speeds and less redundant disk space for having so many versions of `flake-utils`. I think this change is pretty Zig-esque, as that layer of abstraction isn't needed, given how `flake-utils` is essentially only being used for one thing.

I switched to `nixos-unstable` to make it less likely for the formatter to be outdated. Plus, `24.05` is already out, so `23.11` is a bit behind now. Tracking `nixos-unstable` shouldn't pose any issues, given there's only one package to worry about.

I understand your point about `nixfmt-rfc-style`, but it's lack of full stability wouldn't really affect a small flake like this. Some members on the NixOS team have already begun reformatting packages with it. Should I switch back?

Techatrix:
> removing `flake-utils`

All the guides I have seen about nix flakes have used `flake-utils` instead of writing their own implementation. Why should ZLS be special? I don't find the given reason to be enough to justify going against the "status-quo". I doubt this is going to make a reasonable difference to flake lock update speeds and disk space usage.

> switching to `nixfmt-rfc-style`

I would rather just wait for the stable release of `nixfmt`.

> switching to `nixos-unstable`

If we don't use `nixfmt-rfc-style` then there is no worry about it being outdated.

PR #1906, opened 2024-05-31
Bazel support hack

SuperAuguste:


Techatrix:
A global build config is not something that makes a lot of sense since build system state is managed on a per-project basis.
Project specific configuration is possible in ZLS with `zls.build.json` which requires the existence of a `build.zig` to identify the root of the project. But existence of a `build.zig` makes this unnecessary.

Issue #1835, opened 2024-03-21
Base conversion of numeric constants on hover


Beyley:
When you hover over a numeric constant like `0x3a`, `1025`, or `0b1000`, it should auto-convert that in the hover information to base 2, 10, and 16. Ideally this would also apply to `enum` and `union(enum)` fields aswell.

Techatrix:
implemented in #1911

PR #1911, opened 2024-06-04
feat: Integer literal hover support

WillLillis:
This PR implements the base 2, 10, and 16 hover information for integer literals as proposed in #1835.

A few notes:

- I somewhat arbitrarily chose `i256` as the maximum width for parsing integer literals that won't fit within 64 bits as given by `std.zig.parseNumberLiteral()`. Should this be a wider type?

- fmt currently treats negatives a little oddly, but it is being given [some attention](https://github.com/ziglang/zig/issues/20152). Maybe the treatment of negative values here can be cleaned up a bit once some change are made?

- Octal encoding wasn't originally  requested in the issue, but that could be added here as well. Should it?

### DIFF
--- a/src/features/hover.zig
+++ b/src/features/hover.zig
@@ -10,7 +10,6 @@ const tracy = @import("tracy");
 
 const Analyser = @import("../analysis.zig");
 const DocumentStore = @import("../DocumentStore.zig");
-const ParseError = @import("../Server.zig").Error.ParseError;
 
 const data = @import("version_data");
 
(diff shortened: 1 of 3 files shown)


src/features/hover.zig
```diff
@@ -345,6 +346,77 @@ fn hoverDefinitionFieldAccess(
...
+    token_index: Ast.TokenIndex,
+    arena: std.mem.Allocator,
+    markup_kind: types.MarkupKind,
+) error{ OutOfMemory, ParseError }!?[]const u8 {
+    const tree = handle.tree;
+    // number literals get tokenized separately from their minus sign
+    const is_negative = tree.tokens.items(.tag)[token_index -| 1] == .minus;
+    const num_slice = tree.tokenSlice(token_index);
+    const parsed = std.zig.parseNumberLiteral(num_slice);
+    // fmt doesn't allow floats to be formatted with binary or hex specifiers
```
Techatrix, line 360: You would need to `@bitcast` the float to a unsigned integer with the same number of bits like `@bitcast(u32, some_f32)` and then format it.

This is not something that would be very useful since we don't know which float type is being used.

Techatrix (changes requested):
> I somewhat arbitrarily chose i256 as the maximum width for parsing integer literals that won't fit within 64 bits as given by std.zig.parseNumberLiteral(). Should this be a wider type?

`i256` should be avoided since `std.math.big.int.Const` already has formatting functions for this use case. I am not sure what the use case for supporting big ints would be. I would rather suggest to drop support for `.big_int` to keep it simple.

> Octal encoding wasn't originally requested in the issue, but that could be added here as well. Should it?

yes

---

The used hover text doesn't mention what format it is showcasing. Here are some proposals. I would personally go for Option 1a and 1b but I'd be happy to discuss any alternatives.

# Option 1a (with markdown)

|     |            |
| --- | ---------- |
| BIN | 0b00001100 |
| OCT | 0o14        |
| DEC | 12         |
| HEX | 0xC        |

# Option 2a (with markdown)

| BIN        | OCT | DEC | HEX |
| ---------- | --- | --- | --- |
| 0b00001100 | 0o14 | 12  | 0xC |

# Option 3a (with markdown)

```
BIN: 0b00001100
OCT: 0o14
DEC: 12
HEX: 0xC
```

# Option 1b (without markdown)

BIN: 0b00001100
OCT: 0o12
DEC: 12
HEX: 0xC

# Option 2b (without markdown)

BIN | 0b00001100
OCT | 0o12
DEC | 12
HEX | 0xC

# Option 3b (without markdown)

BIN        | OCT | DEC | HEX
0b00001100 | 0o12 | 12  | 0xC

src/features/hover.zig
WillLillis: Gotcha, that makes sense. Should the comment be removed then?

Techatrix: The comment and the if below can be removed. Don't forget to adjust the switch case for .float and .failure to return null instead of unreachable.

Techatrix:
> I ran into some trouble getting the table to render correctly, but I believe it's a neovim-specific issue if I'm reading [this](https://github.com/nvimdev/lspsaga.nvim/issues/1012) correctly.

I just tested [Helix](https://helix-editor.com/) out of curiosity and they also have this problem:


Looks like markdown tables are too advanced for these editors. So we either switch to a style that doesn't use tables or stick with them but try to improve the layout similar to this:


Doesn't look to bad I guess...

Techatrix:
I've made some improvements to this PR:

- give the markdown table consistent spacing
- add the table column names to give the table more spacing
- add a test case for plaintext markup kind
- some code refactoring

WillLillis:
Thanks for the feedback and help!

Techatrix (approved):
Thank You! Very `base`d

Issue #1706, opened 2024-01-13
Renaming a local variable should be instant. 


matklad:
If I use Zls to rename a _local_ variable inside a single function, there's a noticable delay. Reading the code:

https://github.com/zigtools/zls/blob/abe83cf2291381291d4c20629a5018f124c4bed2/src/features/references.zig#L433-L441

it seems that the issue is that Zls tries to find references to this local variable in every project file. 

This is needless work, due to the way Zig is structured, it often is possible to greatly restrict the search scope looking at the kind of decl. 

For example, local variables should be looked up only in the current function. Anything that's not `pub` (fields are implicitly `pub`) can be restricted to a single file. 

Implementation wise, a good pattern that works for symbol search is to provide a _scope_ parameter, which is a set of files (or even ranges within a file), where the search should be performed. In general, two scopes are in play:

* Scope passed by the caller tells which references they are interested in (eg, highlihting only looks for references in the same file)
* Scope derived from the decl conservatively estimates the area where references could _theoretically_ be found. 

Then, we can _intersect_ both scope and only iterate what's relevant in

https://github.com/zigtools/zls/blob/abe83cf2291381291d4c20629a5018f124c4bed2/src/features/references.zig#L154

matklad:
I believe this is fixed these days? Been instant for me for a long time, thanks!

PR #1917, opened 2024-06-13
perf: Improve local variable reference finding

WillLillis:
This PR improves the time to find references and rename local variables. As suggested by @Techatrix in the discord, for certain cases the search for references can be limited to the current document rather than the entire project. After some brief testing with function-local variables in the zls project I saw a typical 4,500ms to <10ms improvement with debug builds. Testing with some of the larger files in zig itself, I saw a typical 20,000ms to 2,500ms improvement. Greater improvements are seen when using a `ReleaseSafe` build. I can spend some time gathering more numbers if that would be helpful :)

Related Issue: #1706

### DIFF
--- a/src/features/references.zig
+++ b/src/features/references.zig
@@ -3,6 +3,7 @@ const Ast = std.zig.Ast;
 const log = std.log.scoped(.zls_references);
 
 const Server = @import("../Server.zig");
+const DocumentScope = @import("../DocumentScope.zig");
 const DocumentStore = @import("../DocumentStore.zig");
 const Analyser = @import("../analysis.zig");
 const types = @import("../lsp.zig");
@@ -458,6 +459,21 @@ pub fn referencesHandler(server: *Server, arena: std.mem.Allocator, request: Gen
         else => true,
     };
 
+    // highlight requests only pertain to the current document, otherwise we can try to narrow things down
+    const workspace = if (request == .highlight) false else blk: {
+        const scope_index = Analyser.innermostBlockScopeIndex(try handle.getDocumentScope(), source_index);
+        if (scope_index.unwrap()) |idx| {
+            const scope_tag = DocumentScope.getScopeTag(try handle.getDocumentScope(), idx);
+            switch (scope_tag) {
+                .function, .block => {
+                    break :blk false; // scan only the current document
+                },
+                else => {},
+            }
+        }
+        break :blk true;
+    };
+
     const locations = if (decl.decl == .label)
         try labelReferences(arena, decl, server.offset_encoding, include_decl)
     else
@@ -468,7 +484,7 @@ pub fn referencesHandler(server: *Server, arena: std.mem.Allocator, request: Gen
             server.offset_encoding,
             include_decl,
             server.config.skip_std_references,
-            request != .highlight, // scan the entire workspace except for highlight
+            workspace,
         );
 
     switch (request) {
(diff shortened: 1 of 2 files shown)


WillLillis:
Took another pass over this and sped up private container-scoped decls as well. In my own testing this didn't break anything/ cause any refs to be missed, but would appreciate a second look-over.

Techatrix (changes requested):
After taking a closer look at the references code, I realized that this extra logic should be moved [here](https://github.com/zigtools/zls/blob/b4f4d35354f39fd190b9ce681c75affafaa0a1ab/src/features/references.zig#L239) since its the only place where `workspace` is being used. This makes it clear that we only need to handle the case where the declaration is an `.ast_node`. Otherwise the logic looks fine.

---

The code looked a bit different from what I would have written so I left some suggestion on how to write it in a more "idiomatic" style. In the end, this is probably what I would have written:
```zig
const workspace = if (request == .highlight) false else blk: {
    const doc_scope = try handle.getDocumentScope();
    const scope_index = Analyser.innermostBlockScopeIndex(doc_scope, source_index).unwrap() orelse break :blk true;
    break :blk switch (doc_scope.getScopeTag(scope_index)) {
        .function, .block => false,
        .container, .container_usingnamespace => decl.isPublic(),
        .other => true,
    };
};
```
Hope that helps :shrug:

WillLillis:
> After taking a closer look at the references code, I realized that this extra logic should be moved [here](https://github.com/zigtools/zls/blob/b4f4d35354f39fd190b9ce681c75affafaa0a1ab/src/features/references.zig#L239) since its the only place where `workspace` is being used. This makes it clear that we only need to handle the case where the declaration is an `.ast_node`. Otherwise the logic looks fine.
> 
> The code looked a bit different from what I would have written so I left some suggestion on how to write it in a more "idiomatic" style. In the end, this is probably what I would have written:
> 
> ```zig
> const workspace = if (request == .highlight) false else blk: {
>     const doc_scope = try handle.getDocumentScope();
>     const scope_index = Analyser.innermostBlockScopeIndex(doc_scope, source_index).unwrap() orelse break :blk true;
>     break :blk switch (doc_scope.getScopeTag(scope_index)) {
>         .function, .block => false,
>         .container, .container_usingnamespace => decl.isPublic(),
>         .other => true,
>     };
> };
> ```
> 
> Hope that helps 🤷

That was really helpful, thank you! Getting this kind of feedback is always really appreciated so that so my cide can be cleaner moving forward :). I think I'll do another read through the language reference to brush up as well.

src/features/references.zig
```diff
@@ -236,6 +236,18 @@ fn symbolReferences(
     switch (decl_handle.decl) {
         .ast_node => {
             try builder.collectReferences(curr_handle, 0);
+
+            const source_index = offsets.positionToIndex(curr_handle.tree.source, request.position(), encoding);
```
Techatrix, line 240: The scope should be looked up where the declaration is located and not where the cursor is. The symbol/decl may be in a different file that where the request was initiated.
```suggestion
            const source_index = offsets.tokenToIndex(decl_handle.handle.tree, decl_handle.nameToken());
```

Here is an example that can cause a crash:
```zig
// A.zig
pub const Foo = struct {};
```
```zig
// B.zig
const A = @import("A.zig");
fn foo() void {
    _ = A.Foo;
    //     ^ find all references here
}
```

Techatrix (changes requested):

Techatrix (approved):

PR #1920, opened 2024-06-14
Sentinel hover fix

### LINKED ISSUE #1840: Inlay hints/hover: Missing sentinel value
### Zig Version

0.12.0-dev.3451+405502286

0.12.0-dev.496+96eddd06

vscode

```zig
test {
    const a: [:0] i1 = undefined;
    const b = a;
    _ = b;
}
```

inlay_hint_hover_missing_sentinel_value

The above example isn't that severe as hovering over `b` will show `const a: [:0] i1 = undefined;`

Example 2:
[Missing sentinel value.webm](https://github.com/zigtools/zls/assets/16590917/6bf41ff5-f980-4e50-8e66-e8e8a33b6717)

### Expected Behavior

It matters

### Relevant log output

_No response_

WillLillis:
This PR fixes sentinel values not appearing on hover. The code is almost entirely @llogick's solution provided in [this comment](https://github.com/zigtools/zls/issues/1840#issuecomment-2156306928) from #1840, with tests added and some small changes for `.Range` accesses.

Fixes #1840

### DIFF
--- a/src/analysis.zig
+++ b/src/analysis.zig
@@ -1173,29 +1173,21 @@ fn allDigits(str: []const u8) bool {
     return true;
 }
 
-const ResolveIntegerLiteralKind = enum {
-    length,
-    sentinel,
-};
-
-fn resolveIntegerLiteral(analyser: *Analyser, node_handle: NodeWithHandle, kind: ResolveIntegerLiteralKind) !?u64 {
-    // When resolve_number_literal_values is set then resolveTypeOfNode will also resolve the value of number literals.
-    // So we can use it to resolve integer values.
-
+fn resolveInternPoolValue(analyser: *Analyser, node_handle: NodeWithHandle) error{OutOfMemory}!?InternPool.Index {
     const old_resolve_number_literal_values = analyser.resolve_number_literal_values;
     analyser.resolve_number_literal_values = true;
     defer analyser.resolve_number_literal_values = old_resolve_number_literal_values;
 
     const resolved_length = try analyser.resolveTypeOfNode(node_handle) orelse return null;
     switch (resolved_length.data) {
-        .ip_index => |payload| {
-            return switch (kind) {
-                .length => analyser.ip.toInt(payload.index, u64),
-                .sentinel => @intFromEnum(payload.index),
-            };
-        },
+        .ip_index => |payload| return payload.index,
         else => return null,
     }
+}
+
+fn resolveIntegerLiteral(analyser: *Analyser, comptime T: type, node_handle: NodeWithHandle) error{OutOfMemory}!?T {
+    const ip_index = try analyser.resolveInternPoolValue(node_handle) orelse return null;
+    return analyser.ip.toInt(ip_index, T);
 }
 
 const primitives = std.StaticStringMap(InternPool.Index).initComptime(.{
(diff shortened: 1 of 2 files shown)


src/analysis.zig
```diff
@@ -1515,12 +1603,34 @@ fn resolveTypeOfNodeUncached(analyser: *Analyser, node_handle: NodeWithHandle) e
...
+                // resolveTypeOfNode can also resolve values that returned as indices into the InternPool.
+                const old_resolve_number_literal_values = analyser.resolve_number_literal_values;
+                analyser.resolve_number_literal_values = true;
+                defer analyser.resolve_number_literal_values = old_resolve_number_literal_values;
+                const sentinel = try analyser.resolveTypeOfNode(.{ .node = ptr_info.ast.sentinel, .handle = handle }) orelse break :blk .none;
+                break :blk switch (sentinel.data) {
+                    .ip_index => |payload| payload.index,
+                    else => .none,
+                };
+            } else .none;
```
Techatrix, line 1616: This should be moved into its own function (name suggestion: `resolveInternPoolValue`) because this exact code appears twice in this file.
The `resolveIntegerLiteral` can then call that function.

Techatrix (changes requested):

src/analysis.zig
WillLillis: Looking at this again, `resolveIntegerLiteral` and the code above are very similar. I originally moved the logic into a separate `resolveInternPoolValue` function, but rethought things since they only really differ at the return point (`analyser.ip.toInt(payload.index, u64)` vs. `payload.index`)

I tried adding a third argument to `resolveIntegerLiteral` avoid duplicating the common logic instead. Happy to rework it if this isn't the right approach!

WillLillis: On second thought, would a wrapper here improve things? 

```diff
diff --git a/src/analysis.zig b/src/analysis.zig
index 69158c7..267bc39 100644
--- a/src/analysis.zig
+++ b/src/analysis.zig
@@ -1198,6 +1198,13 @@ fn resolveIntegerLiteral(analyser: *Analyser, node_handle: NodeWithHandle, kind:
     }
 }

+fn resolveInternPoolValue(analyser: *Analyser, node_handle: NodeWithHandle) !InternPool.Index {
+    if (try resolveIntegerLiteral(analyser, node_handle, .sentinel)) |sentinel| {
+        return @enumFromInt(sentinel);
+    }
+    return .none;
+}
+
 const primitives = std.StaticStringMap(InternPool.Index).initComptime(.{
     .{ "anyerror", .anyerror_type },
     .{ "anyframe", .anyframe_type },
@@ -1613,12 +1620,7 @@ fn resolveTypeOfNodeUncached(analyser: *Analyser, node_handle: NodeWithHandle) e
         => {
             const ptr_info = ast.fullPtrType(tree, node).?;

-            const sentinel: InternPool.Index = blk: {
-                if (try analyser.resolveIntegerLiteral(.{ .node = ptr_info.ast.sentinel, .handle = handle }, .sentinel)) |sent| {
-                    break :blk @enumFromInt(sent);
-                }
-                break :blk .none;
-            };
+            const sentinel = try resolveInternPoolValue(analyser, .{.node = ptr_info.ast.sentinel, .handle = handle});

             const elem_ty = try analyser.resolveTypeOfNodeInternal(.{ .node = ptr_info.ast.child_type, .handle = handle }) orelse return null;
             if (!elem_ty.is_type_val) return null;
@@ -1644,12 +1646,7 @@ fn resolveTypeOfNodeUncached(analyser: *Analyser, node_handle: NodeWithHandle) e

             const elem_count: ?u64 = try analyser.resolveIntegerLiteral(.{ .node = array_info.ast.elem_count, .handle = handle }, .length);

-            const sentinel: InternPool.Index = blk: {
-                if (try analyser.resolveIntegerLiteral(.{ .node = array_info.ast.sentinel, .handle = handle }, .sentinel)) |s| {
-                    break :blk @enumFromInt(s);
-                }
-                break :blk .none;
-            };
+            const sentinel = try resolveInternPoolValue(analyser, .{.node = array_info.ast.sentinel, .handle = handle});

             const elem_ty = try analyser.resolveTypeOfNodeInternal(.{ .node = array_info.ast.elem_type, .handle = handle }) orelse return null;
             if (!elem_ty.is_type_val) return null;
```

Techatrix: I would rather expect `resolveIntegerLiteral` to call `resolveInternPoolValue` and not the other way around. I made the change myself and pushed it to your PR.

WillLillis: Oh I see what you meant now, thank you!

Techatrix (approved):

PR #1934, opened 2024-06-27
Support initialization options

gohryt:
This is needed because of some text editors that does't have support of sending options via i_haz_configuration


gohryt:
Personally i have this error in zed but i also find that it was in some terminal based editors.

Techatrix:
Good to know that Zed uses initializationOptions. Thanks

Techatrix (approved):

Techatrix:
I noticed that the config options where never de-allocated. The `initialization_options` would also be ignored if a `zls.json` was found. I pushed a fix to this PR.

PR #1935, opened 2024-06-30
Inconsistent use of case on the index

Yewzir:
I think it's highly preferable to be consistent in how things are written in terms of what the eye gets, especially on the index page where new people usually look around for what's in store and how it might benefit them. Though it may seem like a minor detail, you'd be surprised how often inconsistencies like these catch the eye. It also happens to be the abbreviation of what this repository's all about and if you use it uppercase two times here and three times there, it's always better to do it everywhere, except for calling commands.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,10 +37,10 @@ zig build -Doptimize=ReleaseSafe
 
 ### Configuration Options
 
-You can configure zls by editing your `zls.json` configuration file.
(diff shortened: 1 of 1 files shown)


Techatrix (approved):
I though I updated all mentions of lowercase zls in 16317d40abbea111b4479abdfca256bac364bb3f. Thanks for the correction

PR #1938, opened 2024-07-03
Fix absolute path detection for build on save diagnostics

jayrod246:
This patch detects drive letters.

I was trying to trigger some diagnostics using std.fmt. For awhile I couldn't tell if anything was working at all. But after a debug session, I figured out that the ':' following a drive letter was causing the line iterator to bail.

After this change I was able to see this:
screenshot

### DIFF
--- a/src/features/diagnostics.zig
+++ b/src/features/diagnostics.zig
@@ -288,7 +288,11 @@ pub fn generateBuildOnSaveDiagnostics(
 
         const src_path = pos_and_diag_iterator.next() orelse continue;
         const absolute_src_path = if (std.fs.path.isAbsolute(src_path)) src_path else blk: {
-            const absolute_src_path = std.fs.path.join(arena, &.{ workspace_path, src_path }) catch continue;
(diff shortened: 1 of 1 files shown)


Rexicon226 (approved):

Techatrix (approved):
This check is a bit hacky but shouldn't cause any problems. I anticipate the build on save diagnostic system to be reworked anyway.

jayrod246:
Yeah I anticipated a rework when I was playing around with this.

I experimented with writing an actual parser for Zig compiler output, that tries to notice `note:`s for example that may appear on lines that follow a compile error. But this worked to get most diagnostics to show up for me.

Techatrix:
The `:` can be very annoying. Especially because it can be part of a valid filename like `/somewhere/:.zig:2:5: error: ...`
I would say that the best way forward is to wait for integration with Zig's compiler protocol which can send errors as `std.zig.ErrorBundle`. They don't require any parser which makes them far easier to deal with.

PR #1948, opened 2024-07-11
Resolve `(try foo())` (works when nested too)

llogick:


Techatrix (changes requested):
If it works when nested too, could you add test cases for that?

Techatrix (approved):

PR #1979, opened 2024-07-28
functions inside namespaces can't have self parameters

Techatrix:
```zig
// before: 'foo' is a function
// after:  'foo' is a function
fn foo() void {}

// before: 'bar' is a method
// after:  'bar' is a function
fn bar(_: @This()) void {}

// before: 'baz' is a method
// after:  'baz' is a function
fn baz(_: anytype) void {}
```

```zig
some_field: u32,
// before: 'foo' is a function
// after:  'foo' is a function
fn foo() void {}

// before: 'bar' is a method
// after:  'bar' is a method
fn bar(_: @This()) void {}

// before: 'baz' is a method
// after:  'baz' is a method
fn baz(_: anytype) void {}
```

### DIFF
--- a/src/analysis.zig
+++ b/src/analysis.zig
@@ -352,6 +352,7 @@ pub fn hasSelfParam(analyser: *Analyser, func_ty: Type) error{OutOfMemory}!bool 
     const fn_token = func_node.handle.tree.nodes.items(.main_token)[func_node.node];
     const in_container = try innermostContainer(func_node.handle, func_node.handle.tree.tokens.items(.start)[fn_token]);
     std.debug.assert(in_container.is_type_val);
+    if (in_container.isNamespace()) return false;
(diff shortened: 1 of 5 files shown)


InKryption:
This seems like an odd change, functions with the first parameter as `anytype` can be, and in fact are treated as methods - see `std.BoundedArray(T, len).slice`

Techatrix:
> This seems like an odd change, functions with the first parameter as `anytype` can be, and in fact are treated as methods - see `std.BoundedArray(T, len).slice`

This PR doesn't change that. the returned struct in `std.BoundedArray` is not a namespace since it has fields like `buffer` and `len`.

arrufat:
Yes, I can confirm that, just tried that code with this PR applied:
image

PR #1982, opened 2024-07-30
resolve type of comparison op to @Vector(N, bool) if lhs is a vector

janis-bhm:
zls incorrectly always assumes that comparison operations resolve to `.bool_type`, even if you compare two `@Vector`s

expected:
```
    try testInlayHints(
        \\const a<@Vector(2,u8)> = @Vector(2, u8){1,2};
        \\const foo<@Vector(2,bool)> = a == a;
    , .{ .kind = .Type });
```
actual:
```
    try testInlayHints(
        \\const a<@Vector(2,u8)> = @Vector(2, u8){1,2};
        \\const foo<bool> = a == a;
    , .{ .kind = .Type });
```

I tried solving this but I'm not sure if this is the right way to do it, here's some thoughts I had:
- also check `rhs`
- maybe instead of `orelse return null;` return `.bool_type` like before

### DIFF
--- a/src/analysis.zig
+++ b/src/analysis.zig
@@ -2071,6 +2071,28 @@ fn resolveTypeOfNodeUncached(analyser: *Analyser, node_handle: NodeWithHandle) e
         .greater_than,
         .less_or_equal,
         .greater_or_equal,
+        => {
+            const ty = try analyser.resolveTypeOfNodeInternal(
+                .{ .node = datas[node].lhs, .handle = handle },
+            ) orelse return null;
+            const typeof = ty.typeOf(analyser);
+
+            if (typeof.data == .ip_index) {
+                const key = analyser.ip.indexToKey(typeof.data.ip_index.index);
+                if (key == .vector_type) {
+                    const vector_ty_ip_index = try analyser.ip.get(analyser.gpa, .{
+                        .vector_type = .{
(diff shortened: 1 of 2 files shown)


src/analysis.zig
```diff
@@ -2071,6 +2071,28 @@ fn resolveTypeOfNodeUncached(analyser: *Analyser, node_handle: NodeWithHandle) e
         .greater_than,
         .less_or_equal,
         .greater_or_equal,
+        => {
+            const ty = try analyser.resolveTypeOfNodeInternal(
+                .{ .node = datas[node].lhs, .handle = handle },
+            ) orelse return null;
```
Techatrix, line 2077: > I tried solving this but I'm not sure if this is the right way to do it, here's some thoughts I had:
> 
>     * also check `rhs`
> 
>     * maybe instead of `orelse return null;` return `.bool_type` like before

Falling back to returning bool seems reasonable.
```suggestion
            ) orelse return try Type.typeValFromIP(analyser, .bool_type);
```
I think we can skip checking the `rhs`, the most common cases should already be handled without it and we save some time that would be spent on analyzing the extra expression.

Techatrix (changes requested):

Techatrix (approved):
thank you

PR #2000, opened 2024-08-16
commit langref.html.in to source control instead of downloading it

Techatrix:
#1409 was mistake

 #2000 :tada:


nektro:
out of curiosity, why do you feel it was a mistake?

Techatrix:
> out of curiosity, why do you feel it was a mistake?

- various package distributions needed workarounds because the build phase can't just randomly access the internet
- the downloaded langref file is not properly cached which means that you can't build ZLS on the next day without internet
- the downloaded langref file is not pinned by commit hash which hurts reproducibility 

https://github.com/ziglang/zig/issues/17895 is the solution.

nektro:
ah makes sense, i mistakenly assumed 17895 was already implemented

BratishkaErik:
Thank you very much from GURU maintainer!

bmwiedemann:
Is there a plan for a release with this fix? How far away would that approximately be?

nektro:
releases are paired with zig releases so there won't one until 0.14. using zls master requires building from source or zigup

PR #2015, opened 2024-08-29
fix completion for branches without braces

### LINKED ISSUE #2014: no completions after branches with no braces
### Zig Version

0.14.0-dev.1349+6a21875dd

### ZLS Version

0.14.0-dev.135+b481aae (with changes to make it compile on master)

_No response_

```zig
pub fn main() void {
    if (true) <cursor>
}
```
ask for completions and nothing happens, same with `while`.

### Expected Behavior

completions

### Relevant log output

_No response_

xdBronch:
fixes #2014 
thanks @llogick :upside_down_face:

### DIFF
--- a/src/analysis.zig
+++ b/src/analysis.zig
@@ -3473,7 +3473,11 @@ pub fn getPositionContext(
                     curr_ctx.ctx = .{ .string_literal = tok.loc };
                 },
                 .identifier => switch (curr_ctx.ctx) {
-                    .empty, .pre_label, .var_access => curr_ctx.ctx = .{ .var_access = tok.loc },
+                    .empty,
+                    .pre_label,
+                    .var_access,
+                    .parens_expr,
+                    => curr_ctx.ctx = .{ .var_access = tok.loc },
                     .label => |filled| if (!filled) {
                         curr_ctx.ctx = .{ .label = true };
                     } else {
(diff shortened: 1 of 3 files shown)


src/features/completions.zig
```diff
@@ -876,7 +876,7 @@ pub fn completionAtIndex(
 
     switch (pos_context) {
         .builtin => try completeBuiltin(&builder),
-        .var_access, .empty => try completeGlobal(&builder),
+        .var_access, .empty, .parens_expr => try completeGlobal(&builder),
```
Techatrix, line 879: According to the following [comment](https://github.com/zigtools/zls/blob/18c1c3947d52f6db55118731621d724a799481da/src/analysis.zig#L3282), `parens_expr` is only supposed to be used internally.

`parens_expr` should be converted to `var_access` [here](https://github.com/zigtools/zls/blob/18c1c3947d52f6db55118731621d724a799481da/src/analysis.zig#L3557-L3565).

src/analysis.zig
```diff
@@ -3482,6 +3482,7 @@ pub fn getPositionContext(
                     .enum_literal => curr_ctx.ctx = .{
                         .enum_literal = tokenLocAppend(curr_ctx.ctx.loc().?, tok),
                     },
+                    .parens_expr => curr_ctx.ctx = .{ .var_access = tok.loc },
```
Techatrix, line 3485: There already exists a switch case for this:
```zig
.empty, .pre_label, .var_access => curr_ctx.ctx = .{ .var_access = tok.loc },
```
You can add `parens_expr` there instead of duplicating it.

Techatrix (changes requested):

Techatrix (approved):
:+1:

PR #2011, opened 2024-08-27
Update Zig master `@branchHint` and `builtin` type fields

mochalins:

PR #2019, opened 2024-08-31
Update the build runner

llogick:
Use `std.meta.declarations` for compatibility with latest and previous, supported, Zig versions


Techatrix (approved):
This got me confused for a moment.
This is one of the rare locations in the build runner code that are not covered by the build_runner.yml workflow. This is why CI on #2011 passed.

PR #2024, opened 2024-09-10
add snippets for format/panic/log

xdBronch:
zigs formatting is real ugly for the multiline strings here xd

### DIFF
--- a/src/snippets.zig
+++ b/src/snippets.zig
@@ -21,7 +21,14 @@ pub const top_level_decl_data = [_]Snipped{
     .{ .label = "union tagged", .kind = .Snippet, .text = "const $1 = union(${2:enum}) {$0};" },
     .{ .label = "test", .kind = .Snippet, .text = "test \"$1\" {$0}" },
     .{ .label = "main", .kind = .Snippet, .text = "pub fn main() !void {$0}" },
-    .{ .label = "std_options", .kind = .Snippet, .text = "pub const std_options: std.Options = .{${0}};" },
+    .{ .label = "std_options", .kind = .Snippet, .text = "pub const std_options: std.Options = .{$0};" },
+    .{ .label = "panic", .kind = .Snippet, .text = 
+    \\pub fn panic(
+    \\    msg: []const u8,
+    \\    trace: ?*std.builtin.StackTrace,
+    \\    ret_addr: ?usize,
+    \\) noreturn {$0}
+    },
 };
 
 pub const generic = [_]Snipped{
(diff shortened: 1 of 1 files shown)


src/snippets.zig
```diff
@@ -85,6 +100,14 @@ pub const generic = [_]Snipped{
     .{ .label = "log warn", .kind = .Snippet, .text = "std.log.warn(\"$1\", .{$0});" },
     .{ .label = "log info", .kind = .Snippet, .text = "std.log.info(\"$1\", .{$0});" },
     .{ .label = "log debug", .kind = .Snippet, .text = "std.log.debug(\"$1\", .{$0});" },
+    .{ .label = "format", .kind = .Snippet, .text = 
+    \\pub fn format(
+    \\    self: ${1:@This()},
```
Techatrix, line 105: Naming something `self` should be avoided if possible.
```suggestion
    \\    ${1:self}: ${2:@This()},
```

src/snippets.zig
```diff
@@ -21,7 +21,22 @@ pub const top_level_decl_data = [_]Snipped{
     .{ .label = "union tagged", .kind = .Snippet, .text = "const $1 = union(${2:enum}) {$0};" },
     .{ .label = "test", .kind = .Snippet, .text = "test \"$1\" {$0}" },
     .{ .label = "main", .kind = .Snippet, .text = "pub fn main() !void {$0}" },
-    .{ .label = "std_options", .kind = .Snippet, .text = "pub const std_options: std.Options = .{${0}};" },
+    .{ .label = "std_options", .kind = .Snippet, .text = "pub const std_options: std.Options = .{$0};" },
+    .{ .label = "logFn", .kind = .Snippet, .text = 
+    \\fn logFn(
```
Techatrix, line 26: In contrast to all other top-level-snippets like `main`, `std_options`, `panic`, `root`, this one requires being "hooked up" to `std_options`. I could foresee someone using this snippet and wondering why doesn't do anything on its own.

I would suggest to do one of the following:
- remove this snippet
- give the snippet a doc comment that explains what need to be done for it to work
- give the LSP completion item some documentation that explains what need to be done for it to work

If you have another idea to make sure that this snippet is understood correctly by the user, feel free to suggest an alternative.

Techatrix (changes requested):

src/snippets.zig
xdBronch: do you have any preference? 
i wouldnt mind just removing it, i dont find myself overriding the logger often
a doc comment i imagine could be a little noisy since its mostly just communicating to the snippet user rather than actually documenting code. 
not sure how to do the last one

xdBronch: yeah i think im gonna just settle for that :sweat_smile:, not a terribly important snippet i dont think

Techatrix (approved):

PR #2030, opened 2024-09-15
Fix potential build.zig search

PanSashko:
The existing implementation does not find a build.zig in the same directory with the source file.


Techatrix (changes requested):
The code looks good, i just had one small issue with one of the comments.

Techatrix (approved):
very nice

PR #2032, opened 2024-09-15
WIP: fix: multiline string literals

mnemnion:
Fix for Zig #21358, #21360, which change tokenizing of multiline string literals to not include newlines.  See [ziglang/zig#21371](https://github.com/ziglang/zig/pull/21371/).

This just changes the tokenizer, ZLS no longer SIGKILLs immediately, the integer underflow is gone from the tests, but several tests don't pass (which is to be expected).

I wanted to get some eyes on the fix, since I'm unfamiliar with the codebase, before trying to modify the tests to pass.  Changing tests to pass a newly introduced bug would be unfriendly, after all.

Also, the fix includes doc comments, I don't think that affects ZLS the same way, but I could be wrong and it seemed worth mentioning.

### DIFF
--- a/src/analysis.zig
+++ b/src/analysis.zig
@@ -2064,10 +2064,9 @@ fn resolveTypeOfNodeUncached(analyser: *Analyser, node_handle: NodeWithHandle) e
 
             var length: u64 = 0;
 
-            for (start..end + 1, 0..) |token_index, i| {
+            for (start..end + 1) |token_index| {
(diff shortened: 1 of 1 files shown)


Techatrix:
I think you missed #2031.

mnemnion:
Yes, I was on a commit from yesterday (last week actually dd78968d4c8d),  it looks like you got the fix in right when I started looking into it.  

Since there's no need for this branch, I'll close it.

Issue #1939, opened 2024-07-04
Inconsistent completion of function parentheses when snippets are disabled


castholm:
### Zig Version

0.13.0

### Zig Language Server Version

0.13.0

### Client / Code Editor / Extensions

vscode-zig

### Steps to Reproduce and Observed Behavior

I have `zig.zls.enableSnippets` set to `false` in the settings. Despite this, ZLS completes function call parentheses differently depending on the signature, sometimes inserting just the function name and other times inserting the function name followed by `()`. This inconsistent behavior is undesirable and hurts muscle memory. I disabled snippets because I prefer just completing the names without any automatic insertion of parentheses.

```zig
fn functionZero() void {}

fn functionOne(a: i32) void {
    _ = a;
}

fn functionAnytype(a: anytype) void {
    _ = a;
}

const Foo = struct {
    fn method(foo: Foo) void {
        _ = foo;
    }
};

pub fn main() void {
    const foo: Foo = .{};
    _ = &foo;

    // Below, '_' represents the position of the caret

    // "funZ_"
    // <tab>
    // "functionZero()_"

    // "funO_"
    // <tab>
    // "functionOne_"

    // "funA_"
    // <tab>
    // "functionAnytype()_"

    // "foo.m_"
    // <tab>
    // "foo.method()_"
}
```

Of note is that when snippets are enabled, completion of the `anytype` function places the caret after the inserted `()` and not inbetween, so this is a bug regardless of whether the user has enabled snippets.

### Expected Behavior

When snippets are disabled, only the name should be inserted. The parentheses should never be inserted regardless of whether the call takes any arguments or not.

I fixed this problem for myself locally by making the following changes to `completions.zig`:

```diff
diff --git a/src/features/completions.zig b/src/features/completions.zig
index 599f781..9cc4ff9 100644
--- a/src/features/completions.zig
+++ b/src/features/completions.zig
@@ -329,7 +329,8 @@ fn functionTypeCompletion(
     const new_text = switch (new_text_format) {
         .only_name => func_name,
         .snippet => blk: {
-            if (use_snippets and builder.server.config.enable_argument_placeholders) {
+            if (!use_snippets) break :blk func_name;
+            if (builder.server.config.enable_argument_placeholders) {
                 break :blk try std.fmt.allocPrint(builder.arena, "{}", .{Analyser.fmtFunction(.{
                     .fn_proto = func,
                     .tree = &tree,
[...]

PR #2037, opened 2024-09-22
Add option to opt-out of closing argument parameter brace when completing a function

Nathan-Franck:
Still creates the first open parameter brace

This helps me when I want to wrap an existing value in a new function, since I tend to program inside-out, and the closing brace was creating a lot of friction.

I don't know if maybe this setting should just be grouped with enable_argument_placeholders somehow, since they conflict with each other when both enabled.

### DIFF
--- a/src/Config.zig
+++ b/src/Config.zig
@@ -10,6 +10,9 @@ enable_snippets: bool = true,
 /// Whether to enable function argument placeholder completions
 enable_argument_placeholders: bool = true,
 
+/// Whether to automatically close parameter braces
+enable_auto_close_parameter_braces: bool = true,
+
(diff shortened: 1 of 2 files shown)


Techatrix:
I would like to avoid the adding of a separate config option and avoid a text edit that adds unclosed parentheses.
I rather prefer to have `enableSnippets == false` never add parentheses similar to what has been proposed in #1939. You can then control the opening and closing parentheses manually if you prefer.

Nathan-Franck:
Yeah his change with enable_snippets: false works just fine! Personally adding the opening bracket is totally fine, mostly I'm just looking to have the brackets not get in the way.

PR #2038, opened 2024-09-24
Replace assert with logic to avoid panics

gohryt:
There were panic on this strings when using Zed. I don't really sure it's correct fix, but it works for me.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -415,7 +415,7 @@ fn release(b: *Build, target_queries: []const std.Target.Query, release_artifact
             compress_cmd.step.max_rss = switch (extension) {
                 .zip => 160 * 1024 * 1024, // 160 MiB
                 .@"tar.xz" => 256 * 1024 * 1024, // 256 MiB
-                .@"tar.gz" => 8 * 1024 * 1024, // 8 MiB
+                .@"tar.gz" => 16 * 1024 * 1024, // 12 MiB
             };
             switch (extension) {
                 .zip => {


Techatrix:
The root cause of these invalid inputs need to be identified. If they come from Zed directly then it needs to stop doing that. We can't just try to "patch" bad input. The assumption that the start is before the end is made all over the codebase.

ZLS should also reject invalid input from the client/editor but that is not going to fully resolve the problem. It's the difference between crashing or not providing a response to the client/editor. Both cases will appear as if the ZLS is not working correctly.

gohryt:
I understand, but there is a problem beacuse we can't debug it in any clear way. If u know any good way for debugging it - tell me and i'll do it.

Techatrix:
You can let ZLS log every request it receives [here](https://github.com/zigtools/zls/blob/4790cb1ab425ba97e82a22e909b4d42b37a5cb95/src/Server.zig#L1816). I'm not sure if Zed has any way of enabling message tracing so you can just manually remove that `if (server.message_tracing)` check. The stack trace in a debug build should also show how the assertion was reached from the request handler. I can provides some more tips if you wish but I hope that helps.

I could also try and take a look at this myself if you can provide steps on how you caused this problem. Any information like the [Bug Report](https://github.com/zigtools/zls/issues/new?assignees=&labels=bug&projects=&template=bug.yml) template would be appreciated.

Issue #2040, opened 2024-09-29
fanotify_init failed due to old kernel; requires 5.17+


mocompute:
### Zig Version

0.14.0-dev.1671+085cc54aa

### ZLS Version

0-14-0.dev

### Client / Code Editor / Extensions

emacs eglot

### Steps to Reproduce and Observed Behavior

Mistakenly thought this was a zig issue, migrating it here. (From https://github.com/ziglang/zig/issues/21543)

I ran into this trying to diagnose a zls problem, where it is not recognizing any of my package dependencies. Watching the zls.log, I see it invokes a build executable, which generates some JSON output with deps_build_roots, packages, etc. Basically the metadata that zls needs.

However, immediately after the JSON, this error is printed, and it exits with status 1. So zls gives up and it won't navigate to types or declarations in my dependencies.

fanotify_init failed due to old kernel; requires 5.17+

My uname -a:

Linux nixos 5.15.153.1-microsoft-standard-WSL2 #1 SMP Fri Mar 29 23:14:13 UTC 2024 x86_64 GNU/Linux

### Expected Behavior

Possibly don't invoke watch mode on old kernels.

### Relevant log output

```shell
error (store ): Failed to execute build runner to collect build configuration, command:
/run/current-system/sw/bin/zig build --build-runner /home/mo/.cache/zls/build_runner/4ebeb852873127b5ece4bcd8e405fc28/build_runner.zig
Error: error: fanotify_init failed due to old kernel; requires 5.17+
```

mocompute:
Seems a simple fix: catch the error from Build.init and return:

https://github.com/zigtools/zls/blob/db05a1ce337fa53927e0635cc4ec5145723584a4/src/build_runner/0.12.0.zig#L421-L422

PR #2041, opened 2024-09-29
Ignore failure in Watch.init and just return

mocompute:
Addresses #2040 but a change in Zig is required before it closes our bug: https://github.com/ziglang/zig/issues/21543#issuecomment-2381281698


mocompute:
I've confirmed that I can now navigate dependency package sources again with 0.14.0-dev through zls with this and the ziglang/zig patch applied.

PR #2043, opened 2024-10-01
fix: enum member declarations can now be renamed correctly.

### LINKED ISSUE #2042: Cannot rename enum fields using VSCode extension.
### Zig Version

0.13.0

### ZLS Version

0.13.0

Visual Studio Code-1.93.1

Init new project with `zig init`.
Edit `main.zig` as follows:
```zig
const std = @import("std");

pub const SomeEnum = enum {
    zero,
    one,
};

pub fn main() !void {
    const se = SomeEnum.one;
    std.log.debug("se: {?}", .{se});
}
```

Right click (or Ctrl+R) the field of enum `SomeEnum`, for example `zero`, to rename. Enter new name and press enter. Popup appears at the right bottom of VSCode saying that `Rename failed to apply edits`.

Note that I can rename struct field.

### Expected Behavior

Enum field itself and all the usage of the field are renamed.

### Relevant log output

No errors were shown in the log:

```shell
info : ( main ): Starting ZLS 0.13.0 @ '/usr/bin/zls'
info : (server): Client is 'Visual Studio Code-1.93.1'
info : (server): No config file zls.json found. This is not an error.
warning: (config): failed to open entry in PATH '/usr/bin/ccache/zig': error.NotDir
info : (server): Set config option 'builtin_path' to '/home/user/.cache/zls/builtin.zig'
info : (server): Set config option 'zig_lib_path' to '/home/user/.zvm/0.13.0/lib'
info : (server): Set config option 'zig_exe_path' to '/home/user/.zvm/bin/zig'
info : (server): Set config option 'build_runner_path' to '/home/user/.cache/zls/build_runner/21872970afd69e48a0847077e5196711/build_runner.zig'
info : (server): Set config option 'global_cache_path' to '/home/user/.cache/zls'
info : (server): Set config option 'enable_build_on_save' to 'true'
info : (server): Set config option 'enable_autofix' to 'true'
info : (server): Set config option 'semantic_tokens' to 'partial'
info : (server): Set config option 'warn_style' to 'true'
info : (server): Set config option 'highlight_global_var_declarations' to 'true'
```

srijan-paul:
Fixes #2042.

Currently, renaming enum field members will return two duplicate edit actions in the LSP response.
On VSCode, this causes the editor to raise an error.
On Emacs, Neovim, and Helix, the edits made are incorrect.

The`symbolReferences` function tries to find all references to a declaration *and* the declaration itself (when `include_decls` flag is set). However, when searching for the references to an enum member like in here:

```zig
const Number = enum {
  zero, // <-- rename triggered from here
  one
};

const myZero = Number.zero; // <-- rename should happen here too
_ = myZero;
```

The first identifier `zero` is treated first as a declaration and then *also as a reference*.
This edge case can be handled by ensuring the name token for a symbol's declaration is never the same as its reference's name token.

Ideally, the `lookup` function should never be called for a container field init identifier, but the way AST traversal works right now, there is no straightforward way to figure out if the node we're currently on declares an enum member.

Before:

https://github.com/user-attachments/assets/289b466b-f8a0-4edd-9516-cf70a6530543

After:

https://github.com/user-attachments/assets/142d1995-9377-4bda-9488-53508a582bcd

### DIFF
--- a/src/features/references.zig
+++ b/src/features/references.zig
@@ -68,6 +68,8 @@ const Builder = struct {
     locations: std.ArrayListUnmanaged(types.Location) = .{},
     /// this is the declaration we are searching for
     decl_handle: Analyser.DeclWithHandle,
+    /// Whether the `decl_handle` has been added
+    did_add_decl_handle: bool = false,
     analyser: *Analyser,
     encoding: offsets.Encoding,
 
@@ -81,6 +83,12 @@ const Builder = struct {
     }
 
     fn add(self: *Builder, handle: *DocumentStore.Handle, token_index: Ast.TokenIndex) error{OutOfMemory}!void {
+        if (self.decl_handle.handle == handle and
+            self.decl_handle.nameToken() == token_index)
+        {
+            if (self.did_add_decl_handle) return;
+            self.did_add_decl_handle = true;
+        }
         try self.locations.append(self.allocator, .{
             .uri = handle.uri,
             .range = offsets.tokenToRange(handle.tree, token_index, self.encoding),
--- a/tests/lsp_features/references.zig
+++ b/tests/lsp_features/references.zig
@@ -97,6 +97,16 @@ test "for/while capture" {
     );
 }
 
+test "enum field access" {
+    try testReferences(
+        \\const E = enum {
+        \\  <0>,
+        \\  bar
+        \\};
+        \\const e = E.<0>;
+    );
+}
+
 test "struct field access" {
     try testReferences(
         \\const S = struct {<0>: u32 = 3};


Techatrix (changes requested):
This check can be simplified by comparing the two name tokens for equality:
```
child.handle == handle and child.nameToken() == name_token
```
We can also move this to the `Builder.add` function to generalize the solution and avoid unnecessarily running this check on decls that we going to be ignored anyway. 

This is going to give us the following patch (includes a test case):
```patch
diff --git a/src/features/references.zig b/src/features/references.zig
index f475086..6ed8095 100644
--- a/src/features/references.zig
+++ b/src/features/references.zig
@@ -68,6 +68,8 @@ const Builder = struct {
     locations: std.ArrayListUnmanaged(types.Location) = .{},
     /// this is the declaration we are searching for
     decl_handle: Analyser.DeclWithHandle,
+    /// Whether the `decl_handle` has been added
+    did_add_decl_handle: bool = false,
     analyser: *Analyser,
     encoding: offsets.Encoding,
 
@@ -81,6 +83,12 @@ const Builder = struct {
     }
 
     fn add(self: *Builder, handle: *DocumentStore.Handle, token_index: Ast.TokenIndex) error{OutOfMemory}!void {
+        if (self.decl_handle.handle == handle and
+            self.decl_handle.nameToken() == token_index)
+        {
+            if (self.did_add_decl_handle) return;
+            self.did_add_decl_handle = true;
+        }
         try self.locations.append(self.allocator, .{
             .uri = handle.uri,
             .range = offsets.tokenToRange(handle.tree, token_index, self.encoding),
diff --git a/tests/lsp_features/references.zig b/tests/lsp_features/references.zig
index 250ccd9..c11c588 100644
--- a/tests/lsp_features/references.zig
+++ b/tests/lsp_features/references.zig
@@ -97,6 +97,16 @@ test "for/while capture" {
     );
 }
 
+test "enum field access" {
+    try testReferences(
+        \\const E = enum {
+        \\  <0>,
+        \\  bar
+        \\};
+        \\const e = E.<0>;
+    );
+}
+
 test "struct field access" {
     try testReferences(
         \\const S = struct {<0>: u32 = 3};

```

srijan-paul:
Ah yes, the tokens are indices anyway.
Updated the PR.

Techatrix (approved):

PR #2046, opened 2024-10-07
fix: goto type definition for annotated declarations

### LINKED ISSUE #2020: `typeDefinition` goes to variables type annotation when provided instead of the types definition
### Zig Version

0.14.0-dev.1407+0b9fccf50

### ZLS Version

0.14.0-dev.145+ace6f6d

_No response_

```zig
const S = struct {};
const A: S = S{};
const B = S{};
```
goto type definition on `A` takes you to its own type annotation while doing so on `B` takes you to the actual definition of `S`

### Expected Behavior

both should go to the definition of `S`

### Relevant log output

_No response_

srijan-paul:
Fixes #2020.
Includes a test as well.

Previously, to serve `textDocument/typeDefinition`, we just called `resolveType`, which would usually return annotations.
But we never followed those annotations to find out where they were declared.

So we'd go from the `variable` node to the `Ty` annotation node, but never to the struct that `Ty` is initialized with:
```zig
const Ty = struct{};
const variable: Ty = Ty{};
```

Now, we attempt to resolve `Ty` further (which lands us at `struct`).
If that fails, we return the annotation's location as we previously did.


Techatrix (changes requested):
I tried out moving `resolveType` to the top and only if that fails, use `typeDeclarationNode`. It seems to get the job done while being much simpler.

Excluding the added test and refactoring, the core changes end up being this:
```patch
diff --git a/src/features/goto.zig b/src/features/goto.zig
index 1e28721..c32cce2 100644
--- a/src/features/goto.zig
+++ b/src/features/goto.zig
@@ -37,16 +37,11 @@ fn gotoDefinitionSymbol(
         .declaration => try decl_handle.definitionToken(analyser, false),
         .definition => try decl_handle.definitionToken(analyser, true),
         .type_definition => blk: {
-            // we should actually resolve the definition but declaration is can be more reliably implemented.
-            const type_declaration = try decl_handle.typeDeclarationNode() orelse {
-                // just resolve the type and guess
-                if (try decl_handle.resolveType(analyser)) |resolved_type| {
-                    if (try resolved_type.typeDefinitionToken()) |token_handle| {
-                        break :blk token_handle;
-                    }
-                }
-                return null;
-            };
+            if (try decl_handle.resolveType(analyser)) |ty| {
+                if (try ty.typeDefinitionToken()) |token_handle| break :blk token_handle;
+            }
+            const type_declaration = try decl_handle.typeDeclarationNode() orelse return null;
+
             const target_range = offsets.nodeToRange(type_declaration.handle.tree, type_declaration.node, offset_encoding);
 
             return types.DefinitionLink{
```
Be aware that I have not tested this beyond just running `zig build test`.

Techatrix:
I noticed that your commits are signed but cannot be verified by GitHub. You may wanna investigate why this is happening. Perhaps you forgot to [add the GPG key](https://github.com/settings/keys) to GitHub.

srijan-paul:
Thanks, updated PR.
Added another test for a global without annotation.

> I noticed that your commits are signed but cannot be verified by GitHub.

Yes, I didn't add my GPG keys to GitHub.
Would you prefer if my commits were verified?

Techatrix:
> Would you prefer if my commits were verified?

I don't have any strong preference on this. Signed commits are nice to have but not required in this project. But if decide to sign your commits, make sure they are properly signed.

Techatrix (approved):

PR #2049, opened 2024-10-09
Remove unused imports

srijan-paul:


Techatrix (approved):
That code snippet in `semantic_tokens.zig` looks like something I accidentally added while testing #1764.

Issue #1093, opened 2023-03-25
add better support for `"source.fixAll"`


Techatrix:
If you don't add "source.fixAll" to "editor.codeActionsOnSave" autofix won't work in VS-Code (workaround in #1094). It would be helpful if vscode-zig did this for you, but I'm not sure how to achieve this without also enabling "source.fixAll" on other languages. I think you could consider this an oversight by VS-Code that there is not way to enable this on a per language level which is probably why as an example eslint is using "source.fixAll.eslint". So we either fix VS-Code or add "source.fixAll.zls"?

PR #2051, opened 2024-10-09
Code action to organize imports

### LINKED ISSUE #1637: code action for organizing imports
I've attempted to add a new code actions that would automatically sort your `@import`s in #881 but I never got around to finishing it and it started rot away.

If there is anyone interested to continue where I left of, my old code should provide a good starting point.

Sekky61:
# Organize Imports Diagnostic and Code Action

This PR adds a diagnostic and a code action to organize imports in the current Zig file. The code action is available in the command palette as `organize @import`.

It is designed to be fairly customizable, so that we can arrive at a acceptable sorting style, like different grouping or case sensitivity.
All decisions, especially those around diagnostics are very much open to critique as I do not have experience with the LSP.

Closes #1637, continues work from #881.

## Design

An import is a top level declaration of a variable that uses the `@import` builtin.
A top level declaration that uses import variables (like `const print = std.debug.print;`) is not considered an import and stays in place.

The sorting splits imports into three separated groups: builtin packages, packages and project files.
The builtin packages are sorted in the order: `std`, `builtin`, `build_options`.
The sorting inside other groups is done by comparing the declaration names (the `x` in `const x = @import("file.zig");`).

The feature is tested both in unit tests and in neovim. The unit tests required a new helper function.

### Diagnostic

The diagnostic is enabled under the `warn_style` zls config option, which is disabled by default.
The diagnostic's range is the entire line of the import which would be moved by organizing imports.
That means one diagnostic per wrongly placed import line.

The diagnostic does not consider the correctness of separators between imports.

### Code Action

The code action is available based on the diagnostic.
This means that users with the `warn_style` option disabled will not be offered the code action.

The code action always sorts. The previous PR had optimization using `isSorted` to avoid sorting when it was not necessary.
This, however, did not work well in some corner cases.

The documentation comments (`///`) are moved with the import they are attached to.
It also respects `Top-Level Doc Comments` (`//!`), which must appear before any expression. This is handled using the commments token location.

This feature required running `generateCodeAction` for each diagnostic provided in request.
Previously, it ran the function only on diagnostics from `getAstCheckDiagnostics`.

## Considerations and notes

- Top-level declarations (that includes imports) are order-independent ([Source](https://ziglang.org/documentation/master/#Values) and tested)
- `zig fmt` and `textDocument/formatting` does not interfere with the sorting.
- Only root declarations are considered for sorting (no import inside functions).
- Will not work with `@import(variable);` (we don't do comptime)
- Resolving one diagnostic will resolve all diagnostics in the file (since all imports are moved).
- I removed the part dealing with intersection with action's range (`builder.range` field which no longer exists). If it was relevant, I can study it further.
- The code action works without `warn_style` as seen in tests. It is just that user cannot call it without the diagnostic.
- `std.sort.sort` does not exist anymore. I picked `std.sort.heap` as a replacement. My understanding is that we do not need a stable sort here, but feel free to correct me and suggest a different sort from the `std` library.

## Unresolved questions

- [x] Allow for `const print = std.debug.print;` to be sorted as well in some way?
- [x] How to treat `@import("root")`? (Currently a package, which does not seem right)
- [x] The `getSortSlice` function ported from previous PR seems weird. I would just do `return self.name` and not look in the import path. Any reasoning behind this?
- [ ] What about grouping by directory for the file imports? Perhaps if there are > 1 file imports from the same directory, they could be in a group of their own.
- [ ] I think `DiagnosticKind` could be parsed from the diagnostic code, which would be cleaner and would easily allow more versions of the diagnostic message. Or use the `data` field of the diagnostic message.
- [ ] What about `@embedFile`?
- [x] How to implement ignoring the diagnostic? `Ast.zig`, for example, imports extra stuff near end of the file for tests.

### DIFF
--- a/src/features/code_actions.zig
+++ b/src/features/code_actions.zig
@@ -373,10 +373,11 @@ fn handleUnorganizedImport(builder: *Builder, actions: *std.ArrayListUnmanaged(t
 
     const imports = try getImportsDecls(tree, builder.arena);
 
+    // The optimization is disabled because it does not detect the case where imports and other decls are mixed
     // if (std.sort.isSorted(ImportDecl, imports.items, tree, ImportDecl.lessThan)) return;
 
     const sorted_imports = try builder.arena.dupe(ImportDecl, imports.items);
-    std.sort.heap(ImportDecl, sorted_imports, tree, ImportDecl.lessThan);
+    std.mem.sort(ImportDecl, sorted_imports, tree, ImportDecl.lessThan);
 
     var edits = std.ArrayListUnmanaged(types.TextEdit){};
 
--- a/src/features/code_actions.zig
+++ b/src/features/code_actions.zig
@@ -586,25 +586,22 @@ pub fn getImportsDecls(builder: *Builder, allocator: std.mem.Allocator) error{Ou
                 if (import_decl.var_decl == node) continue :next_decl;
             }
             const var_decl = tree.simpleVarDecl(node);
-            const base_token = var_decl.ast.init_node;
-
-            var inode = base_token;
+            var current_node = var_decl.ast.init_node;
             const import: ImportDecl = found_decl: while (true) {
-                const token = node_tokens[inode];
-                switch (node_tags[inode]) {
+                const token = node_tokens[current_node];
+                switch (node_tags[current_node]) {
                     .builtin_call_two, .builtin_call_two_comma => {
-                        // @import("string") case
-                        const init_node = inode;
-                        const call_name = offsets.tokenToSlice(tree, token);
-                        if (!std.mem.eql(u8, call_name, "@import")) continue :next_decl;
+                        // `>@import("string")<` case
+                        const builtin_name = offsets.tokenToSlice(tree, token);
+                        if (!std.mem.eql(u8, builtin_name, "@import")) continue :next_decl;
                         // TODO what about @embedFile ?
 
-                        if (node_data[init_node].lhs == 0 or node_data[init_node].rhs != 0) continue :next_decl;
-                        const import_param_node = node_data[init_node].lhs;
-                        if (node_tags[import_param_node] != .string_literal) continue :next_decl;
+                        if (node_data[current_node].lhs == 0 or node_data[current_node].rhs != 0) continue :next_decl;
+                        const param_node = node_data[current_node].lhs;
+                        if (node_tags[param_node] != .string_literal) continue :next_decl;
 
                         const name_token = var_decl.ast.mut_token + 1;
-                        const value_token = node_tokens[import_param_node];
+                        const value_token = node_tokens[param_node];
 
                         break :found_decl .{
                             .var_decl = node,
@@ -614,20 +611,20 @@ pub fn getImportsDecls(builder: *Builder, allocator: std.mem.Allocator) error{Ou
                         };
                     },
                     .field_access => {
-                        // `@import("foo").bar` or `foo.bar` case
+                        // `@import("foo").>bar<` or `foo.>bar<` case
                         // drill down to the base import
-                        inode = node_data[inode].lhs;
+                        current_node = node_data[current_node].lhs;
                         continue;
                     },
                     .identifier => {
-                        // `std.ascii` case - Might be an import
+                        // `>std<.ascii` case - Might be an alias
                         const slice = offsets.tokenToSlice(tree, token);
                         const idx = offsets.tokenToIndex(tree, token);
                         const symbolDecl = try builder.analyser.lookupSymbolGlobal(builder.handle, slice, idx) orelse continue :next_decl;
-                        const declIdx = symbolDecl.decl.ast_node;
+                        const decl_found = symbolDecl.decl.ast_node;
                         // if the decl is in known imports, add this one as well
                         for (imports.items) |import_decl| {
-                            if (import_decl.var_decl == declIdx) {
+                            if (import_decl.var_decl == decl_found) {
                                 break :found_decl .{
                                     .var_decl = node,
                                     .first_comment_token = Analyser.getDocCommentTokenIndex(tree.tokens.items(.tag), node_tokens[node]),
--- a/src/features/code_actions.zig
+++ b/src/features/code_actions.zig
@@ -413,14 +413,9 @@ fn handleUnorganizedImport(builder: *Builder, actions: *std.ArrayListUnmanaged(t
         }
         try writer.writeByte('\n');
 
-        const insert_pos: types.Position = pos: {
-            const tokens = tree.tokens.items(.tag);
-            if (tokens.len < 2 or tokens[0] != .container_doc_comment) {
-                break :pos .{ .line = 0, .character = 0 };
-            }
-            // Insert after doc comment
-            break :pos offsets.tokenToPosition(tree, 1, builder.offset_encoding);
-        };
+        const tokens = tree.tokens.items(.tag);
+        const first_token = std.mem.indexOfNone(std.zig.Token.Tag, tokens, &.{.container_doc_comment}) orelse tokens.len;
+        const insert_pos = offsets.tokenToPosition(tree, @intCast(first_token), builder.offset_encoding);
 
         try edits.append(builder.arena, .{
             .range = .{ .start = insert_pos, .end = insert_pos },
--- a/src/features/code_actions.zig
+++ b/src/features/code_actions.zig
@@ -47,7 +47,6 @@ pub const Builder = struct {
                 // fix: remove code
             },
             .var_never_mutated => try handleVariableNeverMutated(builder, actions, loc),
-            .unorganized_import => try handleUnorganizedImport(builder, actions),
         }
     }
 
@@ -552,15 +551,9 @@ pub const ImportDecl = struct {
             last_token += 1;
         }
 
-        var end = offsets.tokenToLoc(tree, last_token).end;
+        const end = offsets.tokenToLoc(tree, last_token).end;
         if (!include_line_break) return end;
-        while (end < tree.source.len) {
-            switch (tree.source[end]) {
-                ' ', '\t', '\n' => end += 1,
-                else => break,
-            }
-        }
-        return end;
+        return std.mem.indexOfNonePos(u8, tree.source, end, &.{ ' ', '\t', '\n' }) orelse tree.source.len;
     }
 
     /// similar to `offsets.nodeToLoc` but will also include preceding comments and postfix semicolon and line break
@@ -603,7 +596,7 @@ pub fn getImportsDecls(builder: *Builder, allocator: std.mem.Allocator) error{Ou
                         // @import("string") case
                         const init_node = inode;
                         const call_name = offsets.tokenToSlice(tree, token);
-                        if (!std.mem.eql(u8, call_name, "@import")) continue;
+                        if (!std.mem.eql(u8, call_name, "@import")) continue :next_decl;
                         // TODO what about @embedFile ?
 
                         if (node_data[init_node].lhs == 0 or node_data[init_node].rhs != 0) continue :next_decl;
@@ -621,7 +614,7 @@ pub fn getImportsDecls(builder: *Builder, allocator: std.mem.Allocator) error{Ou
                         };
                     },
                     .field_access => {
-                        // `@import("foo").foo.bar` or `@import("foo").foo.bar` case
+                        // `@import("foo").bar` or `foo.bar` case
                         // drill down to the base import
                         inode = node_data[inode].lhs;
                         continue;
@@ -816,7 +809,6 @@ const DiagnosticKind = union(enum) {
     undeclared_identifier,
     unreachable_code,
     var_never_mutated,
-    unorganized_import,
 
     const IdCat = enum {
         @"function parameter",
@@ -852,8 +844,6 @@ const DiagnosticKind = union(enum) {
             return .undeclared_identifier;
         } else if (std.mem.eql(u8, msg, "local variable is never mutated")) {
             return .var_never_mutated;
-        } else if (std.mem.eql(u8, msg, "unorganized @import statement")) {
-            return .unorganized_import;
         }
         return null;
     }
--- a/src/features/code_actions.zig
+++ b/src/features/code_actions.zig
@@ -51,6 +51,13 @@ pub const Builder = struct {
         }
     }
 
+    pub fn generateOrganizeImportsAction(
+        builder: *Builder,
+        actions: *std.ArrayListUnmanaged(types.CodeAction),
+    ) error{OutOfMemory}!void {
+        try handleUnorganizedImport(builder, actions);
+    }
+
     pub fn createTextEditLoc(self: *Builder, loc: offsets.Loc, new_text: []const u8) types.TextEdit {
         const range = offsets.locToRange(self.handle.tree.source, loc, self.offset_encoding);
         return types.TextEdit{ .range = range, .newText = new_text };
--- a/src/features/code_actions.zig
+++ b/src/features/code_actions.zig
@@ -47,6 +47,7 @@ pub const Builder = struct {
                 // fix: remove code
             },
             .var_never_mutated => try handleVariableNeverMutated(builder, actions, loc),
+            .unorganized_import => try handleUnorganizedImport(builder, actions),
         }
     }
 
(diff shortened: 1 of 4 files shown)


leecannon:
I'm not sure about this being behind `warn_style`, the [style guide](https://ziglang.org/documentation/master/#Style-Guide) makes no mention of the order or location of imports.

A problem for having a "blessed" location in the file for imports, is a few projects (including zig itself) are actually moving to have all imports at the bottom of files as they feel imports are just boilerplate.

Also I can't get the code action to apply in vscode, neither as a quick fix nor in the command palette.

image
image

Sekky61:
Interesting stuff! I agree, I just wanted the diagnostic to not be on by default (people have preferences and it could be annoying). So two solutions that come to mind are 1) creating an entirely new config flag or 2) scrapping the diagnostic and having the code action just available whenever.

I will take a look at the problem you mention. It is supposed to appear when having the cursor on the diagnostic.

Techatrix:
Thank you for taking the time to revive #881.The detailed an thorough description of this PR is very helpful. I will try to answer/address some of the points that you have mentioned.

> * How to treat `@import("root")`? (Currently a package, which does not seem right)

There are three "builtin packages":
- `std`
- `builtin`
- `root`

`build_options` is not one of them since it added through the `build.zig`. I could understand adding it to the other "builtin packages" since its a very common convention.

> * `std.sort.sort` does not exist anymore.

This function has been moved to `std.mem.sort`.

> * Will not work with `@import(variable);` (we don't do comptime)

Zig requires that the first parameter of `@import` is a string literal so this is not an issue.

> * The `getSortSlice` function ported from previous PR seems weird. I would just do `return self.name` and not look in the import path. Any reasoning behind this?

IIRC, this function was added to make it easier to control which string/name is used to do sorting. #881 sorted imported files like `const foo = @import("bar/baz.zig")` by the filename (`baz.zig`) instead of `foo`. It could become useful if we decode to change the logic later on.

> * I removed the part dealing with intersection with action's range (`builder.range` field which no longer exists). If it was relevant, I can study it further.

This was added as an optimization. The given range tells us for which part of the document the code actions should be computed. If the imports are not visible then there is no need to compute the code action. It not necessary to bring this back to get this PR merged.

> The documentation comments (`///`) are moved with the import they are attached to.

What about non doc-comments (`//`)? One TODO task I had in #881 was "preserve comments". This was especially tricky for non doc-comments. The currently implementation moves them all after the imports. I could foresee a situation where the users want to keep them between the imports for some reason. I am not sure what the best approach would be. We can leave this as a followup task.

> Also I can't get the code action to apply in vscode, neither as a quick fix nor in the command palette.

I have the same problem. Using the "Source Action..." command in VS Code is the only way I was able to run the code actions.

---

I would also prefer to not have this feature rely on `warn_style`. I plan on removing this option at some point to replace it with a linter. It is possible to have code actions without reporting a diagnostic. With #1093, editors that support it can run `source.organizeImports` code actions automatically on save.

src/features/diagnostics.zig
```diff
@@ -118,6 +118,47 @@ pub fn generateDiagnostics(server: *Server, arena: std.mem.Allocator, handle: *D
                     else => {},
                 }
             }
+
+            // Organize imports
+            const imports = try code_actions.getImportsDecls(tree, arena);
+            const rootDecls = tree.rootDecls();
+            std.log.debug("Imports: {any}", .{imports.items});
+            std.log.debug("RootDecls: {any}", .{rootDecls});
```
Techatrix, line 126:
```suggestion
```

src/features/diagnostics.zig
```diff
@@ -118,6 +118,47 @@ pub fn generateDiagnostics(server: *Server, arena: std.mem.Allocator, handle: *D
...
+                    });
+                }
+            }
+            // Now check the order of the imports at the correct position by sorting the indexes
+            const sort_checked = imports.items[0..first_bad_import];
+            var decl_indexes = try arena.alloc(usize, sort_checked.len);
+            for (decl_indexes[0..decl_indexes.len], 0..) |*decl_index, i| {
+                decl_index.* = i;
+            }
+            std.sort.block(usize, decl_indexes, .{ sort_checked, tree }, sortIdxLessThen);
```
Techatrix, line 151:
```suggestion
            std.mem.sort(usize, decl_indexes, .{ sort_checked, tree }, sortIdxLessThen);
```

tests/lsp_features/code_actions.zig
```diff
@@ -378,7 +538,8 @@ fn testAutofixOptions(before: []const u8, after: []const u8, want_zir: bool) !vo
     const handle = ctx.server.document_store.getHandle(uri).?;
 
     var diagnostics: std.ArrayListUnmanaged(types.Diagnostic) = .{};
-    try zls.diagnostics.getAstCheckDiagnostics(ctx.server, ctx.arena.allocator(), handle, &diagnostics);
+    // try zls.diagnostics.getAstCheckDiagnostics(ctx.server, ctx.arena.allocator(), handle, &diagnostics);
+    defer diagnostics.deinit(allocator);
```
Techatrix, line 542: No deinit is necessary because the diagnostics are allocated with an arena allocator.
```suggestion
    try zls.diagnostics.getAstCheckDiagnostics(ctx.server, ctx.arena.allocator(), handle, &diagnostics);
```
Why the autofix tests are failed should be investigated.

tests/lsp_features/code_actions.zig
```diff
@@ -363,6 +363,166 @@ test "ignore autofix comment whitespace" {
...
+        .start = .{ .line = 1, .character = 0 },
+        .end = .{ .line = 2, .character = 0 },
+    },
+    .severity = .Hint,
+    .code = .{ .string = "unorganized_import" },
+    .source = "zls",
+    .message = "unorganized @import statement",
+};
+
+test "organize imports" {
```
Techatrix, line 377: The calls to `testDiagnostic` can be moved into separate test decls. Here are some suggestions:
```zig
test "organize imports" { ... }
test "organize imports - preserve doc comments" { ... }
test "organize imports - top-level doc-comment" { ... }
test "organize imports - duplicate builtin import" { ... }
```

tests/lsp_features/code_actions.zig
```diff
@@ -415,3 +578,52 @@ fn testAutofixOptions(before: []const u8, after: []const u8, want_zir: bool) !vo
...
+    defer diagnostics.deinit(allocator);
+
+    const params = types.CodeActionParams{
+        .textDocument = .{ .uri = uri },
+        .range = .{
+            .start = .{ .line = 0, .character = 0 },
+            .end = offsets.indexToPosition(before, before.len, ctx.server.offset_encoding),
+        },
+        .context = .{ .diagnostics = diagnostics.items },
+    };
```
Techatrix, line 601:
```suggestion
    const params = types.CodeActionParams{
        .textDocument = .{ .uri = uri },
        .range = .{
            .start = .{ .line = 0, .character = 0 },
            .end = offsets.indexToPosition(before, before.len, ctx.server.offset_encoding),
        },
        .context = .{ .diagnostics = &.{diagnostic} },
    };
```

src/features/code_actions.zig
```diff
@@ -366,6 +367,231 @@ fn handleVariableNeverMutated(builder: *Builder, actions: *std.ArrayListUnmanage
...
+        }
+        try writer.writeByte('\n');
+
+        const insert_pos: types.Position = pos: {
+            const tokens = tree.tokens.items(.tag);
+            if (tokens.len < 2 or tokens[0] != .container_doc_comment) {
+                break :pos .{ .line = 0, .character = 0 };
+            }
+            // Insert after doc comment
+            break :pos offsets.tokenToPosition(tree, 1, builder.offset_encoding);
```
Techatrix, line 403: In my limited testing I did on this PR, I have encountered an issue with files that had top-level doc-comments.
Doc comments are emitted per-line. If a top-level doc-comments spans across multiple lines then `getSourceStartIndex` will return the position after the first line.

This should resolved the issue:
```suggestion

        const tokens = tree.tokens.items(.tag);
        const first_token = std.mem.indexOfNone(std.zig.Token.Tag, tokens, &.{.container_doc_comment}) orelse tokens.len;
        const insert_pos = offsets.tokenToPosition(tree, @intCast(first_token), builder.offset_encoding);
```

This should also have test coverage.

src/features/code_actions.zig
```diff
@@ -366,6 +367,231 @@ fn handleVariableNeverMutated(builder: *Builder, actions: *std.ArrayListUnmanage
...
+fn handleUnorganizedImport(builder: *Builder, actions: *std.ArrayListUnmanaged(types.CodeAction)) !void {
+    const tree = builder.handle.tree;
+    if (tree.errors.len != 0) return;
+
+    const imports = try getImportsDecls(tree, builder.arena);
+
+    // if (std.sort.isSorted(ImportDecl, imports.items, tree, ImportDecl.lessThan)) return;
+
+    const sorted_imports = try builder.arena.dupe(ImportDecl, imports.items);
+    std.sort.heap(ImportDecl, sorted_imports, tree, ImportDecl.lessThan);
```
Techatrix, line 379:
```suggestion
    std.mem.sort(ImportDecl, sorted_imports, tree, ImportDecl.lessThan);
```

Techatrix (changes requested):
Here are some thing I noticed from looking over the code. I will give this a more thorough review later on.

src/features/code_actions.zig
```diff
@@ -366,6 +367,231 @@ fn handleVariableNeverMutated(builder: *Builder, actions: *std.ArrayListUnmanage
...
+    pub fn getSortSlice(self: ImportDecl) []const u8 {
+        switch (self.getKind()) {
+            .file => {
+                if (std.mem.indexOfScalar(u8, self.value, '/') != null) {
+                    return self.value[1 .. self.value.len - 1];
+                }
+                return self.name;
+            },
+            // There used to be unreachable for other than file and package, but the user
+            // can just write @import("std") twice.
```
Techatrix, line 495: Great discovery! I never though about duplicate imports.

Techatrix:
> * Allow for `const print = std.debug.print;` to be sorted as well in some way?

There are two possibles I could imagine:

### Move alias to the end of each section

```zig
const std = @import("std");
const Ast = std.zig.Ast;
const Server = std.zig.Server;

const lsp = @import("lsp");
const AnyTransport = lsp.AnyTransport;
const types = lsp.types;

const DocumentScope = @import("../DocumentScope.zig");
const Declaration = DocumentScope.Declaration;
const Scope = DocumentScope.Scope;
```

### Move alias after all imports

```zig
const std = @import("std");

const lsp = @import("lsp");

const DocumentScope = @import("../DocumentScope.zig");

// Not sure if we should sort by variable name or "path". I think path is more reasonable
const Ast = std.zig.Ast;
const Server = std.zig.Server;
const AnyTransport = lsp.AnyTransport;
const types = lsp.types;
const Declaration = DocumentScope.Declaration;
const Scope = DocumentScope.Scope;
```

I would personally prefer the first option but I am open to suggestions.

> * What about `@embedFile`?

Maybe move them into their own section after all `@import("file.zig")` imports.

Sekky61:
I am happy to say that I made progress on the VS Code situation.
First of all, the import organizing does work with `editor.codeActionsOnSave` setting:


https://github.com/user-attachments/assets/a11379e6-0e01-42d6-84c4-90cd90cd40c7

But as you pointed out, the Organize Imports does not show up in the Command pallet and the shortcut does not work either. So I investigated with other LSPs.
Rust-analyzer did _not_ work either. But Typescript LSP did.

After looking around the code, I found the difference in **registering `codeActionProvider`**.
This is the change:

```diff
--- a/src/Server.zig
+++ b/src/Server.zig
@@ -615,7 +615,9 @@ fn initializeHandler(server: *Server, arena: std.mem.Allocator, request: types.I
             },
             .documentHighlightProvider = .{ .bool = true },
             .hoverProvider = .{ .bool = true },
-            .codeActionProvider = .{ .bool = true },
+            .codeActionProvider = .{ .CodeActionOptions = .{ .codeActionKinds = &.{
+                .@"source.organizeImports",
+            } } },
             .declarationProvider = .{ .bool = true },
             .definitionProvider = .{ .bool = true },
             .typeDefinitionProvider = .{ .bool = true },
```
And it's there.

20241017_18h01m53s_grim

I suppose it has to be explicitly listed as supported. So how do we implement this? Simply create an array of all the code actions we support?

Techatrix:
> I suppose it has to be explicitly listed as supported. So how do we implement this? Simply create an array of all the code actions we support?

I don't think there is any alternative to just hard-coding the supported code action kinds. I also noticed that Sublime Text LSP requires `source.fixAll` to be explicitly listed so that their `lsp_code_actions_on_save` setting works.

Sekky61:
Moving aliases to the right group has been implemented. I went for grouping based on the parent import:

```zig
const abc = @import("abc.zig");
const abc_related = abc.related;
const abc_related2 = abc.related2;
const xyz = @import("xyz.zig");
const xyz_related = xyz.related;
```

I am interested in what you think could be improved

src/features/code_actions.zig
```diff
@@ -366,6 +385,281 @@ fn handleVariableNeverMutated(builder: *Builder, actions: *std.ArrayListUnmanage
...
+
+            var inode = base_token;
+            const import: ImportDecl = found_decl: while (true) {
+                const token = node_tokens[inode];
+                switch (node_tags[inode]) {
+                    .builtin_call_two, .builtin_call_two_comma => {
+                        // @import("string") case
+                        const init_node = inode;
+                        const call_name = offsets.tokenToSlice(tree, token);
+                        if (!std.mem.eql(u8, call_name, "@import")) continue;
```
Techatrix, line 608: This `continue` will always be an infinite loop. Nothing is being modified on each iteration so you will hit this continue again and again.
Example Test:
```zig
test "organize imports - @embedFile" {
    try testDiagnostic(
        \\const foo = @embedFile("foo.zig");
    , "infinite loop go brrrrrrrr");
}
```
```suggestion
                        if (!std.mem.eql(u8, call_name, "@import")) break;
```

src/features/code_actions.zig
```diff
@@ -524,6 +818,7 @@ const DiagnosticKind = union(enum) {
     undeclared_identifier,
     unreachable_code,
     var_never_mutated,
+    unorganized_import,
```
Techatrix, line 821:
```suggestion
```

src/features/code_actions.zig
```diff
@@ -559,6 +854,8 @@ const DiagnosticKind = union(enum) {
             return .undeclared_identifier;
         } else if (std.mem.eql(u8, msg, "local variable is never mutated")) {
             return .var_never_mutated;
+        } else if (std.mem.eql(u8, msg, "unorganized @import statement")) {
+            return .unorganized_import;
```
Techatrix, line 858:
```suggestion
```

Techatrix (changes requested):
I plan to give some of the code in `getImportsDecls` a closer look later on but this is what I found so far. The sorting behavior seems to work well and I did not encounter issues besides the infinite loop thing.

Techatrix (changes requested):
Sorry for repeatedly requesting changes but this time it is going to be the last. I promise :)

> What about grouping by directory for the file imports? Perhaps if there are > 1 file imports from the same directory, they could be in a group of their own.

No sure, I wouldn't be against this. I do have an alternative idea on how to do the grouping that is a bit more flexible so that we wouldn't need to decide these things. I will try to open a follow-up issue about it (unless I forget about it). 

> What about `@embedFile`?

I would suggest to either move them into their own section all imports and aliases or just leave them be. Ignoring them will kind of achieve the same thing. This can always be changed afterwards.

Is there anything else you would still like to discuss before this gets merged?

src/features/code_actions.zig
```diff
@@ -366,6 +382,272 @@ fn handleVariableNeverMutated(builder: *Builder, actions: *std.ArrayListUnmanage
...
+
+    const node_tags = tree.nodes.items(.tag);
+    const node_data = tree.nodes.items(.data);
+    const node_tokens = tree.nodes.items(.main_token);
+
+    // iterate until no more imports are found
+    var updated = true;
+    while (updated) {
+        updated = false;
+        next_decl: for (tree.rootDecls()) |node| {
```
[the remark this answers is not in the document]
Sekky61: I tried to come up with something smarter, but couldn't. The problem is that they can all depend on each other like a chain:

```zig
const Rational = big.rational; // Resolved in iteration #4
const big = math.big;          // Resolved in iteration #3
const math = std.math;         // Resolved in iteration #2
const std = @import("std");    // Resolved in iteration #1
```

But Zig compiler has to deal with this somehow, so it might be worth it to take inspiration there.

I could attempt to improve it in a followup PR.

Sekky61:
Thanks for collaborating and taking the time to teach me more about Zig!

The followup tasks would need some benchmarking (never optimize blindly). Do you have a recommended microbenchmark setup for Zig?

Techatrix:
> The followup tasks would need some benchmarking (never optimize blindly). Do you have a recommended microbenchmark setup for Zig?

This [post on ziggit.dev](https://ziggit.dev/t/benchmarking/4675) may be useful to you. My usually workflow involves perf, [hotspot](https://github.com/KDAB/hotspot) and [poop](https://github.com/andrewrk/poop).
I created a separate branch at [techatrix/code-action-organize-imports-bench](https://github.com/zigtools/zls/tree/techatrix/code-action-organize-imports-bench) which features a benchmark and some optimizations. I chose to create a standalone executable instead of opting for some microbenchmarking library. It creates a synthetic zig file with various configuration option. I chose to test on files with thousands of imports to stress test the implementation and to make any startup time of the benchmark negligible in comparison.

If you wish to use this as an opportunity to learn more about optimizing code then you can just use my benchmarking program and try to optimize it yourself. I would still suggest to use f27d726035d1084b9cc8ec32f659ed534dd45ed1 because it just performs a more efficient version of calling `offsets.locToRange` in a loop on the same document. The changes to `getImportsDecls` are featured in the next commit.

Here is a showcase of the optimized version performed [poop](https://github.com/andrewrk/poop):
<details>

<summary>Show Results</summary>

```
poop ./zig-out/bin/bench2-safe-old ./zig-out/bin/bench2-safe-new
Benchmark 1 (3 runs): ./zig-out/bin/bench2-safe-old
  measurement          mean ± σ            min … max           outliers         delta
  wall_time          17.1s  ±  130ms    17.0s  … 17.3s           0 ( 0%)        0%
  peak_rss           50.1MB ± 56.8KB    50.1MB … 50.2MB          0 ( 0%)        0%
  cpu_cycles         65.4G  ±  241M     65.3G  … 65.7G           0 ( 0%)        0%
  instructions        109G  ± 42.5K      109G  …  109G           0 ( 0%)        0%
  cache_references   70.9M  ± 2.14M     69.1M  … 73.3M           0 ( 0%)        0%
  cache_misses        389K  ±  116K      257K  …  477K           0 ( 0%)        0%
  branch_misses       129M  ± 38.3K      129M  …  129M           0 ( 0%)        0%
Benchmark 2 (4 runs): ./zig-out/bin/bench2-safe-new
  measurement          mean ± σ            min … max           outliers         delta
  wall_time          1.31s  ± 2.75ms    1.31s  … 1.32s           0 ( 0%)        ⚡- 92.3% ±  0.9%
  peak_rss            108MB ± 74.6KB     108MB …  108MB          0 ( 0%)        💩+115.2% ±  0.3%
  cpu_cycles         4.83G  ± 8.43M     4.82G  … 4.84G           0 ( 0%)        ⚡- 92.6% ±  0.5%
  instructions       16.4G  ±  105K     16.4G  … 16.4G           0 ( 0%)        ⚡- 84.9% ±  0.0%
  cache_references   16.4M  ± 3.69M     13.9M  … 21.8M           0 ( 0%)        ⚡- 76.9% ±  8.7%
  cache_misses        143K  ± 47.9K      110K  …  213K           0 ( 0%)        ⚡- 63.2% ± 41.5%
  branch_misses      4.45M  ± 14.7K     4.43M  … 4.47M           0 ( 0%)        ⚡- 96.6% ±  0.0%
```
The speedup is less extreme on smaller (more realistic) files.

</details>

Sekky61:
Great success!

<details>
  <summary>Bench - reproduced results</summary>
  
  ```bash
❯  poop zig-out/bin/bench-orig zig-out/bin/bench-techatrix
Benchmark 1 (3 runs): zig-out/bin/bench-orig
  measurement          mean ± σ            min … max           outliers         delta
  wall_time          95.0s  ± 2.43s     93.4s  … 97.8s           0 ( 0%)        0%
  peak_rss           53.1MB ±  135KB    53.0MB … 53.2MB          0 ( 0%)        0%
  cpu_cycles          425G  ± 1.60G      424G  …  427G           0 ( 0%)        0%
  instructions       1.01T  ±  246K     1.01T  … 1.01T           0 ( 0%)        0%
  cache_references   1.58G  ± 35.9M     1.55G  … 1.62G           0 ( 0%)        0%
  cache_misses       43.3M  ± 5.08M     38.1M  … 48.2M           0 ( 0%)        0%
  branch_misses       141M  ± 42.0K      141M  …  141M           0 ( 0%)        0%
Benchmark 2 (3 runs): zig-out/bin/bench-techatrix
  measurement          mean ± σ            min … max           outliers         delta
  wall_time          43.2s  ±  189ms    43.0s  … 43.4s           0 ( 0%)        ⚡- 54.5% ±  4.1%
  peak_rss            111MB ±  154KB     111MB …  111MB          0 ( 0%)        💩+108.8% ±  0.6%
  cpu_cycles          192G  ± 1.01G      191G  …  193G           0 ( 0%)        ⚡- 54.7% ±  0.7%
  instructions        478G  ±  228K      478G  …  478G           0 ( 0%)        ⚡- 52.7% ±  0.0%
  cache_references    540M  ± 36.6M      514M  …  582M           0 ( 0%)        ⚡- 65.8% ±  5.2%
  cache_misses       15.7M  ± 1.01M     14.7M  … 16.7M           0 ( 0%)        ⚡- 63.8% ± 19.1%
  branch_misses      11.5M  ± 68.8K     11.5M  … 11.6M           0 ( 0%)        ⚡- 91.8% ±  0.1%
```
</details>

I noticed from the profiler that the `lookupSymbolGlobal` was taking >90% of total time.
I tried to cache known decls by string name, it got faster by about 30%, and +10% memory consumption.

Then I circumvented the `lookupSymbolGlobal` entirely by getting the root scope manually.
This should be okay, since we only care about decls in root scope.
This yielded _another 94% improvement_ for a **total improvement of 97%**! 

<details>
  <summary>Bench - direct root scope lookup (improvement over the first optimization)</summary>
  
  ```bash
Benchmark 1 (3 runs): bench-techatrix
  measurement          mean ± σ            min … max           outliers         delta
  wall_time          42.7s  ±  604ms    42.1s  … 43.3s           0 ( 0%)        0%
  peak_rss            111MB ± 18.8KB     111MB …  111MB          0 ( 0%)        0%
  cpu_cycles          190G  ± 1.55G      189G  …  191G           0 ( 0%)        0%
  instructions        478G  ±  855K      478G  …  478G           0 ( 0%)        0%
  cache_references    552M  ± 29.6M      530M  …  585M           0 ( 0%)        0%
  cache_misses       19.6M  ±  947K     18.5M  … 20.4M           0 ( 0%)        0%
  branch_misses      11.6M  ± 48.9K     11.5M  … 11.6M           0 ( 0%)        0%
Benchmark 2 (3 runs): bench-root-scope-no-strings
  measurement          mean ± σ            min … max           outliers         delta
  wall_time          2.56s  ± 23.3ms    2.54s  … 2.59s           0 ( 0%)        ⚡- 94.0% ±  2.3%
  peak_rss            111MB ± 75.9KB     111MB …  111MB          0 ( 0%)          -  0.1% ±  0.1%
  cpu_cycles         9.39G  ± 26.2M     9.37G  … 9.42G           0 ( 0%)        ⚡- 95.0% ±  1.3%
  instructions       27.0G  ±  413K     27.0G  … 27.0G           0 ( 0%)        ⚡- 94.4% ±  0.0%
  cache_references    107M  ±  571K      106M  …  107M           0 ( 0%)        ⚡- 80.6% ±  8.6%
  cache_misses       5.33M  ± 64.1K     5.28M  … 5.40M           0 ( 0%)        ⚡- 72.7% ±  7.8%
  branch_misses      10.6M  ± 33.6K     10.5M  … 10.6M           0 ( 0%)        ⚡-  8.8% ±  0.8%
```
</details>

<details>
  <summary>Bench - total difference</summary>
  
```bash
Benchmark 1 (3 runs): bench-before
  measurement          mean ± σ            min … max           outliers         delta
  wall_time          94.9s  ±  292ms    94.6s  … 95.2s           0 ( 0%)        0%
  peak_rss           53.2MB ±  119KB    53.1MB … 53.3MB          0 ( 0%)        0%
  cpu_cycles          427G  ± 1.03G      426G  …  428G           0 ( 0%)        0%
  instructions       1.01T  ±  639K     1.01T  … 1.01T           0 ( 0%)        0%
  cache_references   1.57G  ± 28.1M     1.54G  … 1.59G           0 ( 0%)        0%
  cache_misses       46.0M  ± 4.86M     42.0M  … 51.4M           0 ( 0%)        0%
  branch_misses       141M  ± 47.9K      141M  …  141M           0 ( 0%)        0%
Benchmark 2 (3 runs): bench-final
  measurement          mean ± σ            min … max           outliers         delta
  wall_time          2.58s  ± 2.60ms    2.57s  … 2.58s           0 ( 0%)        ⚡- 97.3% ±  0.5%
  peak_rss            111MB ±  108KB     110MB …  111MB          0 ( 0%)        💩+107.9% ±  0.5%
  cpu_cycles         9.41G  ± 9.30M     9.40G  … 9.42G           0 ( 0%)        ⚡- 97.8% ±  0.4%
  instructions       27.0G  ±  162K     27.0G  … 27.0G           0 ( 0%)        ⚡- 97.3% ±  0.0%
  cache_references    107M  ± 1.19M      107M  …  109M           0 ( 0%)        ⚡- 93.2% ±  2.9%
  cache_misses       5.21M  ± 91.3K     5.12M  … 5.30M           0 ( 0%)        ⚡- 88.7% ± 16.9%
  branch_misses      10.5M  ± 11.1K     10.5M  … 10.5M           0 ( 0%)        ⚡- 92.5% ±  0.1%
```
</details>

Techatrix:
I should have mentioned that the configuration options in the benchmarking program are meant to be used with a Release Build (`zig build bench -Doptimize=ReleaseSafe`). Waiting a few minutes for benchmark results can be quite annoying.

Anyway, the removal of `lookupSymbolGlobal` made for a nice improvement!

Techatrix (approved):
Thank you again for reviving #881, your work is greatly appreciated!

PR #2057, opened 2024-10-24
fix: use tag type for calling convention

ngynkvn:
Fix for `0.14.0-dev.1983+6bf52b050`

`std.builtin.CallingConvention` was recently changed to `union(enum(u8))`, with some deprecated names as well. This swaps the field to be `CallingConvention.Tag` type instead and replaces the deprecated names with their equivalent

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -6,7 +6,7 @@ const builtin = @import("builtin");
 const zls_version = std.SemanticVersion{ .major = 0, .minor = 14, .patch = 0, .pre = "dev" };
 
 /// Specify the minimum Zig version that is required to compile and test ZLS:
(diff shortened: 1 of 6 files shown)


ngynkvn:
I guess I shouldn't have expected this to be so simple 🙃 
```
fn() align(4) callconv(.C) type␃

fn() align(4) callconv(.aarch64_aapcs_darwin) type␃

expected:
fn() align(4) callconv(.C) type
                        ^ ('\x43')
found:
fn() align(4) callconv(.aarch64_aapcs_darwin) type
                        ^ ('\x61')
```
I pushed up another change to adjust the test cases to fit the type change I made, but I don't think this will make them pass as per above. Feel free to close this if I'm chasing a red herring

PR #2058, opened 2024-10-25
wrap number hover in code block

xdBronch:
without this zed would put everything on 1 line, arguably this is probably something they should fix but the `md` also gives the numbers highlighting :man_shrugging:. didnt notice this make anything worse on nvim or helix
image

image

### DIFF
--- a/src/Server.zig
+++ b/src/Server.zig
@@ -1558,7 +1558,15 @@ fn hoverHandler(server: *Server, arena: std.mem.Allocator, request: types.HoverP
     var analyser = server.initAnalyser(handle);
     defer analyser.deinit();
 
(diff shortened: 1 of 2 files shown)


nektro:
vscode assumes all comment text is markdown

xdBronch:
sorry but whats your point? does this make anything worse in vscode or just pointing out its not necessary there? i dont have it installed so i cant test.

xdBronch:
sorry for the delay, you commented just after i ran `cargo clean` on zed and i was putting off rebuilding since the dependency tree is like 1300 crates lol

Techatrix:
> sorry for the delay, you commented just after i ran `cargo clean` on zed and i was putting off rebuilding since the dependency tree is like 1300 crates lol

I fell your pain. I too had my fun compiling all those crates.

Techatrix (approved):

Issue #1989, opened 2024-08-05
No type inference when orelse RHS is a block


Vexu:
### Zig Version

0.14.0-dev.564+75cf7fca9

### Zig Language Server Version

dfe958e4d1a9ba503cfafd39164f090daea70109

### Client / Code Editor / Extensions

vscode

### Steps to Reproduce and Observed Behavior

```zig
test {
    var s: ?u32 = undefined;
    var a = s orelse {
        return;
    };
    var b = s orelse return;
    var c = s.?;
    _ = .{ &s, &a, &b, &c };
}
```

### Expected Behavior

`a`, `b`, and `c` should all be inferred to be `u32`s but `a` is not.

7phs:
The issue should now be resolved after merging my MR.

Techatrix:
fixed by #2084

PR #2084, opened 2024-11-12
Resolve return type of `orelse` with block on RHS (with return)

7phs:
Fixes completion issues for cases with `orelse` and `return` (or block) expressions.

Currently, only the right side of an `orelse` expression is analyzed, which lacks type information in cases where `orelse` is used with block (`return` inside). This PR addresses the issue by ensuring that the left side of the expression is also analyzed.

This fix resolves issue #1989 .


src/analysis.zig
```diff
@@ -1610,8 +1610,15 @@ fn resolveTypeOfNodeUncached(analyser: *Analyser, node_handle: NodeWithHandle) e
...
-                    return try analyser.resolveOrelseType(base_type, type_right);
+                    const type_right = try analyser.resolveTypeOfNodeInternal(.{ .node = datas[node].rhs, .handle = handle });
+                    if (type_right) |tr| {
+                        return try analyser.resolveOrelseType(base_type, tr);
+                    }
+                    const type_left = try analyser.resolveTypeOfNodeInternal(.{ .node = datas[node].lhs, .handle = handle }) orelse return null;
+                    return try analyser.resolveOrelseType(base_type, switch (type_left.data) {
+                        .optional => |o| o.*,
+                        else => type_left,
+                    });
```
Techatrix, line 1621: The left hand side of the expression is actually being resolved in the old code. See the definition of `base_type`. `type_left` and `base_type` are the same which means that the code should be equivalent to:
```suggestion
                    const type_right = try analyser.resolveTypeOfNodeInternal(.{ .node = datas[node].rhs, .handle = handle }) orelse return try analyser.resolveOptionalUnwrap(base_type);
                    return try analyser.resolveOrelseType(base_type, type_right);
```

Techatrix (changes requested):

src/analysis.zig
7phs: Thank you for the suggestion!

Applied.

PR #2086, opened 2024-11-17
check whether the required Linux release is present for std.build.Watch

### LINKED ISSUE #2040: fanotify_init failed due to old kernel; requires 5.17+

Techatrix:
@mocompute This should fix #2040 without requiring a change in Zig. Can you let me know if this resolves your issue?

fixes #2040
closes #2041

### DIFF
--- a/src/build_runner/0.12.0.zig
+++ b/src/build_runner/0.12.0.zig
@@ -413,9 +413,16 @@ pub fn main() !void {
         seed,
     );
 
-    const watch_suported = comptime switch (builtin.os.tag) {
-        .linux => builtin.zig_version.order(file_watch_version) != .lt,
-        .windows => builtin.zig_version.order(file_watch_windows_version) != .lt,
+    const watch_suported = switch (builtin.os.tag) {
+        .linux => blk: {
+            if (comptime builtin.zig_version.order(file_watch_version) == .lt) break :blk false;
+
+            // std.build.Watch requires `FAN_REPORT_TARGET_FID` which is Linux 5.17+
+            const utsname = std.posix.uname();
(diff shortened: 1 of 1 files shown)


mocompute:
@Techatrix It seems to, though I can't quite get the right combination of versions to reproduce the fanotify error again on my current system. But the fix looks reasonable. Many thanks!

Issue #2060, opened 2024-10-25
Check diagnostics not highlighting full type issue


iamcdonald:
### Zig Version

0.14.0-dev.2034+56996a280

### ZLS Version

0.14.0-dev.179+5acc066

### Client / Code Editor / Extensions

Zed 0.158.2

### Steps to Reproduce and Observed Behavior

Can recreate with this [repo](https://github.com/iamcdonald/zed-syntax-highlighting-test).
- open `test.zig`
- change `i32` to another wrong type i.e. `i24`
- only the `i` of the wrong type (`i32`) is highlighted rather than the full type

I'm finding this behavior is the same for all diagnostic responses i.e. any issue is apportioned to only one character of the offending token.

### Expected Behavior

I expected the entire `i32` to be underlined but only the first character is.
My assumption would be that the `end` in the diagnostics message should maybe be `{"line":3,"character":37}` but I could be wrong.

### Relevant log output

```shell
debug (message): received: {"jsonrpc":"2.0","method":"textDocument/didSave","params":{"textDocument":{"uri":"file:///Users/ooooo/Documents/code/zig-audio/zed-syntax-highlighting-test/src/test.zig"},"text":"const std = @import(\"std\");\n\ntest \"test\" {\n    const x: i16 = std.math.maxInt(i24);\n    _ = x;\n}\n"}}
debug (server): Took 0ms to process notification-textDocument/didSave on Thread 1781099
info  ( diag ): Running build-on-save: file:///Users/ooooo/Documents/code/zig-audio/zed-syntax-highlighting-test/build.zig (check)
debug (message): sent: {"jsonrpc":"2.0","method":"textDocument/publishDiagnostics","params":{"uri":"file:///Users/ooooo/Documents/code/zig-audio/zed-syntax-highlighting-test/src/test.zig","diagnostics":[{"range":{"start":{"line":3,"character":35},"end":{"line":3,"character":35}},"severity":1,"code":"zig_build","source":"zls","message":"type 'i16' cannot represent integer value '8388607'","relatedInformation":[]}]}}
```

### Further Digging
From looking through the `zls` code the culprit appears to be [here](https://github.com/zigtools/zls/blob/5acc06679321aeff5f6d2f05ebb52520f734a8b8/src/features/diagnostics.zig#L331-L338). 

As mentioned in the comment it would appear that (given we're interpreting the `zig build check` output) a fix would be
- read the src file
- tokenize using `std.zig.Tokenizer`
[...]

Issue #2018, opened 2024-08-31
use reference trace to provide better build-on-save source locations for errors


Techatrix:
Example:
```zig
const std = @import("std");

pub fn main() !void {
    std.log.info("{}", .{""});
}
```
Output:
```
$ zig run foo.zig -freference-trace=12
/nix/store/li7vdsrycv9zgp0phkxgsgn9fayn7cx0-zig-0.14.0-dev.1371+5723fcaac/lib/std/fmt.zig:663:17: error: cannot format array without a specifier (i.e. {s} or {any})
                @compileError("cannot format array without a specifier (i.e. {s} or {any})");
                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
referenced by:
    formatType__anon_9361: /nix/store/li7vdsrycv9zgp0phkxgsgn9fayn7cx0-zig-0.14.0-dev.1371+5723fcaac/lib/std/fmt.zig:627:38
    format__anon_6673: /nix/store/li7vdsrycv9zgp0phkxgsgn9fayn7cx0-zig-0.14.0-dev.1371+5723fcaac/lib/std/fmt.zig:188:23
    print__anon_3923: /nix/store/li7vdsrycv9zgp0phkxgsgn9fayn7cx0-zig-0.14.0-dev.1371+5723fcaac/lib/std/io/Writer.zig:24:26
    defaultLog__anon_2963: /nix/store/li7vdsrycv9zgp0phkxgsgn9fayn7cx0-zig-0.14.0-dev.1371+5723fcaac/lib/std/io.zig:324:47
    log__anon_2562: /nix/store/li7vdsrycv9zgp0phkxgsgn9fayn7cx0-zig-0.14.0-dev.1371+5723fcaac/lib/std/log.zig:125:22
    info__anon_1614: /nix/store/li7vdsrycv9zgp0phkxgsgn9fayn7cx0-zig-0.14.0-dev.1371+5723fcaac/lib/std/log.zig:194:16
    main: foo.zig:4:17
    posixCallMainAndExit: /nix/store/li7vdsrycv9zgp0phkxgsgn9fayn7cx0-zig-0.14.0-dev.1371+5723fcaac/lib/std/start.zig:615:37
[...]

PR #2096, opened 2024-11-29
new build on save with watch mode

### LINKED ISSUE #2001: Provide a configuration option to use incremental compilation for build-on-save
After ziglang/zig#21063 is merged, the Zig compiler is able to perform certain incremental updates correctly. There's still a ways to go before it's generally usable, but `-fno-emit-bin` builds will be the first thing to work correctly, and some do under that PR. For instance, simple incremental updates to the compiler itself typically do the right thing now.

In the future, incremental compilation will be able to help the Zig compiler expose detailed semantic information to ZLS or a similar project (ziglang/zig#615). However, ZLS can still make good use of incremental compilation before upstream progress is made there.

If ZLS' build runner introduces support for `std.Build.Watch` like the standard build runner does, which should be a relatively straightforward enhancement, it can begin using incremental compilation for build-on-save. (Incremental compilation currently only works if the compiler process stays running via the compiler server protocol: serialization and deserialization of compiler state to disk is a WIP feature.)

This functionality should be behind a config option which can be disabled by default. It's definitely possible to crash the compiler right now, or even to get false positive compile errors; incremental compilation is not stable, even when not emitting a binary. When we *are* emitting a binary, it'll basically crash immediately, so this will currently only work if the user is providing via `build_on_save_step` the name of a step which will pass `-fno-emit-bin` to the compiler. Nevertheless, it might be valuable to expose this as an option for people to try out.

### LINKED ISSUE #2095: Diagnostics relatedInformation contain invalid URLs
### Zig Version

latest

### ZLS Version

latest

kakoune

Open a Zig file containing

```zig
pub fn main() void {}
pub fn main() void {}
```

zig-lsp sends diagnostics with `"uri":"/"` which seems invalid?

For more details, see the original report at https://github.com/kakoune-lsp/kakoune-lsp/issues/811

### Expected Behavior

`"uri"://file/path/to/file.zig"`

### Relevant log output

LSP: DEBG From server zls: {"jsonrpc":"2.0","method":"textDocument/publishDiagnostics","params":{"uri":"file:///tmp/afac/main.zig","diagnostics":[{"range":{"start":{"line":0,"character":7},"end":{"line":0,"character":11}},"severity":1,"code":"ast-check","source":"zls","message":"duplicate struct member name 'main'","relatedInformation":[{"location":{"uri":"/","range":{"start":{"line":1,"character":7},"end":{"line":1,"character":11}}},"message":"duplicate name here"},{"location":{"uri":"/","range":{"start":{"line":0,"character":0},"end":{"line":0,"character":3}}},"message":"struct declared here"}]}]}}

Techatrix:
This PR replaces the current build on save implementation with a new one that is based on Zig's `zig build --watch` feature. This can be combined with the `build_on_save_args` config option to enable incremental compilation. fixes #2001

Here is a small showcase where I use the equivalent of `zig build check --watch -fincremental`:

<details>

[Showcase](https://github.com/user-attachments/assets/920d432e-6d4e-4f62-8163-345ebffadb32)

<summary>Show Video</summary>

I used the [Error Lens](https://marketplace.visualstudio.com/items?itemName=usernamehw.errorlens) extension to display the error messages inline.

</details>

The minimum runtime Zig version has been bumped from `0.12.0` to `0.14.0-dev.310+9d38e82b5`. This means that the latest mach nominated zig version (`0.14.0-dev.1911+3bf89f55c`) is still supported.
The build on save diagnostics are no longer removed after editing the document. The source location of the error message is not being updated when the document gets modified but this can be implement in the future.

- fixes  #2095
- partially addresses #2060
- errors are stored as `std.zig.ErrorBundle` which unblocks #2018

I have only tested this on Linux and Windows with VS Code. Who know what kind of bugs are still lurking around :shrug:


leecannon:
I've tried this with a few of my projects and it works well :zap:

lishaduck:
@Techatrix, is there a way to turn this off? I skimmed the diff and I didn't see one.
This breaks build_on_save in VS Code on macOS (#2221/https://github.com/ziglang/zig/issues/21905)

kristoff-it:
On macOS there are two bugs related to this feature, one is the one linked above, the second is that --watch crashes on macOS ([fixed here](https://github.com/ziglang/zig/pull/23247), should become available in 0.14.1)

Techatrix:
> @Techatrix, is there a way to turn this off? I skimmed the diff and I didn't see one.

No, this PR has **replaced** the previous build on save implementation. Excluding bugs like these, there should be no reason to need the old implementation.

If no progress is being made on https://github.com/ziglang/zig/issues/21905, I will look into patching ZLS.

PR #2097, opened 2024-11-30
Code action to convert string literal

Sekky61:
# Code action to rewrite string literal <-> multiline string

The code action will be available if the cursor is _inside_ the string literal and if it is convertible.

Code action name in both directions: `refactor.convertStringLiteral`.

Tested in: Neovim, VS Code

## Converting string literal -> multiline string

Multiline string literals have no escapes, so strings with some escapes cannot be converted:

| Escape Sequence |                                Name                               | Convertible to multiline |
|:---------------:|:-----------------------------------------------------------------:|--------------------------|
| `\n`            | Newline                                                           |                        ✅|
| `\r`            | Carriage Return                                                   |                        ❎|
| `\t`            | Tab                                                               |                        ❎|
| `\\`            | Backslash                                                         |                        ✅|
| `\'`            | Single Quote                                                      |                        ✅|
| `\"`            | Double Quote                                                      |                        ✅|
| `\xNN`          | hexadecimal 8-bit byte value (2 digits)                           |                        ❓|
| `\u{NNNNNN}`    | hexadecimal Unicode scalar value UTF-8 encoded (1 or more digits) |                        ❓|

It is possible to convert at least some of the hex values to the utf-8 codes.

<details>
<summary>string literal -> multiline string example</summary>

```zig
const foo = "Hello\nWorld";
```

```zig
const foo = 
\\Hello
\\World
;
```

</details>

## Converting multiline string -> string literal

Escapes must be added in some cases.
Newline `\n` must be added for every line of the multiline string except for the last one.

From ascii select characters can be found in the string, which makes things easier.
Quoting the tokenizer for multiline string:
```zig
    0x01...0x09, 0x0b...0x0c, 0x0e...0x1f, 0x7f => continue :state .invalid,
```

<details>
    <summary>multiline string -> string literal</summary>

```zig
const foo = 
\\1. Eggs
\\2. "Bacon"
\\
;
```

```zig
const foo = "1. Eggs\n2. \"Bacon\"\n";
```

</details>

## Discussion

The code action does not appear when cursor is on the markup (`\\` or `"`).
I would like to fix this because this way it is not possible to use the action on empty strings.
It is also less convenient for the user.
The position context might be updated to not exclude the markup, but there might be some consequences.

Other notes:
- test names, `@import` arguments, `pub extern "something"` cannot be multiline strings (`test "addition" { ... }`)
- Invalid escape (like `" \ "`) is not a concern because such source code cannot be tokenized, therefore code actions are not generated
- It doesn't interact with chars (`'q'`) - it is a different type
- `std.zig.stringEscape` could not be used because it escapes unicode (emojis etc.) as hexcodes, which I do not like
- Relevant PR: [https://github.com/ziglang/zig/issues/19299](https://github.com/ziglang/zig/issues/19299)
- Replaces #1931

Questions:
- Did I understand `\r` correctly?
- Should indentation be part of the produced edit, or does fmt take care of it afterwards? Right now there is a hardcoded 4 space indent
- Performance-wise, there might be a lot of `appendSlice` calls in a loop. An upper bound of the final string's length could be determined and preallocated.

### DIFF
--- a/src/features/code_actions.zig
+++ b/src/features/code_actions.zig
@@ -149,7 +149,7 @@ pub fn generateStringLiteralCodeActions(
     };
     // Check for disallowed characters and utf-8 validity
     for (parsed) |c| {
-        if (c == '\n' or c == '\r') continue;
+        if (c == '\n') continue;
         if (std.ascii.isControl(c)) return;
     }
     if (!std.unicode.utf8ValidateSlice(parsed)) return;
--- a/src/features/code_actions.zig
+++ b/src/features/code_actions.zig
@@ -75,6 +75,29 @@ pub const Builder = struct {
         return only_kinds.contains(kind);
     }
 
+    pub fn generateCodeActionsInRange(
+        builder: *Builder,
+        range: types.Range,
+        actions: *std.ArrayListUnmanaged(types.CodeAction),
+    ) error{OutOfMemory}!void {
+        const tree = builder.handle.tree;
+        const source_index = offsets.positionToIndex(tree.source, range.start, builder.offset_encoding);
+
+        const token_idx = offsets.sourceIndexToTokenIndex(tree, source_index);
+        const token_tags = tree.tokens.items(.tag);
+        const position_token = token_tags[token_idx];
+
+        const ctx = try Analyser.getPositionContext(builder.arena, builder.handle.tree, source_index, true);
+
+        switch (ctx) {
+            .string_literal => switch (position_token) {
+                .string_literal => try generateStringLiteralCodeActions(builder, token_idx, actions),
+                else => {},
+            },
+            else => {},
+        }
+    }
+
     pub fn createTextEditLoc(self: *Builder, loc: offsets.Loc, new_text: []const u8) types.TextEdit {
         const range = offsets.locToRange(self.handle.tree.source, loc, self.offset_encoding);
         return types.TextEdit{ .range = range, .newText = new_text };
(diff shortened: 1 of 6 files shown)


src/features/code_actions.zig
```diff
@@ -93,6 +118,104 @@ pub const Builder = struct {
...
+    const parsed = std.zig.string_literal.parseAlloc(builder.arena, token_text) catch |err| switch (err) {
+        error.InvalidLiteral => return,
+        else => |other| return other,
+    };
+    // Check for disallowed characters
+    for (parsed) |c| {
+        switch (c) {
+            0x01...0x09, 0x0b...0x0c, 0x0e...0x1f, 0x7f => return,
+            else => {},
+        }
```
Techatrix, line 143: Something like this should be more readable:
```suggestion
        if (c == '\n' or c == '\r') continue;
        if (std.ascii.isControl(c)) return;
```

Also, read my final review comment about carriage returns.

src/features/code_actions.zig
```diff
@@ -93,6 +118,104 @@ pub const Builder = struct {
...
+        var i = offsets.tokenToLoc(builder.handle.tree, @intCast(end - 1)).end + 1;
+        const source = builder.handle.tree.source;
+        while (i < source.len) : (i += 1) {
+            switch (source[i]) {
+                '\n', '\r' => {},
+                else => break,
+            }
+        }
+        break :blk i;
+    };
```
Techatrix, line 208:
```suggestion
    const last_token_end = std.mem.indexOfNonePos(
        u8,
        builder.handle.tree.source,
        offsets.tokenToLoc(builder.handle.tree, @intCast(end - 1)).end + 1,
        "\n\r",
    ) orelse builder.handle.tree.source.len;
```

Techatrix (changes requested):
> The code action does not appear when cursor is on the markup (`\\` or `"`).
> I would like to fix this because this way it is not possible to use the action on empty strings.
> It is also less convenient for the user.
> The position context might be updated to not exclude the markup, but there might be some consequences.

Looking at all uses of `Analyser.PositionContext`, the only place where `Analyser.PositionContext.string_literal` is being used is for filesystem completions inside string literals in `build.zig` files. https://github.com/zigtools/zls/pull/1668
As long as that continues to work, you are free to adjust `getPositionContext` to return `.string_literal` as needed.

> * Did I understand `\r` correctly?

You should take a look at https://github.com/ziglang/zig-spec/issues/38. TLDR a multi-line string literal can't contain carriage returns.

> Should indentation be part of the produced edit, or does fmt take care of it afterwards? Right now there is a hardcoded 4 space indent

Usually the user will have to invoke formatting manually after applying the code action (usually with format on save). Ideally we would produce code that is already valid to `zig fmt` but it isn't easy to cover all cases.

---

I would suggest to reject string literals that contain `\xNN` or `\u{NNNNNN}` escape sequences. They can be used to produce invalid UTF-8:

const foo = "\xaa";

I tested this in VS Code and it showed the "convert to a multiline string literal" code action but applying it did nothing.

Techatrix (changes requested):
You should also add a `if (!builder.wantKind(.{ .custom_value = "refactor.convertStringLiteral" })) return;` check. A tracy zone for each function would also be nice.

See https://github.com/zigtools/zls/pull/2087

Sekky61:
Thanks for the feedback. I deviated in one point: I added a [utf8ValidateSlice](https://ziglang.org/documentation/master/std/#std.unicode.utf8ValidateSlice) which I think could catch most of the conversion problems.

Techatrix (commented):
I followed up to c129bd341f7babe5e51676ea2a4e733700731894 so that the remaining test cases work as well. This means that the string literal code action should work in the following situation:
```zig
const foo = "hello world"<cursor>;
```
There was also one change missing where carriage returns should be rejected when converting to multi-line string literals. I made the change myself.

I am currently running some [fuzzing](https://github.com/zigtools/sus) take make sure that no regressions have been introduced to the position context. If that doesn't report any issue, this is ready to be merged.

PR #2111, opened 2024-12-17
update build runner for latest zig master

leecannon:
Update the build runner to work after https://github.com/ziglang/zig/pull/22243.


Techatrix:
[zig-overlay](https://github.com/mitchellh/zig-overlay/commits/main/) (`0.14.0-dev.2502+d12c0bf90` ) is currently behind [ziglang.org](https://ziglang.org/download/) (`0.14.0-dev.2509+0c457fa7e`) so the minimum build version can be satisfied. There shouldn't be any need to increase the minimum build version since the change is backwards compatible.

PR #2127, opened 2024-12-31
escape identifiers when renaming

### LINKED ISSUE #2124: Renamed identifier is not wrapped with `@""` when required
### Zig Version

0.14.0-dev.2548+0f17cbfc6

### ZLS Version

0.14.0-dev.325+8406f8c7

VsCode 1.96.0

```zig
// file.zig
const a_valid_name = 1;
```
Rename `a_valid_name` to `an invalid name`.
```zig
const an invalid name = 1;
```

This is especially annoying when using identifiers that shadow keywords or primitive values.
Workaround is to manually wrap the identifier with `@""` when renaming. 

### Expected Behavior

```zig
const @"an invalid name" = 1;
```

### Relevant log output

_No response_

xdBronch:
closes #2124
this isnt completely ideal since it always escapes primitives despite those being valid field names but `zig fmt` automatically canonicalizes those so... its fine i suppose?

### DIFF
--- a/src/features/references.zig
+++ b/src/features/references.zig
@@ -497,13 +497,14 @@ pub fn referencesHandler(server: *Server, arena: std.mem.Allocator, request: Gen
 
     switch (request) {
         .rename => |rename| {
+            const escaped_rename = try std.fmt.allocPrint(arena, "{}", .{std.zig.fmtId(rename.newName)});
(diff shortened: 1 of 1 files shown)


Techatrix (approved):
> this isnt completely ideal since it always escapes primitives despite those being valid field names but zig fmt automatically canonicalizes those so... its fine i suppose?

Ideally we wouldn't rely on `zig fmt` but this is good enough for now.

PR #2131, opened 2025-01-04
embrace RLS even harder

xdBronch:


src/DocumentStore.zig
```diff
@@ -184,7 +184,7 @@ pub const Handle = struct {
     /// private field
     impl: struct {
         /// @bitCast from/to `Status`
-        status: std.atomic.Value(u32) = .init(@bitCast(Status{})),
+        status: std.atomic.Value(u32) = .init(0),
```
xdBronch, line 187: was a little bit iffy on this one, it arguably loses a bit of meaning but im not sure it matters too much?

src/DocumentStore.zig
```diff
@@ -234,14 +234,14 @@ pub const Handle = struct {
...
+        zir_outdated: bool = false,
         /// true if a thread has acquired the permission to compute the `std.zig.Zoir`
         has_zoir_lock: bool = false,
         /// all other threads will wait until the given thread has computed the `std.zig.Zoir` before reading it.
         /// true if `handle.impl.zoir` has been set
         has_zoir: bool = false,
-        zoir_outdated: bool = undefined,
-        _: u23 = undefined,
+        zoir_outdated: bool = false,
+        _: u23 = 0,
```
xdBronch, line 244: not related to RLS at all but defaulting to undefined always feels really gross to me so this is more personal opinion than an objective change, fine with reverting it along with the above change though.

src/DocumentStore.zig
Techatrix: initalizing to zero assumes that `Status{}` is all zeroes. This may not be the case in the future so I would prefer the old code.

src/DocumentStore.zig
Techatrix: This is done because the `[zir;zoir]_outdated` field should never be accessed before checking `has_[zir;zoir]`. I don't mind this change so feel free to keep it. What should be reverted however is the change from `_: u23 = undefined,` to `_: u23 = 0,`. This field should never be accessed so there is no need to initalize it.

xdBronch: seems reasonable, part of why i changed these was so that i could change the `.init` in the other comment but no need anymore

src/DocumentScope.zig
```diff
@@ -493,10 +493,10 @@ const ScopeContext = struct {
             .loc = loc,
             .parent_scope = context.current_scope,
             .child_scopes = .{
-                .small = [_]Scope.OptionalIndex{.none} ** Scope.ChildScopes.small_size,
+                .small = @splat(.none),
             },
             .child_declarations = .{
-                .small = [_]Declaration.OptionalIndex{.none} ** Scope.ChildDeclarations.small_size,
+                .small = @splat(.none),
```
xdBronch, line 499: just wanna acknowledge how great this one is, thanks mlugg

Techatrix (approved):

Issue #1392, opened 2023-08-08
incorrect completions on nested generic type like std.ArrayList


CGQAQ:
### Zig Version

0.12.0-dev.1+155c27f   build with -Ddata-version="0.11.0"

### Zig Language Server Version

0.12.0-dev.28+81bfc90

### Steps to Reproduce

code completion of `std.ArrayList` is completely broken

### Expected Behavior

should give me either `x` or `y` of `struct Foo`

### Actual Behavior

![image](https://github.com/zigtools/zls/assets/15936231/03ff6557-60e8-441b-a405-2713995cdcf8)
![image](https://github.com/zigtools/zls/assets/15936231/73c8ab48-33d6-422f-9496-8a2203092e71)

Sometimes will EVEN give the wrong complete items
![image](https://github.com/zigtools/zls/assets/15936231/3071ff7b-38e2-4980-a0bf-cad8534bae19)
But my `collected_data_list` is `ArrayList` of `ArrayList`
![image](https://github.com/zigtools/zls/assets/15936231/7199479d-aec4-4416-a71e-5b56a17a4327)

### P.S.
Other code completion works
![image](https://github.com/zigtools/zls/assets/15936231/c6d24b20-5e87-460c-bc98-b66905a3a397)

<details><summary>The code</summary>
<p>

```zig
const std = @import("std");

pub fn main() !void {
    const allocator = std.heap.page_allocator;

    const Foo = struct {
        x: u8,
        y: u8,
    };
    
    const ArrayList = std.ArrayList;
    const list_of_list = ArrayList(Foo).init(allocator);
    defer list_of_list.deinit();

    list_of_list.

    for (list_of_list.items) |list| {
        _ = list;
        // list.
    }
}

``` 

</p>
</details>

nullptrdevs:
Would you happen to be using Copilot?

CGQAQ:
> Would you happen to be using Copilot?

Yes, I am using Copilot

Xefreh:
did you tried latest commit, code completion was not working for me before but right now it does

CGQAQ:
> did you tried latest commit, code completion was not working for me before but right now it does

no, just checkout the latest commit same for me, and I disabled copilot as well

nullptrdevs:
There's nothing wrong w/ Copilot, it's just that I knew ZLS doesn't provide completions for any of those but `list_of_list.`
[...]

Issue #1841, opened 2024-03-28
Allocator usage causes zls inline types to get very confused


Beyley:
### Zig Version

0.12.0-dev.3439+31a7f22b8

### Zig Language Server Version

0.12.0-dev.496+96eddd0

### Client / Code Editor / Extensions

VSCode with official extension

### Steps to Reproduce and Observed Behavior

Put this code into a file
```
const std = @import("std");

pub fn main() !void {
    const one = try std.heap.c_allocator.alloc(u16, 10);
    _ = one; // autofix
    const two = try std.heap.c_allocator.alloc(u8, 10);
    _ = two; // autofix
}
```

You'll see ZLS incorrectly show the type of `two` as `[]u16`

This behaviour gets very weird when its spread across multiple functions				
https://github.com/zigtools/zls/assets/34501060/c0f9498b-a24e-403b-921e-b026b93a516c

### Expected Behavior

The type of all the alloc calls is correctly resolved

Sekky61:
I think I got to the root of this. The cause is caching of the node types ([link to code](https://github.com/zigtools/zls/blob/93b7bbd0d96ee7088c46d1373a2810e84f7d47be/src/analysis.zig#L1235C5-L1235C83)). The cache does not take into account the fact that the return type depends on the arguments to the function.

Here is the minimal reproduction:
```zig
fn foo(comptime T: type) T {
    return 42;
}

const bar = foo(u1); // hints `u1`
const baz = foo(u2); // hints `u1`
```

![Selection_117](https://github.com/zigtools/zls/assets/24541288/a52c933e-6d71-4380-ad9c-08b03c59e9f0)

In this code, the type of `foo` gets resolved once for `bar`, but for `baz`, the type of `foo` is already in cache.

Now what could be the solution here?
- Add function param types to the cache's key
- Do not cache function calls
- Do not cache function calls with type variables
- Some kind of change to `ComptimeInterpreter` (I have not studied it yet)

Sekky61:
Another observation: the reason why hover works most of the time is because the `Analyser` is instantiated for each request (with clean cache). So this bug can happen on hover as well, if the same type needs to be evaluated multiple times within the single hover resolution.

```zig
fn arr(comptime T: anytype) []T {
    return &[_]T{};
}

// The deeper `arr` is evaluated, the outer `arr` is from cache
// should be [][]u1, but both hint and hover is []u1
pub const hover = arr(@TypeOf(arr(u1)));

```

PR #2133, opened 2025-01-04
track comptime types

mfield:
Adds some tracking of comptime types.

changes formatting for comptime structs, e.g. `HashMap(...)` -> `HashMap(i32,void)`. maybe just a preference of mine? but i like seeing it. could easily change the formatter back to `...`.

fixes:
  - repeated use of the same comptime function #1841
  - nested generic structs e.g. #1392


mfield:
sorry for the false start. saw something while reading the diff and made a small change last second without re-building :(
new version is passing tests again!

Techatrix (changes requested):
The analysis backend has an existing mechanism to resolve comptime function type parameters. See [`bound_type_params`](https://github.com/zigtools/zls/blob/f961c8b2d0276d5df93da910c55f0ab5b09c2053/src/analysis.zig#L31). I would have expected this field to be replaced instead of adding another solution on top of it. Removing `bound_type_params` will make most of the existing test fail that deal with generic functions which indicates that this PR is failing to handle them.

Also, have you tried to adding the extra properties of `BoundScope` to `ScopeWithHandle` instead of adding `.bound_scope` to `Type.Data`? If you need to wrap a container with a bound scope, just create a copy of the existing container with extra bound scopes. I tried to test this change myself but could not quite get it to work so there may be a good reason to keep it this way.

I will hold of with additional review comments until the fundamentals are in a good shape. The idea of keeping a stack of bound parameter types is also what I had in mind when thinking about how to resolve #1392.

mfield:
Thanks for the review. Finally had some time to try out your suggestions and got the two big ones working:
- `bound_type_params` is removed
- `bound_params` moved from `BoundScope` to `ScopeWithHandle`

I'm not terribly fond of my change to `DeclWithHandle` nor pushing and popping comptime_state from completions.zig. Would be down to iterate again there if you have any suggestions. The two-pass `collectDeclarationsOfContainer` then `declToCompletion` for each decl collected does make some sense to draw a nice line between analysis.zig and completions.zig so I wasn't sure how to keep comptime_state discovered during `collectDeclarationsOfContainer` available to `declToCompletion`...

Techatrix (changes requested):
Apologies for the late response. I have a bunch of comments that are mostly refactoring suggestions but overall these changes look good to me.

About the changes in `completions.zig`. To avoid them, I tried the following change:

<details>

<summary><code>patch.diff</code></summary>

```diff
diff --git a/src/analysis.zig b/src/analysis.zig
index 8f53529..cf1ad7b 100644
--- a/src/analysis.zig
+++ b/src/analysis.zig
@@ -4141,6 +4141,16 @@ pub const DeclWithHandle = struct {
         const tracy_zone = tracy.trace(@src());
         defer tracy_zone.end();
 
+        const starting_depth = analyser.comptime_state.depth();
+        var pushed = false;
+        if (self.from) |from| {
+            if (from.bound_params.len > 0) {
+                try analyser.comptime_state.push(analyser.gpa, from.bound_params);
+                pushed = true;
+            }
+        }
+        defer if (pushed) analyser.comptime_state.pop(starting_depth);
+
         const tree = self.handle.tree;
         const resolved_ty = switch (self.decl) {
             .ast_node => |node| try analyser.resolveTypeOfNodeInternal(
@@ -4348,6 +4358,10 @@ pub fn collectDeclarationsOfContainer(
     const scope = container_scope.scope;
     const handle = container_scope.handle;
 
+    const starting_depth = analyser.comptime_state.depth();
+    try analyser.comptime_state.push(analyser.gpa, container_scope.bound_params);
+    defer analyser.comptime_state.pop(starting_depth);
+
     const tree = handle.tree;
     const document_scope = try handle.getDocumentScope();
     const container_node = container_scope.toNode();
diff --git a/src/features/completions.zig b/src/features/completions.zig
index 2374702..9bf510b 100644
--- a/src/features/completions.zig
+++ b/src/features/completions.zig
@@ -103,9 +103,6 @@ fn typeToCompletion(builder: *Builder, ty: Analyser.Type) error{OutOfMemory}!voi
             });
         },
         .container => |scope_handle| {
-            const starting_depth = builder.analyser.comptime_state.depth();
-            try builder.analyser.comptime_state.push(builder.analyser.gpa, scope_handle.bound_params);
-            defer builder.analyser.comptime_state.pop(starting_depth);
             var decls: std.ArrayListUnmanaged(Analyser.DeclWithHandle) = .empty;
             try builder.analyser.collectDeclarationsOfContainer(scope_handle, builder.orig_handle, !ty.is_type_val, &decls);
 
@@ -173,15 +170,6 @@ fn declToCompletion(builder: *Builder, decl_handle: Analyser.DeclWithHandle, opt
         doc_comments.appendAssumeCapacity(docs);
     }
 
-    const starting_depth = builder.analyser.comptime_state.depth();
-    var pushed = false;
-    if (decl_handle.from) |from| {
-        if (from.bound_params.len > 0) {
-            try builder.analyser.comptime_state.push(builder.analyser.gpa, from.bound_params);
-            pushed = true;
-        }
-    }
-    defer if (pushed) builder.analyser.comptime_state.pop(starting_depth);
     const maybe_resolved_ty = try decl_handle.resolveType(builder.analyser);
 
     if (maybe_resolved_ty) |resolve_ty| {
```

</details>

It only fails a completion test about `usingnamespace`. I haven't investigated it further but it seems promising. If this turns out be a dead end or to complex, I would be fine to keep the change to completions.zig as is.

Techatrix:
@mfield Do you plan to apply the suggested changes anytime soon? If not, I'd be happy to apply them myself. This PR is a very nice improvement that would be unfortunate to miss out on.

mfield:
Hi, so sorry, completely lost track of this.

I tried your approach to moving `bound_type_params` out of completions and into analysis. The `usingnamespace` case is problematic enough I've backed out of it as well. The difficulty, as I see it, is that resolving the type of a function is too shallow. The `Type` that comes back doesn't describe the types of the parameters or the return type. So you have to remember to apply the same `bound_type_params` context to any operations that involve the function type, e.g. `firstParamIs`.

I think I've applied all the other suggestions. Please let me know if I've missed any.

Techatrix (approved):
The applied changes look all good to me. Thank you very much.

> I tried your approach to moving `bound_type_params` out of completions and into analysis. [...]

Makes sense, this can be revisited with the [removal of usingnamespace](https://github.com/ziglang/zig/issues/20663).

Issue #1201, opened 2023-05-26
Support process report


junnplus:
https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#progress

SuperAuguste:
There's no real use for this at the moment as we do everything in a single thread. A good start could be to split out our build file compilation logic into a worker pool and use this for that.

junnplus:
Actually, this issue originated from:

> Possibly during the first request, zls was not ready yet. In the previous design, the `LspProgressUpdate` event notification was used to trigger a second request. However, it appears that zls did not notify properly. https://github.com/zigtools/zls/blob/33121dacf9f7a178b40418840f5d66f9a9004992/src/Server.zig#L1423

_Originally posted by @junnplus in https://github.com/junnplus/lsp-setup.nvim/issues/49#issuecomment-1563684632_

junnplus:
I carefully examined the nvim log and read through the relevant code of zls. I found that when requesting `textDocument/inlayHint` for the first time, `enable_inlay_hints` was false, which resulted in an empty result being returned.

This is my configuration.
```lua
settings = {
    zls = {
        enable_inlay_hints = true,
        inlay_hints_show_builtin = true,
        inlay_hints_exclude_single_argument = true,
        inlay_hints_hide_redundant_param_names = true,
        inlay_hints_hide_redundant_param_names_last_token = true,
    }
}
```

nvim will send a `workspace/didChangeConfiguration` request before requesting inlay hints, but zls cannot correctly recognize the configuration with `zls` sections when processing this request.

zls will initiate a `workspace/configuration` request later to obtain the configuration, which will require viewing the configuration with the `zls` section.

```
[DEBUG][2023-05-28 17:01:16] .../vim/lsp/rpc.lua:387	"rpc.receive"	{
  id = "i_haz_configuration",
  jsonrpc = "2.0",
  method = "workspace/configuration",
  params = {
    items = { {
        section = "zls.enable_snippets"
      }, {
        section = "zls.enable_ast_check_diagnostics"
      }, {
        section = "zls.enable_autofix"
      }, {
        section = "zls.enable_import_embedfile_argument_completions"
      }, {
        section = "zls.semantic_tokens"
      }, {
        section = "zls.enable_inlay_hints"
      }, {
        section = "zls.inlay_hints_show_builtin"
```

So, why is the configuration different between the requests before and after?

travisstaloch:
[...]

Issue #1071, opened 2023-03-16
Broken cross-file references


SuperAuguste:
`references.zig` uses `DocumentStore.collectDependencies` which collects the imports a file specifies; this incorrect - we should be collecting dependents (files that import the current file), not dependencies (files that the current file imports).

To take this on you'd have to create a new `collectDependents` function that loads in all files in a workspace while checking for imports and returns all files that import the current file.

Jarred-Sumner:
Yes I think this is what I meant with this issue: https://github.com/zigtools/zls/issues/1070

After a recent change, an import 3-4 levels deep through usingnamespace stops working

https://github.com/oven-sh/bun/blob/6f59bc7d2e7aa64956bdd6f56769467853fd0242/src/main.zig#L45

techatrix:
To make it clear what this issue means it practice, here is a symbol references test that is currently failing:

https://github.com/zigtools/zls/blob/fedbf31eb7d7cb55888e656501767cbe423061af/tests/lsp_features/references.zig#L208-L216

The first file declares a public struct and the second file tries to import and use it. Asking for symbol references on the struct declaration in the first file will not find the reference in the second file.

Argmaster:
Hi, I am not sure if this belongs here, I couldn't find issue description that would fit my case exactly, but this one sounds a bit like it, so here is the deal:

In project, ZLS has no idea what `const main = @import("root")` means, gives no suggestions, can not resolve the "root" when I try to go to definition. As a result any symbol referenced through `main` variable become completely opaque.  It makes navigating files grouped in folders really troublesome :C

Found in project called [Cubyz](https://github.com/PixelGuys/Cubyz).

![Image](https://github.com/user-attachments/assets/dda14cd0-4fad-4057-9f3a-4eba5832b19f)

Is that duplicate for this issue?
Are there any workarounds for this?

RoadRoller01:
Wouldn't it be better to first search for the symbol in all files within a workspace, and then process only those files if they are really referencing the symbol.

RoadRoller01:
And I it could also be better if the symbol accorded in a file you immediately check if its really referencing it, Instead of collecting an array list then process it.

dismint:
To clarify this issue, is there any case where a `textDocument/references` works across files? I could not reproduce a scenario where it found references that weren't in the document the request was being made from. Seems like this is more of a missing feature than a bug in that case.

techatrix:
[...]

PR #2141, opened 2025-01-10
Send $/progress messages on build file parsing

sphaerophoria:
```
Initial build processing can be slow. It can be desirable to know when
the initial processing is finished to ensure that LSP queries are
processing the entire code base

Implement LSP protocol for this: $/progress.
Message flow is as follows

* Client publishes that it supports the window/workDoneProgress/create
  command on initialize
* Something triggers a build file invalidation
* Server requests that the client creates a work done progress report
* Server sends the initial work done progress
* Server sends the final work done progress

Note that this commit does not attempt to send intermediate reports.
AFAICT this would involve some fairly complex hooking of the build
runner that probably isn't worth it

Implementation details:
* Followed pattern of DiagnosticsCollection. Give DocumentStore a handle
  to the thread safe transport, and send progress messages directly from
  the build worker thread
* In some cases multiple builds can happen in parallel. We need to
  ensure we aren't double starting the same progress message. Fold them
  all builds into one
```
(EDIT: Updated for most recent commit message)

Example of progress report in vs code
Screenshot_20250109_213649

Apologies if you don't want drive by PRs like this, let me know if there's anything I can do to make this easier on you :)

### DIFF
--- a/src/DocumentStore.zig
+++ b/src/DocumentStore.zig
@@ -6,6 +6,7 @@ const URI = @import("uri.zig");
 const analysis = @import("analysis.zig");
 const offsets = @import("offsets.zig");
 const log = std.log.scoped(.store);
+const lsp = @import("lsp");
 const Ast = std.zig.Ast;
 const BuildAssociatedConfig = @import("BuildAssociatedConfig.zig");
 const BuildConfig = @import("build_runner/shared.zig").BuildConfig;
@@ -25,6 +26,9 @@ handles: std.StringArrayHashMapUnmanaged(*Handle) = .empty,
 build_files: std.StringArrayHashMapUnmanaged(*BuildFile) = .empty,
 cimports: std.AutoArrayHashMapUnmanaged(Hash, translate_c.Result) = .empty,
 diagnostics_collection: *DiagnosticsCollection,
+builds_in_progress: std.atomic.Value(i32) = .init(0),
+transport: ?lsp.AnyTransport = null,
+supports_work_done_progress: bool = false,
 
 pub const Uri = []const u8;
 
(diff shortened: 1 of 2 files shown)


sphaerophoria:
I've stolen some logs from clangd which somewhat validate the set of messages we're sending (more than the feature working in vs code). Note that we don't send the intermediate report, so the message/percentage stuff is not necessary. This is supported by notes in the spec
```
/**
	 * Optional, more detailed associated progress message. Contains
	 * complementary information to the `title`.
	 *
	 * Examples: "3/25 files", "project/src/module2", "node_modules/some_dep".
	 * If unset, the previous progress message (if any) is still valid.
	 */
	message?: string;

	/**
	 * Optional progress percentage to display (value 100 is considered 100%).
	 * If not provided infinite progress is assumed and clients are allowed
	 * to ignore the `percentage` value in subsequent report notifications.
	 *
	 * The value should be steadily rising. Clients are free to ignore values
	 * that are not following this rule. The value range is [0, 100].
	 */
	percentage?: [uinteger](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#uinteger);
```
```
V[05:00:57.713] >>> {"id":0,"jsonrpc":"2.0","method":"window/workDoneProgress/create","params":{"token":"backgroundIndexProgress"}}
V[05:00:57.834] >>> {"jsonrpc":"2.0","method":"$/progress","params":{"token":"backgroundIndexProgress","value":{"kind":"begin","percentage":0,"title":"indexing"}}}
V[05:00:57.834] >>> {"jsonrpc":"2.0","method":"$/progress","params":{"token":"backgroundIndexProgress","value":{"kind":"report","message":"0/1","percentage":0}}}
V[05:01:10.996] >>> {"jsonrpc":"2.0","method":"$/progress","params":{"token":"backgroundIndexProgress","value":{"kind":"report","message":"152/2665","percentage":5}}}
V[05:04:47.141] >>> {"jsonrpc":"2.0","method":"$/progress","params":{"token":"backgroundIndexProgress","value":{"kind":"report","message":"2664/2665","percentage":99}}}
V[05:04:48.587] >>> {"jsonrpc":"2.0","method":"$/progress","params":{"token":"backgroundIndexProgress","value":{"kind":"end"}}}
```

sphaerophoria:
I should also note here, this does not trigger for the build on save path. A similar set of actions could be performed here, but I didn't need it for what I'm working on. If you feel that its important to do it while we're here, I can implement that as well. I think we'd probably want a separate set of messages + in progress tracking var

travisstaloch:
This addresses and possibly closes #1201

Techatrix (changes requested):
Overall this change looks great, there is just one request that can simplify the code.

> * DocumentStore owns the build file invalidation, but Server owns the message serialization/sending.

Unless it is necessary to get the response from requests, It is perfectly fine to send messages without going through Server.zig. For example, diagnostics are send independent of the main server loop. This means that this PR can be simplified by moving the added logic into the documenstore with the following fields:

```diff
diagnostics_collection: *DiagnosticsCollection,
+ builds_in_progress: std.atomic.Value(i32) = .init(0),
+ transport: ?lsp.Transport = null,
+ supports_work_done_progress: bool = false,
```

The `transport` field should be null until [setTransport](https://github.com/zigtools/zls/blob/61c71478a7c2f1281cd62428de3f5b3adcf98056/src/Server.zig#L1904) has been called.

This should make the `WorkerNotifier` unnecessary. I would also suggest to rebase your branch because I recently simplified the message serialization logic in Server.zig which you can use as well. See c942ce73b2fa1e6b8f5f66a3bfe0b4179cb46d15

> Apologies if you don't want drive by PRs like this, let me know if there's anything I can do to make this easier on you :)

Well... It not a drive by PR if you previously watched some videos where someones talks about sending a PR to ZLS. :)

FYI: Not sure if you already encountered this but ZLS's symbol references has an [issue](https://github.com/zigtools/zls/issues/1071) with how it collects references across multiple files. This can cause missing references being reported in your code map.

sphaerophoria:
Ah brilliant, I'll make those changes. I was wondering if I was missing something.

> FYI: Not sure if you already encountered this but ZLS's symbol references has an https://github.com/zigtools/zls/issues/1071 with how it collects references across multiple files. This can cause missing references being reported in your code map.

Ah interesting. I've definitely noticed some odd behavior. It seems to work much better if I've opened every file before retrieving references. I don't know how I manage to find references to a module from main given that bug report. The code seems to line up with what was said, but its not matching behavior I'm seeing. Either way, thanks for the heads up :)

sphaerophoria:
Looking at the build failure, it looks to be an environment issue related to starting kcov, not an actual reduction in coverage caused by my PR. I am probably adding any useful info here, but officially noting that I'm not planning on making any changes :P

Techatrix:
> Looking at the build failure, it looks to be an environment issue related to starting kcov, not an actual reduction in coverage caused by my PR.

I issue should be resolved by ce11ba21b1fc76f1a82011d200344bd660e075cb.

Techatrix (approved):
looks great!

PR #2145, opened 2025-01-14
Pass `--zig-lib-dir` config option to the build runner

GalaxyShard:
Without this patch, the build runner fails to run if build scripts depend on a modified zig library directory even if the config file specifies a different library directory, for example if the project depends on patches to the Zig build system.

I tested this change on a project of mine with a blank config file and with a config file that sets zig_lib_path; it seems to work correctly in both scenarios.


Techatrix (changes requested):
This is missing a check here:
https://github.com/zigtools/zls/blob/61c71478a7c2f1281cd62428de3f5b3adcf98056/src/DocumentStore.zig#L814-L816

and an assertion here:
https://github.com/zigtools/zls/blob/61c71478a7c2f1281cd62428de3f5b3adcf98056/src/DocumentStore.zig#L1039-L1041

Techatrix (approved):
:+1:

PR #2157, opened 2025-01-29
nix: add meta.mainProgram to silence nix warning

joshryandavis:
Nix now warn users if a package does not have `meta.mainProgram` set.

This is because of https://github.com/NixOS/nixpkgs/pull/246386.

This PR simply adds `meta.mainProgram = "zls";` to the `flake.nix`.


Techatrix (approved):
LGTM!

PR #2159, opened 2025-01-29
Allow a package depending on zls to skip 'git describe'

pfgithub:
This lets someone build zls as a dependency of another package with `b.installArtifact(b.dependency("zls", .{ .target = target, .optimize = optimize, .@"commit-hash" = @as([]const u8, "HASH") }).artifact("zls"))` without requiring `git` or erroring if the git output isn't as expected


Techatrix:
I would rather have a build option to specify the entire version string (See #2183). It can be used to also specify the pre release version to check whether nightly versions are older or newer than each other.

PR #2164, opened 2025-02-02
nix: replace flake-utils with polyfill

FalsePattern:
Replacing the logic of `flake-utils.eachDefaultSystem` as a small polyfill reduces the amount of downstream pollution in peoples' flake.lock files.
See also:
https://github.com/mitchellh/zig-overlay/pull/60

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -34,24 +34,6 @@
         "type": "github"
       }
     },
-    "flake-utils_2": {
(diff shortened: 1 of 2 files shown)


Techatrix (approved):
Looks good. I will probably follow this up with some cleanup and minor adjustments to the flake.nix.

PR #2171, opened 2025-02-06
infer `orelse` on a C-pointer as a C-pointer

nolanderc:
A C-pointer `[*c]T` can be `null`, and it is possible to use with `orelse` and `.?` unwrapping.
However, as `[*c]` pointers carry some special semantics, the result is still a `[*c]` pointer.


nolanderc:
Seems the build is failing due to a new version, so this is blocked on #2168

Techatrix (approved):
TIL you can unwrap c pointers

PR #2178, opened 2025-02-09
Put param hint on previous line for multiline string arg

FnControlOption:
Before:

Screenshot 2025-02-08 at 9 53 58 PM

After:

Screenshot 2025-02-08 at 9 54 12 PM


Techatrix (approved):
That's a really nice change. This always annoyed me about inlay hints.

PR #2179, opened 2025-02-09
Update for breaking Zig Allocator API change

mochalins:
Updates ZLS to compile and pass tests with the https://github.com/ziglang/zig/pull/20511/ PR merged in Zig master. 

Of note is the new `remap` function in the binned allocator; it simply calls `resize` internally  à la `remap` functions introduced for several allocators in the Zig standard library (see e.g. `FixedBufferAllocator`), in the aforementioned PR.

This can also be replaced with a naive implementation that copies `resize`, but internally calling the underlying allocator's `remap`. I can make this change upon request, or this can be left for a followup PR (potentially leaving room for an entirely different implementation to make use of `remap`'s looser guarantees).

Also includes the minor `popOrNull` -> `pop` change in https://github.com/ziglang/zig/pull/22721

### DIFF
--- a/src/binned_allocator.zig
+++ b/src/binned_allocator.zig
@@ -108,10 +108,9 @@ pub fn BinnedAllocator(comptime config: Config) type {
         ) bool {
             const self: *Self = @ptrCast(@alignCast(ctx));
 
-            const align_ = alignment.toByteUnits();
             comptime var prev_size: usize = 0;
             inline for (&self.bins) |*bin| {
-                if (buf.len <= bin.size and align_ <= bin.size) {
+                if (buf.len <= bin.size and alignment.toByteUnits() <= bin.size) {
                     // Check it still fits
                     return new_len > prev_size and new_len <= bin.size;
                 }
@@ -126,9 +125,8 @@ pub fn BinnedAllocator(comptime config: Config) type {
         fn free(ctx: *anyopaque, buf: []u8, alignment: std.mem.Alignment, ret_addr: usize) void {
             const self: *Self = @ptrCast(@alignCast(ctx));
 
-            const align_ = alignment.toByteUnits();
             inline for (&self.bins) |*bin| {
-                if (buf.len <= bin.size and align_ <= bin.size) {
+                if (buf.len <= bin.size and alignment.toByteUnits() <= bin.size) {
                     bin.free(buf.ptr);
                     return;
                 }
--- a/src/binned_allocator.zig
+++ b/src/binned_allocator.zig
@@ -77,19 +77,19 @@ pub fn BinnedAllocator(comptime config: Config) type {
         fn alloc(
             ctx: *anyopaque,
             len: usize,
-            log2_align: std.mem.Alignment,
+            alignment: std.mem.Alignment,
             ret_addr: usize,
         ) ?[*]u8 {
             const self: *Self = @ptrCast(@alignCast(ctx));
 
-            const size = @max(len, log2_align.toByteUnits());
+            const size = @max(len, alignment.toByteUnits());
             inline for (&self.bins) |*bin| {
                 if (size <= bin.size) {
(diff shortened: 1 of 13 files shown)


src/binned_allocator.zig
```diff
@@ -94,10 +100,16 @@ pub fn BinnedAllocator(comptime config: Config) type {
...
+            ctx: *anyopaque,
+            buf: []u8,
+            log2_align: std.mem.Alignment,
+            new_len: usize,
+            ret_addr: usize,
+        ) bool {
             const self: *Self = @ptrCast(@alignCast(ctx));
 
-            const align_ = @as(usize, 1) << @as(std.math.Log2Int(usize), @intCast(log2_align));
+            const align_ = @as(usize, 1) << @intFromEnum(log2_align);
```
bullekeup, line 112: Seems like there's a function for that in std.mem.Alignment 

```zig 
const align_ = log2_align.toByteUnits();
```

src/binned_allocator.zig
```diff
@@ -69,14 +69,20 @@ pub fn BinnedAllocator(comptime config: Config) type {
...
+                    .remap = remap,
                 },
             };
         }
 
-        fn alloc(ctx: *anyopaque, len: usize, log2_align: u8, ret_addr: usize) ?[*]u8 {
+        fn alloc(
+            ctx: *anyopaque,
+            len: usize,
+            log2_align: std.mem.Alignment,
```
bullekeup, line 80: Maybe 'log2_align' can be renamed 'alignment' to match std allocators parameter name ? Same for other allocator functions ?

bullekeup:
Hi

I am new to zig and zls and was about to open a PR with the same changes (my first try to contribute in the zig ecosystem), but @mochalins has been quicker ! My changes just have the difference highlighted in my previous comments. 

I don't want to open a conflicting PR so just take my suggestions if you want to

mochalins:
Thank you for the suggestions @bullekeup ! I hadn't noticed that helper function for alignment. I'd love to incorporate your suggestion, but it'd be even better if you would show up in the commit history to get proper credit; would you mind opening a PR against my fork branch so that I can accept it and have your commit show up here?

mochalins:
Requires one more update to handle latest changes from master; will fix shortly.

mochalins:
@Techatrix 

The latest changes to `popOrNull` -> `pop` on Zig master affected the `master` build runner. As such, the `mach-latest` build runner checks on CI were failing; to address this, I created a separate `mach-latest.zig` file in `build_runner` that runs on the CI, and I did not add another `master` build runner check to the CI matrix due to the comment that master build runner is checked in tests.

I wasn't able to keep the build runner to one `master.zig` file, as comptime branching on the builtin Zig version did not properly compare the `pre` part of the version.

Are these changes acceptable?

Techatrix:
A new allocator ([std.heap.SmpAllocator](https://github.com/ziglang/zig/pull/22823)) has been recently add to the standard library. I would rather have ZLS use it instead of the binned allocator.

> I wasn't able to keep the build runner to one `master.zig` file, as comptime branching on the builtin Zig version did not properly compare the `pre` part of the version.

ZLS has been comparing zig versions at comptime for a while now without issues. I tested the following change which worked for me:

```diff
-        while (stack.popOrNull()) |step| {
+        while (switch (comptime builtin.zig_version.order(std.SemanticVersion.parse("0.14.0-dev.3181+914248237") catch unreachable)) {
+            .lt => stack.popOrNull(),
+            .eq, .gt => stack.pop(),
+        }) |step| {
```

This could be happening because the `minimum_runtime_zig_version`  has been updated which means that ZLS won't run the build runner on older zig versions. The `mach-latest.zig` file is only used in the CI. It is never used in ZLS itself making it useless.

To not delay these changes any further, I will apply them myself but I still appreciate all the effort you have put into this PR.

PR #2180, opened 2025-02-09
Detect declarations deprecated with a doc comment

Vexu:
I already get a compile error when using declarations deprecated with `@compileError`, seeing the ones only deprecated with doc comments is much more useful.

On a related note, why aren't the deprecated identifiers rendered with a strikethrough?

```zig
/// Deprecated
const foo = 123;
const bar = @compileError("foo");
comptime {
    foo;
    bar;
}
```
vs
```js
/**
 * @deprecated test
 */
const foo = 1;
foo
```
image

### DIFF
--- a/src/analysis.zig
+++ b/src/analysis.zig
@@ -1429,6 +1429,15 @@ const FindBreaks = struct {
     }
 };
 
+fn isDeprecatedWithDoc(tree: Ast, base_token: Ast.TokenIndex) bool {
+    const token_tags = tree.tokens.items(.tag);
+    const doc_index = getDocCommentTokenIndex(token_tags, base_token) orelse return false;
+    const doc_slice = tree.tokenSlice(doc_index);
+    if (doc_slice.len < "///Deprecated".len) return false;
+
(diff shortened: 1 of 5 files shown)


Techatrix (commented):
> On a related note, why aren't the deprecated identifiers rendered with a strikethrough?

I believe that VS Code (and possibly other editors) use diagnostics decide whether to render strikethrough symbols using [DiagnosticTag::Deprecated](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#diagnosticTag). To use semantic tokens, the following change was necessary:
```json
  "editor.semanticTokenColorCustomizations": {
    "rules": {
      "*.deprecated": {
        "strikethrough": true
      }
    }
  }
```
In the javascript example, you should see a `'foo' is deprecated.` message when hovering over `foo`.

---

Since https://github.com/ziglang/zig/issues/22822 has been accepted, I am not sure we should support signifying deprecation through comments and instead rely on `@deprecated`. The [DiagnosticTag::Deprecated](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#diagnosticTag) should be added to error messages if they contain the word "deprecated" at the start or match the output produced by `@deprecated`.

PR #2181, opened 2025-02-09
Fix type resolution of slice with integer literals

### LINKED ISSUE #1230: Incorrect type inference when slicing a slice with comptime-known range

FnControlOption:
Partially addresses #1230

### DIFF
--- a/src/analysis.zig
+++ b/src/analysis.zig
@@ -1016,6 +1016,14 @@ fn resolveBracketAccessType(analyser: *Analyser, lhs: Type, rhs: BracketAccess) 
             .for_range => return Type.fromIP(analyser, .usize_type, null),
             else => return null,
         },
+        .tuple => |fields| switch (rhs) {
+            .single => |index_maybe| {
+                const index = index_maybe orelse return null;
+                if (index >= fields.len) return null;
+                return fields[index];
+            },
+            .open, .range => return null,
+        },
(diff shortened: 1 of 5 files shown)


Techatrix:
When rebasing this PR on top of #2192, I would suggest to replace the added tests for hover and inlay hints with analysis tests in `tests/analysis`. There are a bunch of TODO tests in `pointer.zig` and `array.zig` that this PR should resolve.

Techatrix (approved):
:tada:

PR #2186, opened 2025-02-12
Move `Type.Data.ip_index` and `.either` to `Type`

FnControlOption:
The `is_type_val` field is redundant for `ip_index` and `either`, so let's turn `Type` into a tagged union with those and the original struct as the possible variants. Currently, I've named the new variant as `dynamic`, but I'm open to suggestions!

<table>
<tr>
<th>Before</th>
<th>After</th>
</tr>
<tr>
<td>

```zig
pub const Type = struct {
    data: Data,
    is_type_val: bool,

    pub const Data = union(enum) {
        // ...

        compile_error: NodeWithHandle,

        either: []const EitherEntry,

        ip_index: struct {
            node: ?NodeWithHandle = null,
            index: InternPool.Index,
        },

        pub const EitherEntry = struct {
            type_data: Data,
            descriptor: []const u8,
        };
    };

    // ...

    pub const TypeWithDescriptor = struct {
        type: Type,
        descriptor: []const u8,
    };

    // ...
};
```

</td>
<td>

```zig
pub const Type = union(enum) {
    dynamic: struct {
        data: Data,
        is_type_val: bool,
    },

    ip_index: struct {
        node: ?NodeWithHandle = null,
        index: InternPool.Index,
    },

    either: []const TypeWithDescriptor,

    pub const Data = union(enum) {
        // ...

        compile_error: NodeWithHandle,
    };

    // ...

    pub const TypeWithDescriptor = struct {
        type: Type,
        descriptor: []const u8,
    };

    // ...
};
```

</td>
</tr>
</table>

### DIFF
--- a/src/analysis.zig
+++ b/src/analysis.zig
@@ -326,29 +326,29 @@ pub fn isInstanceCall(
     call: Ast.full.Call,
     func_ty: Type,
 ) error{OutOfMemory}!bool {
-    std.debug.assert(!func_ty.is_type_val);
+    std.debug.assert(!func_ty.isTypeVal(analyser));
     if (call_handle.tree.nodes.items(.tag)[call.ast.fn_expr] != .field_access) return false;
 
     const container_node: NodeWithHandle = .{ .node = call_handle.tree.nodes.items(.data)[call.ast.fn_expr].lhs, .handle = call_handle };
 
     const container_ty = if (try analyser.resolveTypeOfNodeInternal(container_node)) |container_instance|
-        container_instance.typeOf(analyser)
+        try container_instance.typeOf(analyser)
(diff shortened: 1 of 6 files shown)


Techatrix:
What exactly is this PR trying to improve? The PR description implies code quality by removing a redundant field but the code now requires analysis to switch twice every time it want to extract some information out of `Type`. Having `is_type_val` is not a big issue for `ip_index` and `either` since it avoids computing it (which happens quite often) and the rest of the code remains more readable. Hypothetically, if ZLS was already in the "After" state, I would accept a PR that would change it to the "Before" state.

This change will also conflicts with #2180 since it tries to add a new field to `Type`.

PR #2199, opened 2025-02-22
Resolve type of floating point builtin functions

FnControlOption:
e.g. `sqrt`, `sin`, etc.

### DIFF
--- a/src/analyser/InternPool.zig
+++ b/src/analyser/InternPool.zig
@@ -3250,6 +3250,30 @@ fn isUnknownDeepInternal(
     };
 }
 
+/// Returns true if and only if the type is a fixed-width, signed integer.
+pub fn isSignedInt(ip: *InternPool, ty: Index, target: std.Target) bool {
+    return switch (ty) {
+        .c_char_type => target.cCharSignedness() == .signed,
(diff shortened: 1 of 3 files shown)


Techatrix (changes requested):
[@abs](https://ziglang.org/documentation/master/#abs) is special:
> Returns the absolute value of an integer or a floating point number. Uses a dedicated hardware instruction when available. The return type is always an unsigned integer of the same bit width as the operand if the operand is an integer. Unsigned integer operands are supported. The builtin cannot overflow for signed integer operands.
>
> Supports [Floats](https://ziglang.org/documentation/master/#Floats), [Integers](https://ziglang.org/documentation/master/#Integers) and [Vectors](https://ziglang.org/documentation/master/#Vectors) of floats or integers.

The tests should also check that if an invalid operand is provided, the result is unresolved. Example:
```zig
const foo = @sqrt(null);
//    ^^^ (unknown)()
```

Techatrix:
@FnControlOption did you perhaps overlook this PR?

FnControlOption:
waiting on ziglang/zig#23587

FnControlOption:
Copied functions from Sema and Type

Will update `InternPool.toUnsigned` in a later PR if ziglang/zig#23587 is merged

Techatrix (approved):

PR #2201, opened 2025-02-23
Represent error sets through InternPool

FnControlOption:
Preparatory work for peer type resolution with error unions

Regression 1: link to error set is no longer included in "Go to ..." when hovering.

Regression 2: given error set `E = error{A,B}` and struct `S`, `E!S` is printed as `error{A,B}!S`. However, this is consistent with how Zig would normally print `E!S` in e.g. `std.debug.print`.

```zig
const std = @import("std");
const E = error{A,B};
const S = struct{};
pub fn main() !void {
    std.debug.print("{}", .{E!S}); // => error{A,B}!S
}
```

### DIFF
--- a/src/DocumentScope.zig
+++ b/src/DocumentScope.zig
@@ -115,7 +115,6 @@ fn getDeclNameToken(tree: Ast, node: Ast.Node.Index) ?Ast.TokenIndex {
         .tagged_union_two_trailing,
         .tagged_union_enum_tag,
         .tagged_union_enum_tag_trailing,
-        .error_set_decl,
         .block,
         .block_semicolon,
         .block_two,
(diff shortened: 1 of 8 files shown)


Techatrix (changes requested):
My apologies for the late review on this. I was trying to wait for https://github.com/zigtools/zls/pull/2199 before reviewing this.

The change looks good to me except for the extra 6775ca4814f9900c80d06773314efd39cf6142f5 commit. A decl in the InternPool references an ast node which will go out of sync when the document is updated. I haven't yet figured out how to handle this so I would like to avoid the potential issues that could come with such a change.

If you're okay with it, I'm happy to take over this PR and handle the revert and rebase myself.

Techatrix (approved):

PR #2213, opened 2025-03-05
Upgrade Zig version to 0.14.0 in build.zig.zon and build.zig

terraquad:
This should allow ZLS to build on 0.14.0 again


terraquad:
It seems that these line causes build errors:

https://github.com/zigtools/zls/blob/b9c9b5abe3e175821055e9ae1467c64a48eb3022/build.zig#L623-L629

(Specifically line 626)

I think we should remove this restriction because Zig 0.14.0 can in fact build ZLS and master can be built by tagged releases.

PR #2226, opened 2025-03-10
add config option insert_completion

BiggerBen13:
Adds the option to use insertText completionitems instead of textEdit completionitems implying that filtering needs to be done on the client. This is probably a niche fix as most editors filter regardless, of which which completion kind is received. NeoVim declared this behavior as [intentional](https://github.com/neovim/neovim/issues/31911#issuecomment-2577906280) so this pr represents a workaround, until zls filters auto completions on its own

### DIFF
--- a/src/features/completions.zig
+++ b/src/features/completions.zig
@@ -938,13 +938,28 @@ pub fn completionAtIndex(
     const replace_range = offsets.locToRange(source, .{ .start = start_index, .end = end_index }, server.offset_encoding);
 
     for (completions) |*item| {
-        if (item.textEdit == null) {
-            item.textEdit = if (server.client_capabilities.supports_completion_insert_replace_support)
-                .{ .InsertReplaceEdit = .{ .newText = item.insertText orelse item.label, .insert = insert_range, .replace = replace_range } }
-            else
-                .{ .TextEdit = .{ .newText = item.insertText orelse item.label, .range = insert_range } };
+        switch (builder.server.config.insert_completions) {
+            false => {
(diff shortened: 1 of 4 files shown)


sweiglbosker:
Looks good.

src/features/completions.zig
```diff
@@ -938,13 +938,28 @@ pub fn completionAtIndex(
     const replace_range = offsets.locToRange(source, .{ .start = start_index, .end = end_index }, server.offset_encoding);
 
     for (completions) |*item| {
-        if (item.textEdit == null) {
-            item.textEdit = if (server.client_capabilities.supports_completion_insert_replace_support)
-                .{ .InsertReplaceEdit = .{ .newText = item.insertText orelse item.label, .insert = insert_range, .replace = replace_range } }
-            else
-                .{ .TextEdit = .{ .newText = item.insertText orelse item.label, .range = insert_range } };
+        switch (builder.server.config.insert_completions) {
```
GreyElaina, line 941: uh, why switch

Techatrix:
This workaround should not be necessary with 5449f65d2e042addb521c0d8deff29720a461dee. Let me know if it resolves your issue. If not, please provide a reproducible example that showcases the observed issue.

sweiglbosker:
yeah that solution seems better to me, should work.

Techatrix:
No response from OP so I am going to assume that this issue has been resolved by 5449f65d2e042addb521c0d8deff29720a461dee. If the issue is still present, please open a new issue with a reproducible example.

PR #2229, opened 2025-03-13
Use new type-safe AST

mochalins:


mochalins:
Currently compiles and passes 322/340 tests.

Unsure of how to proceed with the debug tree printing in `src/debug.zig`; temporarily removed it to run tests.

Techatrix:
FYI I already have a local branch with the necessary changes since I'm the one that made https://github.com/ziglang/zig/pull/22397.

PR #2236, opened 2025-04-02
dont give label details for typeless valueless enum and union fields

xdBronch:
maybe this is a bit opinionated but i find it quite noisy depending on how the editor presents details e.g. zed which has no separation 
before:
image
after:
image
the name is still duplicated on `a` unfortunately, i considered just leaving ` = 69` but its a bit awkward imo, i think its alright here. open to other formatting suggestions though

### DIFF
--- a/src/features/completions.zig
+++ b/src/features/completions.zig
@@ -1361,10 +1361,14 @@ fn collectContainerFields(
                     break :insert_text try std.fmt.allocPrint(builder.arena, "{s} = ", .{name});
                 };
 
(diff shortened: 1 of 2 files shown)


Techatrix (approved):
Having the detail field be just ` = 69` may work nicely with an editor that displays the detail to the right of the label without a separator like Zed but can be less optimal for other editors. Such a change should take other editors into account as well.

The detail field being the same as the label is just useless so that is a nice improvement.

PR #2238, opened 2025-04-03
feat: support `workspace/didChangeWatchedFiles` for imported files

### LINKED ISSUE #2160: watch for filesystem events on imported files
### Zig Version

0.14.0-dev.365+6a16b27

### ZLS Version

0.14.0-dev.2802+257054a14

VSCodium 1.94.2 with vscode-zig v0.6.4

1. create the following two files on disk
2. open `root.zig` in the editor (don't open `mod.zig`)
4. request for completions in `root.zig` which will show `Hello` 👍 
3. edit `mod.zig` in a different editor instance, change `Hello` to `World`.
4. request for completions in `root.zig` which will still show `Hello` 👎 

```zig
/// root.zig
const mod = @import("mod.zig");
test {
    mod.<request code completions here>
}
```

```zig
/// mod.zig
pub const Hello = struct {};
```

### Expected Behavior

Show `World` as the completion item.

ZLS should use [workspace/didChangeWatchedFiles](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_didChangeWatchedFiles) to be notified about changed files.

### Relevant log output

```shell

```

WillLillis:
This adds support for `workspace/didChangeConfiguration`. I have a few questions/ points of interest:

- I ran into some minor issues with with the [`FileSystemWatcher`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#fileSystemWatcher)'s [`GlobPattern`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#pattern). 
   - Trying out just `"**/*.{zig,zon}"`  (as well as `"**/*.zig"`) caused the server to receive no notifications. I'm not sure if that was an issue with my implementation and/or Neovim's client implementation.
  - I opted to go with the basic [`Pattern`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#pattern) glob because the [`RelativePattern`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#relativePattern) glob is behind the capability's [registration params](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#didChangeWatchedFilesRegistrationOptions), so there may be some LSP clients that don't implement this feature.
  - Because registering for `"**"` includes *all* files in the project directory (including those in `.zig-cache`), I tried filtering out extraneous notifications by only applying updates to uris already in the document store. I *believe* this should cover all imported files, but I'm not 100% confident.

- Reading the entire file seems pretty expensive for every update, but it looks like the notification doesn't include any edit information. Maybe there's a smarter way around this?

- I wasn't sure if it was more better to use the document store's `refreshDocument` method in the `.Changed` case. The logic between `.Created` and `.Changed` would still be almost identical, except that `refreshDocument` takes a null terminated slice, and requires allocation via the document store. 

- `openDocument` is marked as not thread safe. If this is to be used in the `.Changed` case, should some additional checks/ locking be added at the call site?

- I wasn't sure how to add tests for this. 

Closes #2160

### DIFF
--- a/src/Server.zig
+++ b/src/Server.zig
@@ -17,6 +17,7 @@ const Analyser = @import("analysis.zig");
 const offsets = @import("offsets.zig");
 const tracy = @import("tracy");
 const diff = @import("diff.zig");
+const Uri = @import("uri.zig");
 const InternPool = @import("analyser/analyser.zig").InternPool;
 const DiagnosticsCollection = @import("DiagnosticsCollection.zig");
 const known_folders = @import("known-folders");
@@ -92,6 +93,7 @@ const ClientCapabilities = struct {
     label_details_support: bool = false,
     supports_configuration: bool = false,
     supports_workspace_did_change_configuration_dynamic_registration: bool = false,
+    supports_workspace_did_change_watched_files: bool = false,
     supports_textDocument_definition_linkSupport: bool = false,
     /// The detail entries for big structs such as std.zig.CrossTarget were
     /// bricking the preview window in Sublime Text.
@@ -487,6 +489,11 @@ fn initializeHandler(server: *Server, arena: std.mem.Allocator, request: types.I
                 server.client_capabilities.supports_workspace_did_change_configuration_dynamic_registration = true;
             }
         }
+        if (workspace.didChangeWatchedFiles) |did_change| {
+            if (did_change.dynamicRegistration orelse false) {
+                server.client_capabilities.supports_workspace_did_change_watched_files = true;
+            }
+        }
         if (workspace.semanticTokens) |workspace_semantic_tokens| {
             server.document_store.lsp_capabilities.supports_semantic_tokens_refresh = workspace_semantic_tokens.refreshSupport orelse false;
         }
@@ -610,7 +617,7 @@ fn initializeHandler(server: *Server, arena: std.mem.Allocator, request: types.I
     };
 }
 
-fn initializedHandler(server: *Server, _: std.mem.Allocator, notification: types.InitializedParams) Error!void {
+fn initializedHandler(server: *Server, arena: std.mem.Allocator, notification: types.InitializedParams) Error!void {
     _ = notification;
 
     if (server.status != .initializing) {
@@ -620,7 +627,20 @@ fn initializedHandler(server: *Server, _: std.mem.Allocator, notification: types
     server.status = .initialized;
 
     if (server.client_capabilities.supports_workspace_did_change_configuration_dynamic_registration) {
-        try server.registerCapability("workspace/didChangeConfiguration");
+        try server.registerCapability("workspace/didChangeConfiguration", null);
+    }
+
+    if (server.client_capabilities.supports_workspace_did_change_watched_files) {
+        // `{ "watchers": [ { "globPattern": "**/*.{zig,zon}" } ] }`
+        var watcher = std.json.ObjectMap.init(arena);
+        try watcher.put("globPattern", .{ .string = "**/*.{zig,zon}" });
+        var watchers_arr = try std.ArrayList(std.json.Value).initCapacity(arena, 1);
+        watchers_arr.appendAssumeCapacity(.{ .object = watcher });
+        var fs_watcher_obj: std.json.ObjectMap = std.json.ObjectMap.init(arena);
+        try fs_watcher_obj.put("watchers", .{ .array = watchers_arr });
+        const json_val: ?std.json.Value = .{ .object = fs_watcher_obj };
+
+        try server.registerCapability("workspace/didChangeWatchedFiles", json_val);
     }
 
     if (server.client_capabilities.supports_configuration) {
@@ -646,7 +666,7 @@ fn exitHandler(server: *Server, _: std.mem.Allocator, _: void) Error!void {
     };
 }
 
-fn registerCapability(server: *Server, method: []const u8) Error!void {
+fn registerCapability(server: *Server, method: []const u8, registersOptions: ?types.LSPAny) Error!void {
     const id = try std.fmt.allocPrint(server.allocator, "register-{s}", .{method});
     defer server.allocator.free(id);
 
@@ -659,6 +679,7 @@ fn registerCapability(server: *Server, method: []const u8) Error!void {
             .{
                 .id = id,
                 .method = method,
+                .registerOptions = registersOptions,
             },
         } },
     );
@@ -848,6 +869,23 @@ fn removeWorkspace(server: *Server, uri: types.URI) void {
     } else {
         log.warn("could not remove Workspace Folder: {s}", .{uri});
     }
+}
+
+fn didChangeWatchedFilesHandler(server: *Server, arena: std.mem.Allocator, notification: types.DidChangeWatchedFilesParams) Error!void {
+    _ = arena;
+
+    var updated_files: usize = 0;
+    for (notification.changes) |change| {
+        if (!(std.mem.endsWith(u8, change.uri, ".zig") or std.mem.endsWith(u8, change.uri, ".zon"))) continue;
+        switch (change.type) {
+            .Changed, .Deleted => {
+                try server.document_store.refreshDocumentFromFileSystem(change.uri);
+                updated_files += 1;
+            },
+            else => {},
+        }
+    }
+    log.info("updated {d} watched file(s)", .{updated_files});
 }
 
 fn didChangeWorkspaceFoldersHandler(server: *Server, arena: std.mem.Allocator, notification: types.DidChangeWorkspaceFoldersParams) Error!void {
@@ -1828,6 +1866,7 @@ const HandledNotificationParams = union(enum) {
     @"textDocument/didChange": types.DidChangeTextDocumentParams,
     @"textDocument/didSave": types.DidSaveTextDocumentParams,
     @"textDocument/didClose": types.DidCloseTextDocumentParams,
+    @"workspace/didChangeWatchedFiles": types.DidChangeWatchedFilesParams,
     @"workspace/didChangeWorkspaceFolders": types.DidChangeWorkspaceFoldersParams,
     @"workspace/didChangeConfiguration": types.DidChangeConfigurationParams,
     other: lsp.MethodWithParams,
@@ -1873,6 +1912,7 @@ fn isBlockingMessage(msg: Message) bool {
             .@"textDocument/didChange",
             .@"textDocument/didSave",
             .@"textDocument/didClose",
+            .@"workspace/didChangeWatchedFiles",
             .@"workspace/didChangeWorkspaceFolders",
             .@"workspace/didChangeConfiguration",
             => return true,
@@ -2054,6 +2094,7 @@ pub fn sendNotificationSync(server: *Server, arena: std.mem.Allocator, comptime 
         .@"textDocument/didChange" => try server.changeDocumentHandler(arena, params),
         .@"textDocument/didSave" => try server.saveDocumentHandler(arena, params),
         .@"textDocument/didClose" => try server.closeDocumentHandler(arena, params),
+        .@"workspace/didChangeWatchedFiles" => try server.didChangeWatchedFilesHandler(arena, params),
         .@"workspace/didChangeWorkspaceFolders" => try server.didChangeWorkspaceFoldersHandler(arena, params),
         .@"workspace/didChangeConfiguration" => try server.didChangeConfigurationHandler(arena, params),
         .other => {},
--- a/src/Server.zig
+++ b/src/Server.zig
@@ -879,13 +879,15 @@ fn didChangeWatchedFilesHandler(server: *Server, arena: std.mem.Allocator, notif
         if (!(std.mem.endsWith(u8, change.uri, ".zig") or std.mem.endsWith(u8, change.uri, ".zon"))) continue;
         switch (change.type) {
             .Created, .Changed, .Deleted => {
-                try server.document_store.refreshDocumentFromFileSystem(change.uri);
-                updated_files += 1;
+                const did_update_file = try server.document_store.refreshDocumentFromFileSystem(change.uri);
+                updated_files += @intFromBool(did_update_file);
             },
             else => {},
         }
     }
-    log.info("updated {d} watched file(s)", .{updated_files});
+    if (updated_files != 0) {
+        log.info("updated {d} watched file(s)", .{updated_files});
+    }
 }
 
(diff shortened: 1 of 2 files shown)


Techatrix:
About the file system watch pattern being used: Not sure why neovim wouldn't be able to handle `**/*.{zig,zon}`. Perhaps it would make sense to look into neovim's implementation to check if is a bug. We should also look at other editors to check whether it works and if they have the same limitation as neovim. I will try to look into that after the rest of my comment has been addressed.

About the usage of `refreshDocument` and `openDocument`: It is important to note that the LSP protocol has a [document synchronization](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_synchronization) mechanism that is independent of the actual file system. The `openDocument`, `closeDocument` and `refreshDocument` functions in ZLS are meant to implement this. Even if the document on the file system has been modified as notified by `workspace/didChangeConfiguration`, this event must be ignored if the document is synchronized over the protocol. I see now that the documentation in `DocumentStore.zig` is insufficiently explaining these requirements so I will look into improve them.

To resolve this issue, I would suggest to add the following (untested) function to the DocumentStore:
```zig
/// **Thread safe** takes an exclusive lock when called on different documents
/// **Not thread safe** when called on the same document
pub fn refreshDocumentFromFileSystem(self: *DocumentStore, uri: Uri) !void {
    {
        const handle = self.getHandle(uri) orelse return;
        if (handle.getStatus().open) return;
    }

    {
        self.lock.lock();
        defer self.lock.unlock();

        const kv = self.handles.fetchSwapRemove(uri) orelse return;
        log.debug("Closing document {s}", .{kv.key});
        kv.value.deinit();
        self.allocator.destroy(kv.value);
    }
}
```

This function will essentially remove the document from the document store unless it synchronized over the protocol. If the file system watching reports that the document has been modified or deleted, this function can be called. The existing logic in ZLS will read the new state from disk if it is needed again. This is why there is need to do anything when a file has been created.

Also after parsing the uri from the client, the file extension should be checked.

> * I wasn't sure how to add tests for this.

The document store lacks testing for now so there is no easy way to add tests for this. Should be fine for now.

WillLillis:
> About the usage of `refreshDocument` and `openDocument`: It is important to note that the LSP protocol has a [document synchronization](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_synchronization) mechanism that is independent of the actual file system. The `openDocument`, `closeDocument` and `refreshDocument` functions in ZLS are meant to implement this. Even if the document on the file system has been modified as notified by `workspace/didChangeConfiguration`, this event must be ignored if the document is synchronized over the protocol. I see now that the documentation in `DocumentStore.zig` is insufficiently explaining these requirements so I will look into improve them.
> To resolve this issue, I would suggest to add the following (untested) function to the DocumentStore:
> 
> ```zig
> /// **Thread safe** takes an exclusive lock when called on different documents
> /// **Not thread safe** when called on the same document
> pub fn refreshDocumentFromFileSystem(self: *DocumentStore, uri: Uri) !void {
>     {
>         const handle = self.getHandle(uri) orelse return;
>         if (handle.getStatus().open) return;
>     }
> 
>     {
>         self.lock.lock();
>         defer self.lock.unlock();
> 
>         const kv = self.handles.fetchSwapRemove(uri) orelse return;
>         log.debug("Closing document {s}", .{kv.key});
>         kv.value.deinit();
>         self.allocator.destroy(kv.value);
>     }
> }
> ```
> 
> This function will essentially remove the document from the document store unless it synchronized over the protocol. If the file system watching reports that the document has been modified or deleted, this function can be called. The existing logic in ZLS will read the new state from disk if it is needed again. This is why there is need to do anything when a file has been created.

Ah ok. It sounds obvious now, I just didn't realize those functions were so intended to go hand in hand with `textDocument/didOpen` etc. Thanks for the explanation! I tried out `refreshDocumentFromFileSystem` locally and it's working great. 

> Also after parsing the uri from the client, the file extension should be checked.

Gotcha. I'm assuming just checking for `.zig` or `.zon` should be ok?

> About the file system watch pattern being used: Not sure why neovim wouldn't be able to handle `**/*.{zig,zon}`. Perhaps it would make sense to look into neovim's implementation to check if is a bug. We should also look at other editors to check whether it works and if they have the same limitation as neovim. I will try to look into that after the rest of my comment has been addressed.

Sounds good. I'll push up the other changes for now, and spend some time debugging Neovim's implementation.

WillLillis:
Update: I haven't been able to pin down what's wrong in Neovim, but the `**/*.{zig,zon}` pattern *does* work in Zed.

Techatrix:
> Since this appears to be a neovim issue, maybe a workaround like this makes sense for now?

I think that's reasonable. But instead of checking checking the client name right where the glob pattern is created, I would suggest some changes to be more in line with how ZLS handles other editor specific workarounds:

- add a new field to [ClientCapabilities](https://github.com/zigtools/zls/blob/9149e9a6ffdfe68161af7ad4fb4b2188bf002483/src/Server.zig#L77) that represents this workaround. The name of the field is up to you, just make sure to add a link to the neovim issue so it can be tracked more easily. As a reference, the [`max_detail_len`](https://github.com/zigtools/zls/blob/9149e9a6ffdfe68161af7ad4fb4b2188bf002483/src/Server.zig#L95-L98) field is a workaround for Sublime Text.
- add a check for neovim that enables this new field [here](https://github.com/zigtools/zls/blob/9149e9a6ffdfe68161af7ad4fb4b2188bf002483/src/Server.zig#L362-L379)
- adjust the diff to check for the new field instead of the client name

src/Server.zig
```diff
@@ -846,6 +867,21 @@ fn removeWorkspace(server: *Server, uri: types.URI) void {
...
 
+fn didChangeWatchedFilesHandler(server: *Server, arena: std.mem.Allocator, notification: types.DidChangeWatchedFilesParams) Error!void {
+    _ = arena;
+
+    for (notification.changes) |change| {
+        if (!(std.mem.endsWith(u8, change.uri, ".zig") or std.mem.endsWith(u8, change.uri, ".zon"))) continue;
+        switch (change.type) {
+            .Changed, .Deleted => {
+                try server.document_store.refreshDocumentFromFileSystem(change.uri);
+                log.info("updated watched file: {s}", .{change.uri});
```
Techatrix, line 878: Doesn't this log message get very noisy when using git to switch between two commits with many changed files? If that's the case, would it be more reasonable to instead show a log message that says something like `updated {d} watched file(s)`?

WillLillis: Hadn't thought of that, seems reasonable to me.

Techatrix:
I checked a few editors and here is what I found:
- VS Code - wasn't working in some cases which should be fixed by 8468899b803a4ebc2f0e13e887e5b1aa13abf08e
- Sublime Text - client capabilities say that it is supported by no notifications are send
- JetBrains IDEs - ok
- Helix - unsupported
- Zed - ok
- NeoVim - Isn't enabled by default on Linux but was working when manually enabled
- Emacs - too lazy to setup emacs again
- Kate - unsupported

I also tested the `**/*.{zig,zon}` pattern in neovim which was surprisingly working for me. Perhaps the issue you are experiencing is tied to the file watch backend that is used by neovim. Could you check what `:checkhealth vim.lsp` outputs for you about file watching?

Here is what I get after manually configuring neovim to enable `workspace.didChangeWatchedFiles.dynamicRegistration`:
```
vim.lsp: File Watcher ~
- File watch backend: libuv-watchdirs
- WARNING libuv-watchdirs has known performance issues. Consider installing inotify-tools.
```

After installing `inotify-tools`, I got the following output while the `**/*.{zig,zon}` pattern was still working:
```
vim.lsp: File Watcher ~
- File watch backend: Custom (@/nix/store/sm3ycnp9hcqvq1bpnq45164wvbdk4f4z-neovim-unwrapped-0.11.0/share/nvim/runtime/lua/vim/_watch.lua)
```

WillLillis:
> I checked a few editors and here is what I found:
> 
>     * VS Code - wasn't working in some cases which should be fixed by [8468899](https://github.com/zigtools/zls/commit/8468899b803a4ebc2f0e13e887e5b1aa13abf08e)
> 
>     * Sublime Text - client capabilities say that it is supported by no notifications are send
> 
>     * JetBrains IDEs - ok
> 
>     * Helix - unsupported
> 
>     * Zed - ok
> 
>     * NeoVim - Isn't enabled by default on Linux but was working when manually enabled
> 
>     * Emacs - too lazy to setup emacs again
> 
>     * Kate - unsupported

Awesome, thanks so much for checking all of these!

> I also tested the `**/*.{zig,zon}` pattern in neovim which was surprisingly working for me. Perhaps the issue you are experiencing is tied to the file watch backend that is used by neovim. Could you check what `:checkhealth vim.lsp` outputs for you about file watching?
> 
> Here is what I get after manually configuring neovim to enable `workspace.didChangeWatchedFiles.dynamicRegistration`:
> 
> ```
> vim.lsp: File Watcher ~
> - File watch backend: libuv-watchdirs
> - WARNING libuv-watchdirs has known performance issues. Consider installing inotify-tools.
> ```
> 
> After installing `inotify-tools`, I got the following output while the `**/*.{zig,zon}` pattern was still working:
> 
> ```
> vim.lsp: File Watcher ~
> - File watch backend: Custom (@/nix/store/sm3ycnp9hcqvq1bpnq45164wvbdk4f4z-neovim-unwrapped-0.11.0/share/nvim/runtime/lua/vim/_watch.lua)
> ```

I got the same warning about `libuv-watchdirs` after running checkhealth. After installing `inotify-tools`, the `**/*.{zig,zon}` pattern worked for me! I uninstalled `inotify-tools` after just to make sure and the pattern still doesn't work for with the `libuv-watchdirs` backend. What commit is your neovim install built from? I bisected the regresion to https://github.com/neovim/neovim/commit/b87505e11656163ddf0dbbc16b7d224815873964.

I also noticed the `**` pattern causes some spurious errors to be printed about watching files in `.zig-cache` when the `inotify-tools` backend is in use. Given this, Do you think it makes sense to just use the `**/*.{zig,zon}` pattern?

Techatrix:
> What commit is your neovim install built from?

I used the 0.11.0 release. 

> I uninstalled `inotify-tools` after just to make sure and the pattern still doesn't work for with the `libuv-watchdirs` backend.

This is the thing that is confusing me. Both backends are working for me. If am I understanding this correctly that you are saying that the issues occurs since https://github.com/neovim/neovim/commit/b87505e11656163ddf0dbbc16b7d224815873964? So how can I not reproduce this issue with 0.11?

Also, did you need to manually configure neovim to enable the `workspace.didChangeWatchedFiles.dynamicRegistration` client capability like this:

<details>

<summary>init.lua</summary>

```lua
vim.lsp.config['zls'] = {
    cmd = { '/home/techatrix/repos/zls/zig-out/bin/zls' },
    filetypes = { 'zig' },
    root_markers = { 'build.zig' },
    settings = {
        zls = {
            zig_exe_path = '/home/techatrix/.config/Code/User/globalStorage/ziglang.vscode-zig/zig/linux-x86_64-0.15.0-dev.151+6e8493daa/zig'
        }
    },
    capabilities = {
        workspace = {
            didChangeWatchedFiles = {
                dynamicRegistration = true,
            }
        }
    },
}
vim.lsp.enable('zls')
```

</details>

---

I also tested windows where neovim was behaving as expected. There is just a minor issue with ZLS around missing uri escape sequence normalization which I will resolve later.

WillLillis:
> > I uninstalled `inotify-tools` after just to make sure and the pattern still doesn't work for with the `libuv-watchdirs` backend.
> 
> 
> 
> This is the thing that is confusing me. Both backends are working for me. If am I understanding this correctly that you are saying that the issues occurs since https://github.com/neovim/neovim/commit/b87505e11656163ddf0dbbc16b7d224815873964? So how can I not reproduce this issue with 0.11?

I'm honestly not sure, but I'll spend some more time looking into it. 

> Also, did you need to manually configure neovim to enable the `workspace.didChangeWatchedFiles.dynamicRegistration` client capability like this:
> 
> 
> 
> <details>
> 
> 
> 
> <summary>init.lua</summary>
> 
> 
> 
> ```lua
> 
> vim.lsp.config['zls'] = {
> 
>     cmd = { '/home/techatrix/repos/zls/zig-out/bin/zls' },
> 
>     filetypes = { 'zig' },
> 
>     root_markers = { 'build.zig' },
> 
>     settings = {
> 
>         zls = {
> 
>             zig_exe_path = '/home/techatrix/.config/Code/User/globalStorage/ziglang.vscode-zig/zig/linux-x86_64-0.15.0-dev.151+6e8493daa/zig'
> 
>         }
> 
>     },
> 
>     capabilities = {
> 
>         workspace = {
> 
>             didChangeWatchedFiles = {
> 
>                 dynamicRegistration = true,
> 
>             }
> 
>         }
> 
>     },
> 
> }
> 
> vim.lsp.enable('zls')
> 
> ```
> 
> 
> 
> </details>
> 
> 
> 
> ---


Yep!

> I also tested windows where neovim was behaving as expected. There is just a minor issue with ZLS around missing uri escape sequence normalization which I will resolve later.

👍

Techatrix:
I believe I found the cause of the inconsistency between us. When I do the same steps as https://github.com/zigtools/zls/issues/2160 but move `mod.zig` into its own subdirectory like `src/mod.zig`, I get the same issue. Based on https://github.com/neovim/neovim/issues/29865#issuecomment-2782056471 I assume that this is how you are testing it as well, right?

So if the issue only occurs on linux where file watching is disabled by default and the `libuv-watchdirs` backend is warned against anyway, I think we can just forget about the workaround and always use the `**/*.{zig,zon}` pattern.

WillLillis:
> I believe I found the cause of the inconsistency between us. When I do the same steps as #2160 but move `mod.zig` into its own subdirectory like `src/mod.zig`, I get the same issue. Based on [neovim/neovim#29865 (comment)](https://github.com/neovim/neovim/issues/29865#issuecomment-2782056471) I assume that this is how you are testing it as well, right?

Yeah, sorry I should have mentioned that. I was testing this in a template project created by `zig init`.

> So if the issue only occurs on linux where file watching is disabled by default and the `libuv-watchdirs` backend is warned against anyway, I think we can just forget about the workaround and always use the `**/*.{zig,zon}` pattern.

:+1:

Techatrix (approved):

WillLillis:
Thank you so much for all the help! @Techatrix

PR #2241, opened 2025-04-05
add "declared here" relatedInfo for autofix diagnostics

xdBronch:
image
i wasnt sure where to add tests for diagnostics like this?


Techatrix (approved):
> i wasnt sure where to add tests for diagnostics like this?

There are currently no tests for diagnostics. I have implemented them locally and will push them separately.

Issue #2106, opened 2024-12-10
provide more custom semantic token types/modfiers


Techatrix:
LSP [semantic tokens](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_semanticTokens) offer a predefined set of semantic token types and semantic token modifiers. ZLS is currently emitting mostly standard token types with relatively [few exceptions](https://github.com/zigtools/zls/blob/6c6923d7b1c4984554dade667f664d21f42b2b6c/src/features/semantic_tokens.zig#L36-L43). The problem is that the existing token types are not able to accurately represent the language features of Zig which hurts the advantages that semantic highlighting can offer over syntactic highlighting.

Custom semantic tokens could enable ZLS to eventually explore new features like the following:
- differentiate between mutable and non-mutable values https://github.com/ziglang/vscode-zig/issues/213
- differentiate between different "kinds" of builtin functions https://github.com/zigtools/zls/discussions/2104
- highlight code that may invoke safety-protected Undefined Behavior (e.g. highlight `@enumFromInt` for non-exhaustive enums but not for exhaustive enums, `@intCast` but not `@truncate`, `+` but not `+|`, etc.)
- highlight inline functions at the callsite
- highlight escape sequences inside string literals
- highlight format specifiers like `"{s}"`

This issue encompasses two important tasks:
- specify a set of token types and token modifiers that are tailored to Zig
- investigate how various editors deal with custom semantic tokens

As far as I can tell, [rust-analyzer](https://rust-analyzer.github.io/manual.html#semantic-syntax-highlighting) uses a lot of non standard tokens. Their approach may serve as good source of inspiration.

arrufat:
> * investigate how various editors deal with custom semantic tokens

Regarding this part, I have experience using @kakoune-lsp, which provides a straightforward approach. Here's [my configuration](https://github.com/arrufat/config/blob/bfce4a0c89083ce4431f0b9cecc5aaee21c6c077/kak/languages.kak#L147-L212), where `face` fields are faces defined in the Kakoune editor, which can be set to any combination of color/bold/italic/underline/double underline/curly underline, as shown [here](https://github.com/arrufat/config/blob/master/kak/colors/nord.kak).

PR #2243, opened 2025-04-07
add semantic tokens for escape sequences

### LINKED ISSUE #1928: Special characters not highlighted in strings
### Zig Version

0.12.1

0.13.0, 0.12.0

nvim 0.10, lspconfig

Open file with a special character like \n or \t in a string literal. See that all characters have the same color.
image

### Expected Behavior

Special characters highlighted differently
image

Image above shows what it looks like when ZLS is turned off. 

### Relevant log output

_No response_

xdBronch:
contributes to #2106, closes #1928
i chose `escapeSequence` because rust-analyzer uses it and a couple themes i tried recognized it, though i probably wouldve expected something like `string.escape` to be used.

### DIFF
--- a/src/Server.zig
+++ b/src/Server.zig
@@ -80,6 +80,7 @@ const ClientCapabilities = struct {
     supports_will_save_wait_until: bool = false,
     supports_publish_diagnostics: bool = false,
     supports_code_action_fixall: bool = false,
+    supports_semantic_tokens_overlapping: bool = false,
     hover_supports_md: bool = false,
     signature_help_supports_md: bool = false,
     completion_doc_supports_md: bool = false,
@@ -466,6 +467,9 @@ fn initializeHandler(server: *Server, arena: std.mem.Allocator, request: types.I
                     }
                 }
             }
+        }
+        if (textDocument.semanticTokens) |semanticTokens| {
+            server.client_capabilities.supports_semantic_tokens_overlapping = semanticTokens.overlappingTokenSupport orelse false;
         }
     }
 
@@ -1538,6 +1542,7 @@ fn semanticTokensFullHandler(server: *Server, arena: std.mem.Allocator, request:
         null,
         server.offset_encoding,
         server.config.semantic_tokens == .partial,
+        server.client_capabilities.supports_semantic_tokens_overlapping,
     );
 }
 
@@ -1561,6 +1566,7 @@ fn semanticTokensRangeHandler(server: *Server, arena: std.mem.Allocator, request
         loc,
         server.offset_encoding,
         server.config.semantic_tokens == .partial,
+        server.client_capabilities.supports_semantic_tokens_overlapping,
     );
 }
 
--- a/src/features/semantic_tokens.zig
+++ b/src/features/semantic_tokens.zig
@@ -29,6 +29,7 @@ pub const TokenType = enum(u32) {
     modifier,
     comment,
     string,
+    escapeSequence,
     number,
     regexp,
     operator,
@@ -68,6 +69,7 @@ const Builder = struct {
     token_buffer: std.ArrayListUnmanaged(u32) = .empty,
     encoding: offsets.Encoding,
     limited: bool,
+    overlappingTokenSupport: bool,
 
     fn add(self: *Builder, token: Ast.TokenIndex, token_type: TokenType, token_modifiers: TokenModifiers) error{OutOfMemory}!void {
         const tree = self.handle.tree;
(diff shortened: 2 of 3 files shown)


Techatrix:
Could you describe how you tested these changes? Which editor(s) did you use, and did you need any special theme or extra configuration?

xdBronch:
i used neovim, [tokyonight](https://github.com/folke/tokyonight.nvim) has support for `escapeSequence` but i didnt try many more. i mostly tested using nvims `:Inspect` which shows highlighting info for the character under the cursor. no config was needed but i suppose it will if your theme doesnt support this
image

xdBronch:
im not sure what i had done before that was wrong but yes that does seem to work, thank you.
provided i actually add a `.string` token ;p
```diff
diff --git a/src/features/semantic_tokens.zig b/src/features/semantic_tokens.zig
index 6fbbbae20f..2ed608e9f1 100644
--- a/src/features/semantic_tokens.zig
+++ b/src/features/semantic_tokens.zig
@@ -698,6 +698,7 @@
         .string_literal,
         .char_literal,
         => {
+            try writeToken(builder, main_token, .string);
             const string_start = tree.tokenStart(tree.firstToken(node));
             const string = offsets.nodeToSlice(tree, node);
             var offset: usize = 0;
```
is the intention to just not provide `escapeSequence` if overlapping tokens arent supported?

Techatrix:
> is the intention to just not provide `escapeSequence` if overlapping tokens arent supported?

I did consider keeping the old behavior when overlapping tokens were unsupported but opted no to because I was uncertain about how it would affect other editors. If there is an editor that would benefit from such a change then we can consider adding it.

Also, when the `semanticTokens` config option is set to `partial`, the escape sequence token should still be added. (haven't verified)

xdBronch:
i think the behavior with your changes should be fine, we can reevaluate it if someone has a problem with any of it at some point. in the worst case its the same as status quo. 
> Also, when the semanticTokens config option is set to partial, the escape sequence token should still be added. (haven't verified)

hmm i dont see how it would end up doing that, but most other forms of highlighting (vim plugin, treesitter, textmate) seem to already detect escape sequences so its fine if we dont in such a case. if we're only adding the escape sequences with overlapping support should probably guard it so its not doing any useless work?
<details>
<summary>diff</summary>

```diff
diff --git a/src/features/semantic_tokens.zig b/src/features/semantic_tokens.zig
index 98219d30d4..a919469ee4 100644
--- a/src/features/semantic_tokens.zig
+++ b/src/features/semantic_tokens.zig
@@ -699,17 +699,19 @@
         .char_literal,
         => {
             try writeToken(builder, main_token, .string);
-            const string_start = tree.tokenStart(main_token);
-            const string = offsets.nodeToSlice(tree, node);
-            var offset: usize = 0;
-            while (offset < string.len) {
-                const slash_index = std.mem.indexOfScalarPos(u8, string, offset, '\\') orelse break;
-                offset = slash_index;
-                _ = std.zig.string_literal.parseEscapeSequence(string, &offset);
-                try builder.addDirect(.escapeSequence, .{}, .{
-                    .start = slash_index + string_start,
-                    .end = offset + string_start,
-                });
+            if (builder.overlappingTokenSupport) {
+                const string_start = tree.tokenStart(main_token);
+                const string = offsets.nodeToSlice(tree, node);
+                var offset: usize = 0;
+                while (offset < string.len) {
+                    const slash_index = std.mem.indexOfScalarPos(u8, string, offset, '\\') orelse break;
+                    offset = slash_index;
+                    _ = std.zig.string_literal.parseEscapeSequence(string, &offset);
+                    try builder.addDirect(.escapeSequence, .{}, .{
+                        .start = slash_index + string_start,
+                        .end = offset + string_start,
+                    });
+                }
             }
         },
         .multiline_string_literal => {
```
</details>

Techatrix:
> > Also, when the semanticTokens config option is set to partial, the escape sequence token should still be added. (haven't verified)
> 
> hmm i dont see how it would end up doing that

I just checked and it doesn't actually work because `.escapeSequence` is also not emitted [here](https://github.com/zigtools/zls/blob/0acc14dcf20d45975a6f26100a568c44e9024e7c/src/features/semantic_tokens.zig#L161) when `semantic_tokens == "partial"`. But otherwise it could work because the string token is not going be generated which then means that the escape sequence tokens are not going to overlap so they would be added to the output. But anyway, this isn't really important.

> if we're only adding the escape sequences with overlapping support should probably guard it so its not doing any useless work?

makes sense, should have though of that.

xdBronch:
changes pushed, thanks for your help as always

Techatrix (approved):

PR #2244, opened 2025-04-09
add mutable semantic tokens

xdBronch:


src/analysis.zig
```diff
@@ -348,22 +348,23 @@ pub fn isInstanceCall(
 
     std.debug.assert(container_ty.is_type_val);
 
-    return analyser.firstParamIs(func_ty, container_ty);
+    return analyser.firstParamIs(func_ty, container_ty, null);
 }
 
-pub fn hasSelfParam(analyser: *Analyser, func_ty: Type) error{OutOfMemory}!bool {
```
xdBronch, line 354: could this be done better without an out param maybe?

src/features/semantic_tokens.zig
```diff
@@ -836,6 +830,30 @@ fn writeNodeTokens(builder: *Builder, node: Ast.Node.Index) error{OutOfMemory}!v
             try writeToken(builder, main_token, token_type);
             try writeNodeTokens(builder, rhs);
         },
+        .assign_bit_and,
```
xdBronch, line 833: got this idea from rust-analyzer, i think its pretty neat though i dont have any strong feelings about including it

src/features/semantic_tokens.zig
```diff
@@ -1070,15 +1085,40 @@ fn writeIdentifier(builder: *Builder, name_token: Ast.TokenIndex) error{OutOfMem
...
-        const is_param = child.decl == .function_parameter;
+        const is_param, const is_first_param = switch (child.decl) {
+            .function_parameter => |param| .{ true, param.param_index == 0 },
+            else => .{ false, false },
+        };
 
         if (try child.resolveType(builder.analyser)) |decl_type| {
-            return try colorIdentifierBasedOnType(builder, decl_type, name_token, is_param, .{});
+            const container_type = try builder.analyser.innermostContainer(handle, tree.tokenStart(name_token));
+            const mutable_self_param = is_first_param and mut: {
```
xdBronch, line 1095: this logic is taken from `Analyser.firstParamIs`, any better way?

src/features/semantic_tokens.zig
```diff
@@ -1070,15 +1085,40 @@ fn writeIdentifier(builder: *Builder, name_token: Ast.TokenIndex) error{OutOfMem
         name,
         tree.tokenStart(name_token),
     )) |child| {
-        const is_param = child.decl == .function_parameter;
+        const is_param, const is_first_param = switch (child.decl) {
```
xdBronch, line 1088: i decided to just make this apply to the "self" parameter, not pointers in general

tests/lsp_features/semantic_tokens.zig
```diff
@@ -1907,6 +1907,193 @@ test "weird code" {
...
+
+        .{ "const", .keyword, .{} },
+        .{ "beta", .variable, .{ .declaration = true, .mutable = false } },
+        .{ "=", .operator, .{} },
+        .{ "0", .number, .{} },
+        .{ "=", .operator, .{} },
+        .{ "beta", .variable, .{ .mutable = false } },
+    });
+    try testSemanticTokens(
+        \\fn foo() void {
```
xdBronch, line 1932: some of these have to go in functions since theyre not valid at container level

Techatrix:
If I am reading the code correctly, these are the rules that decide whether the `mutable` token modifier is added:

1. an identifier that refers to a non-const variable declaration. Same as asking whether the identifier can be reassigned.
2. an identifier that refers to the first parameter of a "method" like function that is taken as a non-const pointer.
3. an identifier that refers to a function call with "method" like syntax where the first parameter is a non-const pointer.
4. assignment operators except for `=` (e.g. `+=`, `+=`, `>>="`)

The first rule is the same as what I had previously experimented with in https://github.com/zigtools/zls/compare/master...techatrix/st-readonly-behaviour so I have no issues with that. 

For rule 2 and 3 I would like to see a good justification for why such an oddly specific rule should be added. Or rather, remove these rules entirely. About rule 4, everyone already knows that assignment operators are mutating and the exception of `=` makes this even more confusing. This gives no extra information that can't already be inferred from syntax alone.

The reason why I only implemented the first rule was because it is easy to understand and implement. More sophisticated rules will usually bring pointers into the mix and pointers are a mess.

xdBronch:
for 2 and 3 im not necessarily happy with the specific logic behind them or the special casing but i think the concept is good, i can see value in being able to tell if something might mutate your stuff without needing to look at the signature. 
imo the more important part is marking the function calls as mutable and i added it to the `self` parameter as well as a way to sort of "justify" the function being considered mutable since it relies on the parameter.

4 i really just got the idea from rust-analzer and thought it was interesting. my justification for not including `=` is that it appears far more often than the others and would end up pretty noisy, plus some uses of it dont really portray mutability, e.g. declarations or `_ = ...`. no strong feelings about keeping it

Techatrix:
2 & 3: Identifying modified values isn't easily done in Zig and solely highlight one of many situations that may modify a value does not help. Please remove it.

4: Just because rust-analyzer is doing something, doesn't mean that ZLS should do the same. Please remove it.

tests/lsp_features/semantic_tokens.zig
```diff
@@ -1960,6 +1960,66 @@ test "weird code" {
     , &.{}, .{ .mode = .zon });
 }
 
+test "mutable" {
```
Techatrix, line 1963: These tests don't cover anything that isn't already covered by the existing tests. They should be removed.

Techatrix (changes requested):
Consider this example:
```zig
const S = struct {
    var foo = 5;
};
const bar = S.foo;
```

The `foo` token in `S.foo` should be marked as mutable but isn't.

This patch should suffice:
```patch
index 9fb2af8..31ae5c8 100644
@@ -936,7 +936,7 @@ fn writeNodeTokens(builder: *Builder, node: Ast.Node.Index) error{OutOfMemory}!v
                 }
 
                 if (try decl_type.resolveType(builder.analyser)) |resolved_type| {
-                    try colorIdentifierBasedOnType(builder, resolved_type, field_name_token, false, .{});
+                    try colorIdentifierBasedOnType(builder, resolved_type, field_name_token, false, .{ .mutable = !decl_type.isConst() });
                     return;
                 }
             }
```

For test coverage, the existing ["field access"](https://github.com/zigtools/zls/blob/94546b33bef01f4ebebc3cb18ad6ccb9b351a042/tests/lsp_features/semantic_tokens.zig#L393) test should be expanded to check `var` and `const`.

Techatrix (approved):

PR #2246, opened 2025-04-11
return all referenced types for typeDefinition

### LINKED ISSUE #2052: better `typeDefiniton` on functions
currently asking for the type definition of a function just goes to the beginning of the functions definition. im not sure what exactly is the general consensus for what this should do (or is it specified exactly in the standard?) but clangd and rust-analyzer seem to take you to the definition of the return type, dont have anything else installed rn :p. naively i think maybe itd be nice to give you the definitions of all the params too?

xdBronch:
this needs more (any) testing but just throwing it up for now, its quite possibly not too good! but i think the results can be pretty desirable.
closes #2052

### DIFF
--- a/src/analysis.zig
+++ b/src/analysis.zig
@@ -5600,6 +5600,13 @@ fn addReferencedTypes(
         },
 
         .ip_index, .compile_error => {},
-        .either => {}, // TODO
+        .either => |either| {
+            for (either) |entry| {
+                try analyser.addReferencedTypes(
+                    .{ .data = entry.type_data, .is_type_val = ty.is_type_val },
+                    .{ .referenced_types = referenced_types },
+                );
+            }
+        },
     }
 }
--- a/tests/lsp_features/definition.zig
+++ b/tests/lsp_features/definition.zig
@@ -31,12 +31,12 @@ test "global variable" {
     );
 
     try testDefinition(
-        \\const S = <tdef>struct</tdef> { alpha: u32 };
+        \\const <tdef>S</tdef> = struct { alpha: u32 };
         \\const <>s: S  = S{ .alpha = 5 };
     );
 
     try testDefinition(
-        \\const S = <tdef>struct</tdef> { alpha: u32 };
+        \\const <tdef>S</tdef> = struct { alpha: u32 };
         \\const <>s = S{ .alpha = 5 };
     );
 }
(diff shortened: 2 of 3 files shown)


src/analysis.zig
```diff
@@ -5600,6 +5600,13 @@ fn addReferencedTypes(
         },
 
         .ip_index, .compile_error => {},
-        .either => {}, // TODO
+        .either => |either| {
```
xdBronch, line 5603: this feels too easy to work...

tests/lsp_features/definition.zig
```diff
@@ -107,8 +107,8 @@ test "struct init" {
 
 test "decl literal on generic type" {
     try testDefinition(
-        \\fn Box(comptime T: type) type {
-        \\    return <tdef>struct</tdef> {
+        \\fn <tdef>Box</tdef>(comptime T: type) type {
```
xdBronch, line 110: is this change desirable here?

tests/lsp_features/definition.zig
```diff
@@ -242,6 +242,14 @@ test "non labeled break" {
     );
 }
 
+test "function references" {
+    try testDefinition(
+        \\const <tdef>A</tdef> = struct {};
+        \\const <tdef>B</tdef> = struct {};
+        \\fn fo<>o(_: A) B {}
```
xdBronch, line 249: this fails, need to update the tests to work with `array_of_DefinitionLink`

tests/lsp_features/definition.zig
Techatrix: I'm mostly indifferent to this change. But if we are going to change this then we should at least remain consistent:
```
const <tdef>Foo</tdef> = struct {};
var <>foo = Foo{};
```
```
const Foo = <def>struct</def> {};
var foo = <>Foo{};
```
Either both go to `Foo` or both go to `struct`.

This will also need to handle cases like these where it needs to fall back to the old behavior:
```zig
const s = (struct { alpha: u32 }){ .alpha = 5 };
```
```zig
const Foo = (struct {});
var foo = Foo{};
```

src/analysis.zig
Techatrix: Maybe because it's just that simple. A test case would be need however.

This logic has since been moved [here](https://github.com/zigtools/zls/blob/6c2b97fdd66bcd0484b233163808613c2666db95/src/analysis.zig#L4372).

techatrix:
This PR has gotten out of date so I am closing this for now. Feel free to rebase and open a new PR if there's still interest.

PR #2247, opened 2025-04-14
Resolve types of fields in structs as arguments of decl literals

FnControlOption:
Also started adding tests for the other node types supported by `resolveExpressionTypeFromAncestors`

Just to make sure I'm on the right track, please review the tests I have so far


Techatrix (changes requested):
The `tests/analysis/field_init.zig` file is not specifically about field inits. It's about result location types so it should be named after that.

Techatrix (approved):

PR #2251, opened 2025-04-15
Set readonly token modifier for `const` variables and parameters

LucasSantos91:
The LSP protocol has support for `readonly` tokens, which, in zig terms, corresponds to `const`. This PR enables coloring `const` and `var` variables differently, like so:
Color demonstration

### DIFF
--- a/src/features/semantic_tokens.zig
+++ b/src/features/semantic_tokens.zig
@@ -479,7 +479,7 @@ fn writeNodeTokens(builder: *Builder, node: Ast.Node.Index) error{OutOfMemory}!v
             if (switch_case.ast.values.len == 0) try writeToken(builder, switch_case.ast.arrow_token - 1, .keyword);
             if (switch_case.payload_token) |payload_token| {
                 const actual_payload = payload_token + @intFromBool(tree.tokenTag(payload_token) == .asterisk);
-                try writeTokenMod(builder, actual_payload, .variable, .{ .declaration = true });
+                try writeTokenMod(builder, actual_payload, .variable, .{ .declaration = true, .readonly = true });
             }
             try writeNodeTokens(builder, switch_case.ast.target_expr);
         },
@@ -495,7 +495,7 @@ fn writeNodeTokens(builder: *Builder, node: Ast.Node.Index) error{OutOfMemory}!v
             if (while_node.payload_token) |payload| {
                 const capture_is_ref = tree.tokenTag(payload) == .asterisk;
                 const name_token = payload + @intFromBool(capture_is_ref);
-                try writeTokenMod(builder, name_token, .variable, .{ .declaration = true });
+                try writeTokenMod(builder, name_token, .variable, .{ .declaration = true, .readonly = true });
             }
             if (while_node.ast.cont_expr.unwrap()) |cont_expr| try writeNodeTokens(builder, cont_expr);
             try writeNodeTokens(builder, while_node.ast.then_expr);
@@ -529,7 +529,7 @@ fn writeNodeTokens(builder: *Builder, node: Ast.Node.Index) error{OutOfMemory}!v
                 capture_token = name_token + 2;
 
                 if (tree.tokenTag(name_token) != .identifier) continue;
-                try writeTokenMod(builder, name_token, .variable, .{ .declaration = true });
+                try writeTokenMod(builder, name_token, .variable, .{ .declaration = true, .readonly = true });
             }
             try writeNodeTokens(builder, for_node.ast.then_expr);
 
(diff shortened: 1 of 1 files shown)


Techatrix:
I have previously looked into adding the `readonly` modifier to ZLS which I mentioned in https://github.com/ziglang/vscode-zig/issues/213#issuecomment-2190097399. TLDR highlighting `const` variables in Zig is very noisy. Highlighting `var` variables appears to be the better alternative. A potential implementation of this can be found in https://github.com/zigtools/zls/pull/2244 which adds a `mutable` modified instead. This conflicts with this PR.

LucasSantos91:
Fully agree with your comment:

> Even though I think that an LSP should not decide how to emit semantic token information based on how it affects various editors

Personally, I don't see anything wrong with how VS Code displays the readonly variables. But in any case, the coloring that you didn't like happened because of this [line](https://github.com/microsoft/vscode/blob/9b0cd6551590014e8b91e78425848ccb6a12c6e6/extensions/theme-defaults/themes/dark_plus.json#L105C5-L105C31):

```		{
			"name": "Constants and enums",
			"scope": [
				"variable.other.constant", // Delete this
				"variable.other.enummember"
			],
			"settings": {
				"foreground": "#4FC1FF",
			}
```

In order to fix the issue for you, I've done it already. Just clone [this repo](https://github.com/LucasSantos91/Better-Dark-Modern) into your VS Code extensions folder. We could do this more officially, even add a Zig dedicated theme. But in order to do this properly, we need the LSP to provide us with information, as discussed below.
In Neovim, I've been having a lot of success with this fork. My first picture was exaggerated for demonstration purposes. Here's a more classy color scheme:
color demonstration

I have a branch where, on top of the changes I did here, I also added the static flag where appropriate. My goal was to make non-local constants and variables identifiable at a glance, because I hate looking around a function only to realize an identifier came from outside. I haven't finished altering the colorscheme yet, so pardon if it's a bit tacky, but it drives the point. I was going to PR this branch after this one, because I didn't want to mix readonly and static in one go.
You can see that I added a red tint to static variables. The red is more prominent in mutable static variables (because they're bad). For locals, I did shades of blue, going from more constant to less constant: parameters are dark blue because we have no control over them and are constant; local constants are intermediate blue because we have some control over them but are still constant; local mutable variables are more white-ish, because it draws more attention. Fields are also slightly darker than their parent.
These advanced colorschemes are only possible with rich information from the LSP. By not providing information because of VS Code ugliness, we are levelling down. Also, as shown above, it's trivial to make your own colorschemes. Non-standard modifiers are also not ideal, because we might have colorschemes that are leveraging these modifiers already, and it would be good if we could carry the same colorscheme to as many languages as possible.
Other solutions to the VS Code ugliness:
1) In your settings.json, add this:
```
"editor.semanticTokenColorCustomizations": {
    "rules": {
        "variable.readonly": { "foreground": "#ff0000" }
    }
}
```

2) There's the command-line option that you refered. I think there's a less intrusive way to make it happen, which would take a couple lines:
```
//src/features/semantic_tokens.zig
fn writeTokenMod(builder: *Builder, token_idx: ?Ast.TokenIndex, tok_type: TokenType, tok_mod: TokenModifiers) !void {
    if (token_idx) |ti| {
+      var new_tok_mod = tok_mod;
+      switch(readonly_option){
+          .normal => {},
+          .flip => { new_tok_mod.readonly = !new_tok_mod.readonly; },
+          .disable => { new_tok_mod.readonly = false; },
+      }
        try builder.add(ti, tok_type, tok_mod);
    }
}
```
`readonly_option` would have a corresponding command-line option with these possible values:
- normal: Apply readonly where expected.
- flip: Flip the meaning of readonly. `const` variables will not have it, and `var` variable will have it.
- disable: Don't send the readonly modifier.

Techatrix:
> In order to fix the issue for you, I've done it already. Just clone [this repo](https://github.com/LucasSantos91/Better-Dark-Modern) into your VS Code extensions folder.

> Other solutions to the VS Code ugliness:

Sure, I could customize my theme to change this but I still think that adding the `readonly` token modifier would be a bad addition for semantic highlighting in Zig. 

Let me continue on my previous linked comments: Themes will usually "react" to the presence of a token modifier and not their absence. If no token modifiers are present, the theme will choose the "normal" styling for the given token type. In the context of variables, a token modifier like `readonly` carries the hidden assumption that variables are usually not read-only or immutable. This becomes clear when looking at many of the popular programming languages like C, Python, Go, Java where constness i.e. `readonly` is not possible or opt-in but rarely the default. Zig on the other hand is different in this regard. It strongly favors constants. Even enforcing constness as the only option in cases like parameters and captures. A token modifier like `mutable` can more effectively express the fact that mutability is the special case and constness is the default.

I did not know about this before but rust-anlyzer uses [`mutable`](https://rust-analyzer.github.io/book/features.html?highlight=mutable#token-modifiers) over `readonly`. Probably because of similar reasoning since mutable variables in Rust are opt-in.

> These advanced colorschemes are only possible with rich information from the LSP. By not providing information because of VS Code ugliness, we are levelling down.

The `mutable` token modifier will still give the same information as `readonly` which should not inhibit themes to make use of this.

I also did not reach my opinion on this solely because I found the styling in VS Code default dark theme unappealing. As mention in my linked comment, I had also looked at other editors to see if themes would also react to the `readonly` token modifier but none of the tested themes/editors did. I do have to admit that I could have tried to look into more editors and themes. If you are aware of any, especially default or popular themes, then please let me know. But I wouldn't be surprised if the same argument couldn't be applied to other themes: `readonly` gets "special" styling while not `readonly` gets the "normal" styling.

> 2\. There's the command-line option that you refered. I think there's a less intrusive way to make it happen, which would take a couple lines:

This config option has been added for testing purposes only. I do not plan to actually add this to ZLS.

PR #2252, opened 2025-04-15
fix null value crash in analysis.zig

The-Emperor10:
```zig
fn bar(_: type) void {}
// crash doesn't occur if the return type's paramater has a name. for example: fn foo() fn(_: type) void { ... }
fn foo() fn(type) void {
    return bar;
}
const foobar = foo();
// crash occurs only if the result is assigned to a declaration.
const crash = foobar(u69);
```

This file results in a null value being used crashing zls. I added a null check for `param.name_token` to prevent that.


Techatrix (approved):
Thank you for the fix.

PR #2256, opened 2025-04-17
Fix alignment specs for Zig 0.15.0-dev.369

stephenmartindale:
Attempting to build `zls` with Zig `0.15.0-dev.369` (that being the latest release up on https://ziglang.org/download/ today) was failing because of a few changes where functions in `std` were expecting `std.mem.Alignment.of` instead of alignments specified as `comptime int`s.

I fixed it so the project builds with this nightly. Here is the two-line patch.

I was not able to determine, exactly, which upstream `zig` PR changed is API in `std` but this might be of relevance:

- https://github.com/ziglang/zig/issues/23548

### DIFF
--- a/src/DocumentStore.zig
+++ b/src/DocumentStore.zig
@@ -677,7 +677,7 @@ pub fn getOrLoadHandle(self: *DocumentStore, uri: Uri) ?*Handle {
         file_path,
         max_document_size,
         null,
-        @alignOf(u8),
+        std.mem.Alignment.of(u8),
         0,
     ) catch |err| {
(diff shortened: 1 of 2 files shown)


Askedis:
There is another `@alignOf(u8)` in `tests/analysis_check.zig` line 107, which prevents your build from passing the tests.

I'm just not sure of the github etiquette and if this message is enough to warn you of the problem, or if I should directly do the fixing commit somewhere myself.

Techatrix:
See dbcfaac493ccee788e01a4d25a935f4fd05fe18c...dca91b8091b935328fe1e05d261e2c10af8319f6

PR #2259, opened 2025-04-18
More result location types

FnControlOption:
```zig
const grouped_expression: StructType = (.{ .foo = 1 });
//                                         ^^^^ (u32)()

const @"orelse" = some_optional orelse .{ .foo = 1 };
//                                        ^^^^ (u32)()

const @"catch" = some_error_union catch .{ .foo = 1 };
//                                         ^^^^ (u32)()
```

### DIFF
--- a/src/analysis.zig
+++ b/src/analysis.zig
@@ -5346,6 +5346,26 @@ pub fn resolveExpressionTypeFromAncestors(
             );
         },
 
+        .builtin_call,
+        .builtin_call_comma,
+        .builtin_call_two,
+        .builtin_call_two_comma,
+        => {
+            var buffer: [2]Ast.Node.Index = undefined;
+            const params = tree.builtinCallParams(&buffer, ancestors[0]).?;
+            const call_name = tree.tokenSlice(tree.nodeMainToken(ancestors[0]));
+
+            if (std.mem.eql(u8, call_name, "@as")) {
+                if (params.len != 2) return null;
+                if (params[1] != node) return null;
+                const ty = try analyser.resolveTypeOfNode(.{
+                    .node = params[0],
+                    .handle = handle,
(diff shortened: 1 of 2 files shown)


src/analysis.zig
```diff
@@ -5335,14 +5335,60 @@ pub fn resolveExpressionTypeFromAncestors(
...
+        .builtin_call_comma,
+        .builtin_call_two,
+        .builtin_call_two_comma,
+        => {
+            var buffer: [2]Ast.Node.Index = undefined;
+            const params = tree.builtinCallParams(&buffer, ancestors[0]).?;
+            const call_name = tree.tokenSlice(tree.nodeMainToken(ancestors[0]));
+
+            if (std.mem.eql(u8, call_name, "@as")) {
+                std.debug.assert(params.len == 2);
```
Techatrix, line 5359: This assertion is not valid:
```zig
const foo = @as(.{ .fo<cursor here>o = 5 });
```

Techatrix (changes requested):

src/analysis.zig
FnControlOption: Dropped the commit for `@as` and may revisit it in a future PR

Techatrix: Why not just change it to `if (params.len != 2) return null;`?

FnControlOption: I couldn’t figure out a way to add a test where the parameter count check would be covered

Techatrix: I was hoping that something like this would work:
```zig
const builtin_as2 = @as(.{ .foo = 1 });
//                         ^^^^ (unknown)()
```
But this just returns `error.IdentifierNotFound`. It should be ok to go without a test for now until this is supported.

Techatrix (approved):

PR #2260, opened 2025-04-18
Adds static modifier to semantic tokens where appropriate.

LucasSantos91:
This enables coloring local and static variables differently:
color demonstration

It does not affect VS Code's default themes:
vscode demonstration


Techatrix (changes requested):
The [test coverage for semantic tokens](https://github.com/zigtools/zls/blob/master/tests/lsp_features/semantic_tokens.zig) should be updated. You can use `zig build test -Dtest-filter="lsp_features.semantic_tokens"` to run them.

I pushed a commit to your PR that should improve the error message from the tests when it reports incorrect token modifiers. Hope that helps.

Also be aware that this PR will need to be rebased if #2244 is merged.

src/features/semantic_tokens.zig
```diff
@@ -921,6 +921,32 @@ fn writeNodeTokens(builder: *Builder, node: Ast.Node.Index) error{OutOfMemory}!v
...
+                    }
+                    current_node = left_node;
+                    left_node, current_field_name_token = tree.nodeData(left_node).node_and_token;
+                    left_type = try builder.analyser.resolveTypeOfNode(.{
+                        .node = left_node,
+                        .handle = handle,
+                    }) orelse break :blk .{};
+                    left_type = try builder.analyser.resolveDerefType(left_type) orelse left_type;
+                }
+            };
```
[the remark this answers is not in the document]
LucasSantos91: The problem with this approach is that it only applies to the case where this branch takes the success path:

```zig
+993  if (try lhs_type.lookupSymbol(builder.analyser, symbol_name)) |decl_type| 
```

Which means that the last code path:
```zig
+999            try writeTokenMod(builder, field_name_token, .variable, token_mod);
```
Doesn't get the information of whether this token is static or not.
This part handles fields of variables. We could remove `token_mod` and go back to `.{}`, but then fields of static variables would not get the static modifier.

src/features/semantic_tokens.zig
```diff
@@ -1088,17 +1100,111 @@ fn writeIdentifier(builder: *Builder, name_token: Ast.TokenIndex) error{OutOfMem
...
+                tree.tokenStart(lhs_token),
+            ) orelse break :static_blk false;
+            break :static_blk try lhs_decl.isStatic();
+        };
+        try writeTokenMod(
+            builder,
+            field_name_token,
+            .variable,
+            .{ .static = is_static },
+        );
```
Techatrix, line 1155: This code assumes that the lhs of a field access is guaranteed to be an identifier even though it could be any kind of expression.

This still passes all the tests:
```suggestion
        try writeToken(builder, field_name_token, .variable);
```
No need to over complicate this.

src/analysis.zig
```diff
@@ -3097,6 +3097,11 @@ pub const Type = struct {
         };
     }
 
+    /// If this returns false, the type cannot have the static modifier
+    pub fn isTypeOrFunc(self: Type) bool {
+        return self.is_type_val or self.isFunc() or self.isTypeFunc();
```
Techatrix, line 3102: The `self.isTypeFunc()` is unnecessary and the doc comment should talks about the static modifier even though the function itself has nothing to do with it.

Remove this function and use `self.is_type_val or self.isFunc()` directly.

Techatrix (changes requested):

Techatrix (approved):
Looks good! Thanks for your work.

PR #2265, opened 2025-04-19
Fix various resolved types of const pointers

FnControlOption:
Notable addition: `resolveMutability` is similar to `resolveTypeOfNodeUncached` but returns `.@"var"`, `.@"const"`, or `null`, depending on whether the given AST node refers to a known variable/constant. (I went with an enum instead of a boolean for readability's sake, but I have no strong preference for one or the other.)


FnControlOption:
Will close this PR once all commits have been ported to the new design in #2271

PR #2266, opened 2025-04-20
add prepareRename support

xdBronch:
this allows editors to get proper placeholder text for renaming irregular identifiers with `@""` syntax, before theyd either get the word youre inside or make no attempt at a placeholder.
wasnt sure where/if to add tests, all the logic here is effectively already used and in turn tested in the regular references handler

### DIFF
--- a/src/Server.zig
+++ b/src/Server.zig
@@ -575,7 +575,9 @@ fn initializeHandler(server: *Server, arena: std.mem.Allocator, request: types.I
                     .willSaveWaitUntil = true,
                 },
             },
-            .renameProvider = .{ .bool = true },
+            .renameProvider = .{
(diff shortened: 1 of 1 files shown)


Techatrix:
Please do not give your branches nonsensical names.

Techatrix (approved):

xdBronch:
using better names isnt a problem but fwiw im not going out of my way to call them this :laughing:. i use [jujutsu](https://github.com/jj-vcs/jj) which is generally not branch oriented like git so theres a flag to automatically create a branch when you push and it gets the name from the current commit

Techatrix:
I am not familiar with jujutsu. If you are able to give your branches short and descriptive names then do that. If not, leave them as is.

PR #2267, opened 2025-04-20
Add hover text for primitives

FnControlOption:
For example:

```zig
bool
(type)
```

```zig
true
(bool)
```

```zig
null
(@TypeOf(null))
```


FnControlOption:
Need to check how this handles escaped identifiers like `@“bool”`

PR #2271, opened 2025-04-20
Fix resolved type of pointer to const identifier

FnControlOption:
Alternative to #2265

- Added a `Binding` struct that has two fields: `type: Type` and `is_const: bool`. Currently, `is_const` is `true` for four cases:
  1. An identifier that refers to a primitive, such as `i32` and `true`
  1. An identifier that refers to a bound type param
  1. An identifier `foo` that refers to a previously declared `const foo`
  1. An expression of the form `&foo` (i.e. `&&foo` is of type `*const *Foo` for a given `var foo: Foo`)
- Added three new functions: `resolveBindingOfNode{,Internal,Uncached}`
  - `resolveTypeOfNode{,Internal}` are now just helper functions that delegate to `resolveBindingOfNode{,Internal}`
  - `resolveBindingOfNodeUncached` currently only handles `identifier` and `address_of` nodes, while the rest are still handled by `resolveTypeOfNodeUncached`.
  - The current design allows code in `resolveTypeOfNodeUncached` to be ported incrementally to `resolveBindingOfNodeUncached`.


FnControlOption:
Updated PR description

PR #2277, opened 2025-04-23
Some improvements to hover text

FnControlOption:
- Turned off parameter names when printing function types, which is consistent with how they're normally printed via `std.fmt`
- Added the resolved type (in the parentheses) of a function when hovering over it, which makes the "Go to ..." part less confusing
  Screenshot 2025-04-23 at 10 08 01 AM
- Merged `Analyser.addReferencedTypes` with `Type.format`, which has pretty much the same logic but also handles generic types
  Screenshot 2025-04-23 at 10 08 36 AM

### DIFF
--- a/src/analysis.zig
+++ b/src/analysis.zig
@@ -173,6 +173,7 @@ fn fmtSnippetPlaceholder(bytes: []const u8) std.fmt.Formatter(formatSnippetPlace
 }
 
 pub const FormatParameterOptions = struct {
+    referenced: ?*ReferencedType.Set = null,
     info: Type.Data.Parameter,
     index: usize,
 
@@ -198,6 +199,7 @@ pub fn formatParameter(
 
     const analyser = ctx.analyser;
     const data = ctx.options;
+    const referenced = data.referenced;
     const info = data.info;
 
     if (data.index != 0) {
@@ -234,7 +236,10 @@ pub fn formatParameter(
         if (has_parameter_name) try writer.writeAll(": ");
 
         if (info.type) |ty| {
-            try writer.print("{}", .{ty.fmtTypeVal(analyser, .{ .truncate_container_decls = true })});
+            try writer.print("{}", .{ty.fmtTypeVal(analyser, .{
+                .referenced = referenced,
+                .truncate_container_decls = true,
+            })});
         } else {
             try writer.writeAll("anytype");
         }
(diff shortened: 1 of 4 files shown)


src/analysis.zig
```diff
@@ -285,7 +315,7 @@ pub fn formatFunction(
                                 if (is_comma or tag == .keyword_const) try writer.writeByte(' ');
                             }
                         } else {
-                            try writer.writeAll(offsets.nodeToSlice(tree.*, type_expr));
+                            try analyser.writeTypeFromTypeNode(.of(type_expr, handle), referenced, writer);
```
Techatrix, line 318: formatting/stringifying a function should ideally not perform code analysis. Instead, `Analyser.Type.Data` should have a dedicated field for functions instead of the current `other`. This is more in line with how we represent other types like pointers or arrays.

This is change is going to require some more refactoring so I would suggest to leave that for a separate PR.

src/features/hover.zig
```diff
@@ -124,23 +96,23 @@ fn hoverSymbolRecursive(
...
     }
-    const referenced_types: []const Analyser.ReferencedType = type_references.keys();
+    const referenced_types: []const Analyser.ReferencedType = referenced.keys();
     return try hoverSymbolResolved(
         arena,
         markup_kind,
         doc_strings.items,
         def_str,
-        if (is_fn) null else resolved_type_str,
+        resolved_type_str,
```
Techatrix, line 115: This argument can no longer be null so `hoverSymbolResolved` should take a `[]const u8` instead of `?[]const u8`.

Techatrix (changes requested):

src/analysis.zig
FnControlOption: Would it be okay to merge this PR first? I've found the improvements to hover to be really helpful

Techatrix:
Why has this been converted into a draft? Is it waiting for #2296 perhaps?

FnControlOption:
> Is it waiting for #2296 perhaps?

Yes, but we can merge this one first if you prefer

tests/lsp_features/hover.zig
```diff
@@ -830,6 +903,9 @@ test "type reference cycle" {
         \\    beta: @TypeOf(alpha),
         \\) void
         \\```
+        \\```zig
+        \\(unknown)
```
FnControlOption, line 907: I have another PR in the works: FnControlOption@8d5878bb (Will open after this one is merged)

Techatrix (approved):
Not sure how I feel about bringing back the resolved function type after it has been removed in #1377. I suppose it a matter of personal preference about how much information you would like to see in the hover popup. 

I have recently found a possible solution to this problem that could become useful in the future. I will follow up about it in a new issue.

PR #2279, opened 2025-04-24
resolve field access on parenthesized initializer

xdBronch:


src/analysis.zig
```diff
@@ -3789,6 +3789,21 @@ pub fn getFieldAccessType(
...
+                while (next.tag != .eof) : (next = tokenizer.next()) {
+                    if (next.tag == .r_brace) {
+                        brace_count -= 1;
+                        if (brace_count == 0) break;
+                    } else if (next.tag == .l_brace) {
+                        brace_count += 1;
+                    }
+                } else return null;
+                if (current_type) |*ct| {
+                    ct.is_type_val = false;
```
xdBronch, line 3804: this feels a little dirty? without it the field access was on the type itself rather than an instance of it so field completions didnt work

Techatrix: How about something like this? This should make it more clear what this code is supposed to do.
```zig
if (current_type) |ct| {
    if (ct.isStructType() or ct.isUnionType()) {
        // struct initialization
        current_type = ct.instanceTypeVal(analyser);
    }
}
```

xdBronch: ooh yeah thats better

Techatrix (approved):

PR #2280, opened 2025-04-25
make sure the zls and zig versions are aligned

VivaThapelo:


Techatrix:
This makes no sense. The `-Dversion-string` option is meant to specify the ZLS version so that it isn't resolved using git.

PR #2281, opened 2025-04-26
fix(build): update tracy options to use `std.zig.SanitizeC`

WillLillis:
Upstream change made in ziglang/zig@b3537d0f4adeff824348a4918b495976ae230731

Edit: Looks like I  jumped the gun a bit, I think that build isn't available via https://ziglang.org/download/ yet.

### DIFF
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -4,7 +4,7 @@
     .version = "0.15.0-dev",
     // Must be kept in line with the `minimum_build_zig_version` in `build.zig`.
     // Should be a Zig version that is downloadable from https://ziglang.org/download/ or a mirror.
-    .minimum_zig_version = "0.15.0-dev.355+206bd1ced",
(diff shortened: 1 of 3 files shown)


build.zig.zon
```diff
@@ -4,7 +4,7 @@
     .version = "0.15.0-dev",
     // Must be kept in line with the `minimum_build_zig_version` in `build.zig`.
     // Should be a Zig version that is downloadable from https://ziglang.org/download/ or a mirror.
-    .minimum_zig_version = "0.15.0-dev.355+206bd1ced",
+    .minimum_zig_version = "0.15.0-dev.393+5668c8b7b",
```
Techatrix, line 7: Once https://ziglang.org/download/ provides a recent enough version, the `minimum_zig_version` should updated to that exact version so that is accessible to tooling.

The `minimum_build_zig_version` can remain at `0.15.0-dev.393+5668c8b7b`.

Techatrix (approved):

PR #2275, opened 2025-04-22
add some labelled switch stuff

xdBronch:

PR #2284, opened 2025-04-27
Fix label in switch statement

LucasSantos91:
When a switch was preceded by a label, the `keyword` semantic token was being assigned to the label instead of the switch token, and the label was not receiving its appropriate semantic token.


Techatrix:
#2275 already has a similar fix for this but also includes some other improvements related to labelled switch.

PR #2285, opened 2025-04-28
Only unwrap instances of error union type

FnControlOption:
```zig
const Error = error{ Foo, Bar };
const ErrorUnionType = Error!u32;
const InvalidErrorUnionTypeUnwrap = ErrorUnionType catch |err| err;
//    ^^^^^^^^^^^^^^^^^^^^^^^^^^^ (unknown)()
//                                                        ^^^ (unknown)()
```

### DIFF
--- a/src/analysis.zig
+++ b/src/analysis.zig
@@ -909,6 +909,8 @@ pub fn resolveAddressOf(analyser: *Analyser, ty: Type) error{OutOfMemory}!?Type 
 pub const ErrorUnionSide = enum { error_set, payload };
 
 pub fn resolveUnwrapErrorUnionType(analyser: *Analyser, ty: Type, side: ErrorUnionSide) error{OutOfMemory}!?Type {
+    if (ty.is_type_val) return null;
+
     return switch (ty.data) {
         .error_union => |info| switch (side) {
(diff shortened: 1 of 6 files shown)


Techatrix (changes requested):
There shouldn't be a `invalid.zig` that contains all invalid operations. It would become too large.

The (invalid) pointer operations should go to `pointer.zig`, (invalid) optional operations to `optional.zig` and (invalid) error set/union operations to `error.zig`. The last one doesn't exist yet.

FnControlOption:
> The (invalid) pointer operations should go to `pointer.zig`, (invalid) optional operations to `optional.zig` and (invalid) error set/union operations to `error.zig`.

Problem with that is, the files become un-compilable (or un-runnable, I don't remember exactly), which means we can't do `@compileLog` inside them

Techatrix:
The files are still syntactically valid and Zig would only resolve the declarations that are referenced by the `@compileLog` so I don't really see how this is different from the existing tests we have so far. As an example, files like `array.zig` `pointer.zig` and `builtins.zig` already contain declarations that would cause compile errors if referenced.

FnControlOption:
Ah sorry, you're right. I must have misremembered...

Techatrix (approved):

PR #2290, opened 2025-04-29
remove workaround for number literal hover with Zed editor

### LINKED ISSUE #2289: Markdown tables in hovers should include headers
Markdown tables without headers are not parsed as tables by some parsers (see https://stackoverflow.com/a/17543474), for example this is how hover over value does look in Zed editor:

Image

Also, you can check it yourself at https://markdownlivepreview.com, I see the same results there:

Image

I understand that empty header line may look ugly, but you can fill it with something relatively useful anyway:

Image

Techatrix:
Zed now appears to be able to properly render markdown so there should be no need to keep the workaround that has been added in #2058. This workaround was also missing a newline after the back ticks at the start which causes the first row in the table to be missing. fixes #2289

The downside is that the render markdown version overflows on small numbers and behaves very weird when trying to select text from the table.

Screenshot_20250429_184556

Tested with Zed 0.181.5 on Linux
@xdBronch In case you wish to comment on this.

### DIFF
--- a/src/Server.zig
+++ b/src/Server.zig
@@ -1726,7 +1726,6 @@ fn hoverHandler(server: *Server, arena: std.mem.Allocator, request: types.HoverP
         source_index,
         markup_kind,
         server.offset_encoding,
-        server.client_capabilities.client_name,
     );
 }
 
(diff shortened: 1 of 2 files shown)


Techatrix:
> im curious how/when this happened.

Bisecting Zed versions showed that it's been fixed somewhere between 0.177.11 and 0.178.5. This led me to https://github.com/zed-industries/zed/pull/25954

Techatrix:
The current behavior should be good enough to not need the workaround anymore. Having less editor specific workarounds is always better.

PR #2291, opened 2025-04-29
only give Struct completion type for type values

xdBronch:
the current way never made much sense to me and this seems more in line with what other LSPs do

### DIFF
--- a/src/features/completions.zig
+++ b/src/features/completions.zig
@@ -247,7 +247,16 @@ fn declToCompletion(builder: *Builder, decl_handle: Analyser.DeclWithHandle, opt
         => {
             var kind: types.CompletionItemKind = blk: {
                 const parent_is_type_val = if (options.parent_container_ty) |container_ty| container_ty.is_type_val else null;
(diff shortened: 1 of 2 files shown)


src/features/completions.zig
```diff
@@ -222,7 +223,6 @@ fn declToCompletion(builder: *Builder, decl_handle: Analyser.DeclWithHandle, opt
         if (resolved_ty.data != .compile_error) break :blk null;
 
         const node_with_handle = resolved_ty.data.compile_error;
-        const tree = node_with_handle.handle.tree;
```
Techatrix, line 225: `decl_handle.handle.tree` and `node_with_handle.handle.tree` aren't necessarily the same. The `tree` that is used in this scope assumes the latter.

Techatrix (changes requested):

src/features/completions.zig
xdBronch: didnt even notice they were using different trees :facepalm:

Techatrix (approved):

PR #2307, opened 2025-05-12
Resolve type of arithmetic expressions

FnControlOption:
Relevant sections of the Zig Language Reference:

- [Operators](https://ziglang.org/documentation/master/#Operators)
- [Pointers](https://ziglang.org/documentation/master/#Pointers)
- [C Pointers](https://ziglang.org/documentation/master/#C-Pointers)

### DIFF
--- a/src/analysis.zig
+++ b/src/analysis.zig
@@ -2510,6 +2510,7 @@ fn resolveTypeOfNodeUncached(analyser: *Analyser, node_handle: NodeWithHandle) e
         .bool_not,
         => return Type.fromIP(analyser, .bool_type, null),
 
+        .bit_not,
         .negation,
         .negation_wrap,
         => return try analyser.resolveTypeOfNodeInternal(.of(tree.nodeData(node).node, handle)),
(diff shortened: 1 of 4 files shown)


src/analysis.zig
```diff
@@ -2639,19 +2640,62 @@ fn resolveTypeOfNodeUncached(analyser: *Analyser, node_handle: NodeWithHandle) e
...
         .bit_and,
         .bit_xor,
         .bit_or,
-        .bit_not,
+        => {
+            const lhs, const rhs = tree.nodeData(node).node_and_node;
+            const lhs_ty = try analyser.resolveTypeOfNodeInternal(.of(lhs, handle)) orelse return null;
+            const rhs_ty = try analyser.resolveTypeOfNodeInternal(.of(rhs, handle)) orelse return null;
+            if (lhs_ty.is_type_val) return null;
+            if (rhs_ty.is_type_val) return null;
```
Techatrix, line 2655: No need to resolve the rhs when we are going to discard the result anyway.
```suggestion
            const lhs_ty = try analyser.resolveTypeOfNodeInternal(.of(lhs, handle)) orelse return null;
            if (lhs_ty.is_type_val) return null;
            const rhs_ty = try analyser.resolveTypeOfNodeInternal(.of(rhs, handle)) orelse return null;
            if (rhs_ty.is_type_val) return null;
```

The same change should be applied to the other cases.

Techatrix (changes requested):
The analysis tests should be in the top level scope if possible. If we want to use functions to better structure the tests, that should be done to all of them to be consistent.

The existing tests in `pointer.zig` have a specific layout that can be combined with the new tests to improve the structure. The tests go through all of the pointer kinds and then go through all the different operations. The new tests for `+` and `-` for each pointer kind fit right in.

Techatrix (approved):

PR #2308, opened 2025-05-12
Map files to their corresponding compilation steps

arielherself:
This PR solves the problem discussed in #2305 by categorizing build context (packages, c_macros and include_dirs) into different “compilation units,” which hold modules and files required by a single compilation step.

When ZLS analyzes a symbol, it will find the context of the compilation unit that the current file belongs to. When there are multiple of them, it picks a random one. However, this could be improved by calculating the longest common prefix between the current file and the root file of the compilation unit.

After the changes, only modules required by at least one compilation step (which could be an arbitrary one instead of just of the descendants of the default step) will be analyzed by ZLS. This behavior is also seen in other language servers like [rust-analyzer](https://github.com/rust-lang/rust-analyzer).

### DIFF
--- a/src/DocumentStore.zig
+++ b/src/DocumentStore.zig
@@ -113,10 +113,12 @@ pub const BuildFile = struct {
         const build_config = self.tryLockConfig() orelse return false;
         defer self.unlockConfig();
 
-        try package_uris.ensureUnusedCapacity(allocator, build_config.packages.len);
-        for (build_config.packages) |package| {
-            package_uris.appendAssumeCapacity(try URI.fromPath(allocator, package.path));
+        for (build_config.compilation_unit_info) |compilation_unit| {
+            for (compilation_unit.packages) |package| {
+                try package_uris.append(allocator, try URI.fromPath(allocator,package.path));
+            }
         }
+
         return true;
     }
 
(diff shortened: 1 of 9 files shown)


leecannon:
I have been using this for a couple days and have not hit any issues in a project with 58 “compilation units”.

techatrix:
The build runner is only responsible to collect information that is directly associated with the build system. All further tasks like tying a file to a `build.zig` in part of the document store which already implemented with [`getAssociatedBuildFileUri`](https://github.com/zigtools/zls/blob/c289cb71db210ff45a4a024851abb11ba69ca413/src/DocumentStore.zig#L334) and [`uriAssociatedWithBuild`](https://github.com/zigtools/zls/blob/c289cb71db210ff45a4a024851abb11ba69ca413/src/DocumentStore.zig#L1399C4-L1399C26) to find out whether a file is used by one of it's build system modules. This serves a similar purpose to what has been added to the build runner. This logic should instead be updated to return which module was responsible for the match. 

But more importantly, while this PR appears offer better resolution of a `root` module or similar. It still has an inaccurate view of how compile steps and modules work in the build system. A proper solution should be able to handle these situations:

- a compilation step only has a root module (modules form a graph not a list)
- modules consist of a root source file and can import other modules
- imported modules are not visible their parent module
- a module can be part of multiple compilation steps (may be to complex to support)
- a source file may be part of multiple modules by placing them in different compilation steps (may be to complex to support)

Keep in mind that the build runner code incorrectly uses the name "package" to refer to a "module". 

I appreciate the effort that has been put into this but I rather reject this PR until a more complete solution has been implemented. I plan to work on this myself and I would not recommend working on such a comparatively large task to someone who is new to this codebase.

PR #2310, opened 2025-05-12
Print all errors (with line and column) when running analysis tests

FnControlOption:
I find it useful to get all test failures at once


Techatrix (changes requested):
The changes to `ErrorBuilder.zig` should have test coverage. There are already some existing tests.

The error tracking in `analysis_check.zig` seems very useless. Why not just add something like this to end of the main function:

```zig
if (error_builder.hasMessages()) {
    return error.CheckFailed;
}
```

Techatrix (approved):

PR #2312, opened 2025-05-12
use resolved function type for arg completions

xdBronch:
before the test case would crash on `param_decl.get` since the node was not `.fn_*`


FnControlOption:
Do `node_type.data.function.node` and `node_type.data.function.handle` work? Originally I intended to delete those fields but it turns out generic return types still need them and maybe they’re also needed here

Techatrix (approved):

PR #2315, opened 2025-05-13
resolve parameters of function with invalid return type

### LINKED ISSUE #2314: Semantic hightlitghting does not work when return is @TypeOf(...)
### Zig Version

0.15.0-dev.515+833d4c9ce

### ZLS Version

0.15.0-dev.107+f0d580e5 (https://github.com/zigtools/zls/pull/2313)

Kakoune + Kakoune LSP

Define a function with return type as `@TypeOf(...)`
I noticed this when importing `std.math.clamp`.

### Expected Behavior

```zig
pub fn clamp_generic(val: anytype, lower: anytype, upper: anytype) @TypeOf(val, lower, upper) {}
pub fn clamp_concrete(_: f32, _: f32, _: f32) f32 {}

const clamp_gen = clamp_generic;
const clamp_con = clamp_concrete;
```

Image

As you can see, the `clamp_generic` function does not get properly highlighted outside its definition. However, the hover information shows it as a function:

```
╭───────────────────────────────────────────────────────────────────────────────────────────╮
│ fn clamp_generic(val: anytype, lower: anytype, upper: anytype) @TypeOf(val, lower, upper) │
│                                                                                           │
│ (unknown)                                                                                 │
╰───────────────────────────────────────────────────────────────────────────────────────────╯
```

### Relevant log output

```shell

```

xdBronch:
closes #2314

### DIFF
--- a/src/analysis.zig
+++ b/src/analysis.zig
@@ -2284,7 +2284,7 @@ fn resolveTypeOfNodeUncached(analyser: *Analyser, node_handle: NodeWithHandle) e
                             else => unreachable,
                         }
                     }
-                    break :param_type Type.fromIP(analyser, InternPool.Index.unknown_type, null);
+                    break :param_type Type.fromIP(analyser, .type_type, .unknown_type);
                 };
 
                 try parameters.append(analyser.arena.allocator(), .{
@@ -2295,9 +2295,8 @@ fn resolveTypeOfNodeUncached(analyser: *Analyser, node_handle: NodeWithHandle) e
                 });
             }
 
-            const return_type = try analyser.resolveReturnTypeInternal(handle, fn_proto) orelse {
-                return null;
-            };
+            const return_type = try analyser.resolveReturnTypeInternal(handle, fn_proto) orelse
+                Type.fromIP(analyser, .type_type, .unknown_type);
 
             const info: Type.Data.Function = .{
                 .handle = handle,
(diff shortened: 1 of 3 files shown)


src/analysis.zig
```diff
@@ -2284,7 +2284,7 @@ fn resolveTypeOfNodeUncached(analyser: *Analyser, node_handle: NodeWithHandle) e
                             else => unreachable,
                         }
                     }
-                    break :param_type Type.fromIP(analyser, InternPool.Index.unknown_type, null);
+                    break :param_type Type.fromIP(analyser, .unknown_type, null);
```
FnControlOption, line 2287:
```suggestion
                    break :param_type Type.fromIP(analyser, .type_type, .unknown_type);
```

src/analysis.zig
```diff
@@ -2295,9 +2295,8 @@ fn resolveTypeOfNodeUncached(analyser: *Analyser, node_handle: NodeWithHandle) e
                 });
             }
 
-            const return_type = try analyser.resolveReturnTypeInternal(handle, fn_proto) orelse {
-                return null;
-            };
+            const return_type = try analyser.resolveReturnTypeInternal(handle, fn_proto) orelse
+                Type.fromIP(analyser, .unknown_type, null);
```
FnControlOption, line 2299:
```suggestion
                Type.fromIP(analyser, .type_type, .unknown_type);
```

PR #2331, opened 2025-05-24
resolve relative paths in dynamic configurations

### LINKED ISSUE #2329: zls should resolve relative paths for didChangeConfiguration
#2288 was rejected because eg this `.zed/settings.json` file should be supported:
```json
{
  "lsp": {
    "zls": {
      "settings": {
        "zig_lib_path": "./zig/lib"
      }
    }
  }
}
```

however, the above configuration causes zls to fallback to the default zig lib path from `zig env` since `./zig/lib` is relative. as a result, zls logs the following:

```
config option 'zig_lib_path': expected absolute path but got './zig/lib'
```

dotcarmen:
closes #2329

this PR resolves relative paths in configurations when `.allowRelative = true` is passed to `Server.updateConfiguration`.

as discussed in https://github.com/zigtools/zls/issues/2329#issuecomment-2905746767, this behavior only works when there's exactly 1 workspace assigned to the current zls server.

### DIFF
--- a/src/Server.zig
+++ b/src/Server.zig
@@ -740,6 +740,20 @@ fn handleConfiguration(server: *Server, json: std.json.Value) error{OutOfMemory}
     const arena = arena_allocator.allocator();
 
     var new_config: configuration.Configuration = .{};
+
+    inline for (fields, result) |field, json_value| {
+        var runtime_known_field_name: []const u8 = ""; // avoid unnecessary function instantiations of `std.fmt.format`
+        runtime_known_field_name = field.name;
+
+        const maybe_new_value = std.json.parseFromValueLeaky(field.type, arena, json_value, .{}) catch |err| blk: {
+            log.err("failed to parse configuration option '{s}': {}", .{ runtime_known_field_name, err });
+            break :blk null;
+        };
+        if (maybe_new_value) |new_value| {
+            @field(new_config, field.name) = new_value;
+        }
+    }
+
     const maybe_root_dir: ?[]const u8 = if (server.workspaces.items.len == 1) dir: {
         const uri = std.Uri.parse(server.workspaces.items[0].uri) catch |err| {
             log.err("failed to parse root uri for workspace {s}: {!}", .{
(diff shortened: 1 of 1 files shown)


src/Server.zig
```diff
@@ -969,8 +974,45 @@ pub fn updateConfiguration(
...
+        inline for (&.{
+            "zig_exe_path",
+            "builtin_path",
+            "build_runner_path",
+            "zig_lib_path",
+            "global_cache_path",
+        }) |fieldName| {
+            const field: *?[]const u8 = &@field(new_config, fieldName);
+            if (field.*) |maybe_relative| resolve: {
+                if (std.fs.path.isAbsolute(maybe_relative)) break :resolve;
```
Techatrix, line 998: A config option that is an empty string should also be ignored.

Techatrix (changes requested):
If there is a single workspace folder, the `scopeURI` field should be set to it when requesting configuration items [here](https://github.com/zigtools/zls/blob/a25ea3d7bd7dc339a1b672a68e2d8e7cdf171cad/src/Server.zig#L694).

Resolving relative paths should be done in the [handleConfiguration](https://github.com/zigtools/zls/blob/a25ea3d7bd7dc339a1b672a68e2d8e7cdf171cad/src/Server.zig#L714) function instead of `updateConfiguration`. That is the only place where we receive config options where relative paths would be allowed.

techatrix:
In case you missed them, these [comments](https://github.com/zigtools/zls/pull/2331#pullrequestreview-2876877037) have not been addressed yet.

dotcarmen:
> these comments have not been addressed yet.

addressed now!

src/Server.zig
```diff
@@ -732,6 +739,40 @@ fn handleConfiguration(server: *Server, json: std.json.Value) error{OutOfMemory}
...
+        const uri = std.Uri.parse(server.workspaces.items[0].uri) catch |err| {
+            log.err("failed to parse root uri for workspace {s}: {!}", .{
+                server.workspaces.items[0].uri, err,
+            });
+            break :dir null;
+        };
+        break :dir try uri.path.toRawMaybeAlloc(arena);
+    } else null;
+
+    check_relative: inline for (&.{
```
Techatrix, line 752: This loop does nothing because `new_config` has not been initialized yet. This loop should come after the `inline for (fields, result) |field, json_value| {` loop further below which initializes the values in `new_config`.

Techatrix (changes requested):

PR #2334, opened 2025-05-25
build: Add `s390x-linux` to release targets.

alexrp:
Zig tarballs for this are produced starting from 0.14.1, but the actual support was there starting from 0.14.0.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -35,6 +35,7 @@ const release_targets = [_]std.Target.Query{
     .{ .cpu_arch = .loongarch64, .os_tag = .linux },
     .{ .cpu_arch = .powerpc64le, .os_tag = .linux },
     .{ .cpu_arch = .riscv64, .os_tag = .linux },
+    .{ .cpu_arch = .s390x, .os_tag = .linux },
     .{ .cpu_arch = .x86, .os_tag = .linux },
     .{ .cpu_arch = .x86, .os_tag = .windows },
     .{ .cpu_arch = .x86_64, .os_tag = .linux },


Techatrix:
Given the previously added targets, is the plan to build ZLS for the same targets as Zig? In the future, would that include additional architectures like sparc64, mips64, and others? I'm not familiar with these architectures, so I'm curious whether there are users who actually run ZLS directly on such systems. For example, are there people running their LSP locally on an IBM mainframe, rather than connecting remotely?

I would like to avoid unnecessary CI time and storage costs by only providing prebuilt binaries if they actually turn out to be needed by someone. If that's the case, I have no objections to adding them. As a compromise, these additional targets could be only provided for tagged releases of ZLS.

alexrp:
> Given the previously added targets, is the plan to build ZLS for the same targets as Zig?

That's been my approach so far. But if I suggest targets that you think don't make sense, feel free to reject them. ZLS's release targets don't *necessarily* have to match Zig's.

> In the future, would that include additional architectures like sparc64, mips64, and others?

We only add tarballs for targets that there's real demand for. That means `mips64` is likely to happen, but `sparc64`? Not so much. We also omit tarballs for `armeb`, `aarch64_be`, `powerpc64`, `riscv32`, etc because there either doesn't seem to be much demand for them, or they're extremely niche, or they're rapidly on their way out.

> I'm not familiar with these architectures, so I'm curious whether there are users who actually run ZLS directly on such systems. For example, are there people running their LSP locally on an IBM mainframe, rather than connecting remotely?

Doesn't the language server end up running on the remote machine when you use VS Code's remote development stuff? That was my understanding of it, but I could definitely be wrong.

> I would like to avoid unnecessary CI time and storage costs by only providing prebuilt binaries if they actually turn out to be needed by someone. If that's the case, I have no objections to adding them. As a compromise, these additional targets could be only provided for tagged releases of ZLS.

Oh, I had assumed that was already the case. I definitely agree there's no reason to be building the niche targets on every commit. I would consider the niche ones to be `mips64`, `powerpc64le`, `s390x`, and `x86` currently, FWIW.

PR #2338, opened 2025-05-26
Rename `Type` to `Expr`

FnControlOption:
In ZLS, a `Type` represents one of two things:
- a type (`is_type_val == true`) or
- an expression of some type (`is_type_val == false`).

Furthermore, a `Type` may be interned.

For example, the primitive type `u8` is
- a type (`is_type_val == true`) whose
- interned type (`data.ip_index.type`) is `type` and
- interned value (`data.ip_index.index`) is `u8`.

In special cases such as number literals, a non-type expression may be interned with a specific value.

For example, the number `42` is
- an expression (`is_type_val == false`) whose
- interned type (`data.ip_index.type`) is `comptime_int` and
- interned value (`data.ip_index.index`) is `42`.

Thus, a `Type` is really just an expression that may or may not be a type, so the name `Type` isn’t quite accurate.

I propose renaming `Type` to `Expr` (short for `Expression`).

- `u8` and `ArrayList(u8)` are expressions that resolve to types
- A `usize` variable is an expression with a runtime-known value (`data.ip_index.index == null`)
- A number literal is an expression with a comptime-known value (`data.ip_index.index != null`)

Granted, this will be quite an invasive change, and many variables/fields/parameters will have to be renamed.

### DIFF
--- a/src/analysis.zig
+++ b/src/analysis.zig
@@ -29,7 +29,7 @@ gpa: std.mem.Allocator,
 arena: std.mem.Allocator,
 store: *DocumentStore,
 ip: *InternPool,
-resolved_callsites: std.AutoHashMapUnmanaged(Declaration.Param, ?Type) = .empty,
+resolved_callsites: std.AutoHashMapUnmanaged(Declaration.Param, ?Expr) = .empty,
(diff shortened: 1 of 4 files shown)


FnControlOption:
This is starting to look ugly... Will reopen if I find an actual use for this

PR #2339, opened 2025-05-27
Workspace Symbols

SuperAuguste:
Supersedes #1876.


xdBronch:
does this fix #1071 since files are more eagerly analyzed?

techatrix:
> does this fix #1071 since files are more eagerly analyzed?

We still search for references in dependencies instead of dependents which this PR doesn't fix. But eagerly opening files in a workspace is a necessary prerequisite to fix #1071 which has been implemented here.

nektro:
congrats on the merge! :tada:

PR #2342, opened 2025-05-28
Fix position context of enum/decl literal in `if`/`while`/`for`

FnControlOption:
For example:

```zig
var foo = if (bar) .foo else .bar;
//                 ^^^^
```


Techatrix (changes requested):
This needs to handle `for` and `while` loops as well.

FnControlOption:
Hm… Do you have any example code of loops that have only a decl literal as the body? I know it’s possible to have a break statement with a decl literal, but we already support that 🤔

techatrix:
> Do you have any example code of loops that have only a decl literal as the body?

None that is semantically valid but that doesn't matter regarding the position context. The token sequence `if (expr) .` should produce the same position context as `while (expr) .`. This should also prevent nonsensical code completions like this:

```zig
const foo: ?u32 = undefined;
const bar = while (foo) .<request completions here>
```

PR #2349, opened 2025-05-31
Expand conditional types in hover popup

FnControlOption:
Screenshot 2025-05-30 at 10 10 45 PM

Screenshot 2025-05-30 at 9 51 53 PM

### DIFF
--- a/src/analysis.zig
+++ b/src/analysis.zig
@@ -3479,23 +3479,196 @@ pub const Type = struct {
         return null;
     }
 
-    /// Resolves possible types of a type (single for all except either)
-    /// Drops duplicates
+    /// Resolves all possible types by recursively expanding any conditional types.
+    /// TODO: Drop duplicates
     pub fn getAllTypesWithHandles(ty: Type, arena: std.mem.Allocator) error{OutOfMemory}![]const Type {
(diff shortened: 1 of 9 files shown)


Techatrix (changes requested):
```zig
const foo = if (true) 1 else true;

const bar = if (true)
    .{ foo, foo, foo, foo, foo, foo, foo, foo, foo, foo, foo, foo, foo, foo, foo, foo, foo, foo, foo, foo, foo, foo, foo }
else
    .{ foo, foo, foo, foo, foo, foo, foo, foo, foo, foo, foo, foo, foo, foo, foo, foo, foo, foo, foo, foo, foo, foo, foo };
```

Hovering over `bar` will cause ZLS to consume huge amounts of memory. Not sure when it stops but it's at least 10GB.

Techatrix (changes requested):
There shouldn't be a config option for this. We can pick a limit that is high enough to not affect users with real code but low enough to prevent hand crafted examples like mine. There is no point in giving users choices they shouldn't need to make.

For testing, the option in `analysis.zig` can remain with a default value. The specific tests can then override it if needed.

Techatrix (approved):

PR #2350, opened 2025-05-31
Deprecate the 'semantic_tokens' config option in favor of 'augmentsSyntaxTokens'

Techatrix:
The [augmentsSyntaxTokens](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#semanticTokensClientCapabilities) client capability can be used to indicate that the client is going to combine existing syntax tokens with the semantic tokens of the server. This can not only reduce the amount of data that is transferred over the protocol but can also lead to better highlighting as syntax tokens often more accurately distinguish between different language constructs on the syntax level. An example would be keywords where the LSP protocol has a single standard token type compared to syntax highlighting where different keywords are often differentiated between based on their purpose.

The `semantic_tokens` config had been added for this purpose. The `partial` option is used to prevent LSP semantic tokens to override existing syntax highlighting of the editor. The benefit of the client capability is that this decision is automatically made by supported editors instead of relying on user to select the right configuration.

The following editors should be unaffected:

- VS Code: enables `partial` by default
- LSP4IJ (JetBrains): does not enable `augmentsSyntaxTokens`
- Helix: does not use semantic tokens
- Zed: does not use semantic tokens
- Kate: does not enable `augmentsSyntaxTokens`
- Eglot (Emacs): does not use semantic tokens
- lsp-mode (Emacs): does not enable `augmentsSyntaxTokens`

The following editors are affected:

<details>

<summary>Sublime Text</summary>

Requires the `semantic_highlighting` config option to be enabled.

Before:
sublime-text-before
After:
sublime-text-after

The new output more closely matches the result when disabling semantic highlighting:

sublime-text-none

</details>

<details>

<summary>Neovim</summary>

Tested with [tokyonight.nvim](https://github.com/folke/tokyonight.nvim)

Before:
neovim-before

After:
neovim-after

The [zig.vim](https://github.com/ziglang/zig.vim) plugin maps the `var`, `const`, `comptime` and `threadlocal` keywords to `Function` which doesn't seem right. Possible fix: https://github.com/ziglang/zig.vim/pull/110

</details>

<details>

<summary>CoC (neovim)</summary>

Tested with [tokyonight.nvim](https://github.com/folke/tokyonight.nvim)

Before:
coc-before

After:
coc-after

</details>

Vim defines preferred [highlighting groups](https://neovim.io/doc/user/syntax.html#group-name) to share group names between many languages. The conventions that have been established by many syntax highlighting configurations are incompatible with LSP semantic highlighting and will result in poor highlighting when mixed together. Because of this, `augmentsSyntaxTokens` should be disabled in (Neo)vim when not using tree-sitter based syntax highlighting.

### DIFF
--- a/schema.json
+++ b/schema.json
@@ -33,7 +33,7 @@
             "default": []
         },
         "semantic_tokens": {
-            "description": "Set level of semantic tokens. `partial` only includes information that requires semantic analysis.",
+            "description": "Deprecated. The client should set the 'augmentsSyntaxTokens' capability.\n\nSet level of semantic tokens. `partial` only includes information that requires semantic analysis.",
             "type": "string",
             "enum": [
                 "none",
--- a/src/Config.zig
+++ b/src/Config.zig
@@ -23,6 +23,8 @@ enable_build_on_save: ?bool = null,
 /// If the `build.zig` has declared a 'check' step, it will be preferred over the default 'install' step.
 build_on_save_args: []const []const u8 = &.{},
 
+/// Deprecated. The client should set the 'augmentsSyntaxTokens' capability.
+///
 /// Set level of semantic tokens. `partial` only includes information that requires semantic analysis.
 semantic_tokens: enum {
     none,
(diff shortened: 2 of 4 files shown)


angelozerr:
@Techatrix I suggest thzt yiu create an issue on Lsp4ij to take care of this capability and if you could explain me how to implement it. Thanks!

techatrix:
More research is required to figure out the best way forward. Tracked in #2527

PR #2354, opened 2025-06-07
reference fields in anonymous struct initializers

### LINKED ISSUE #1837: Rename doesn't rename struct initialization

xdBronch:
closes #1837 
i was worried about potential performance problems but it seems to be working fine and the same was being done for the `.enum_literal` case anyway
also fixes a bug where quoted field names weren't renamed since it was getting the entire token loc

### DIFF
--- a/src/features/references.zig
+++ b/src/features/references.zig
@@ -166,16 +166,41 @@ const Builder = struct {
                 const struct_init = tree.fullStructInit(&buffer, node).?;
                 for (struct_init.ast.fields) |value_node| { // the node of `value` in `.name = value`
                     const name_token = tree.firstToken(value_node) - 2; // math our way two token indexes back to get the `name`
-                    const name_loc = offsets.tokenToLoc(tree, name_token);
-                    const name = offsets.locToSlice(tree.source, name_loc);
+                    const name = offsets.identifierTokenToNameSlice(tree, name_token);
                     if (!std.mem.eql(u8, name, decl_name)) continue;
 
-                    const lookup = try builder.analyser.lookupSymbolFieldInit(handle, name, node, &.{}) orelse continue;
+                    const nodes = switch (tree.nodeTag(node)) {
+                        .struct_init_dot,
+                        .struct_init_dot_comma,
+                        .struct_init_dot_two,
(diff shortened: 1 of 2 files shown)


src/features/references.zig
```diff
@@ -166,16 +166,36 @@ const Builder = struct {
...
+                        .struct_init_dot_comma,
+                        .struct_init_dot_two,
+                        .struct_init_dot_two_comma,
+                        => try ast.nodesOverlappingIndex(
+                            builder.allocator,
+                            tree,
+                            tree.tokenStart(name_token),
+                        ),
+                        // if this isn't an anonymous struct the type can be determined from the `T{}` directly
+                        else => &.{node},
```
Techatrix, line 183:
```suggestion
                        .struct_init_one,
                        .struct_init_one_comma,
                        .struct_init,
                        .struct_init_comma,
                        => &.{node},
                        else => unreachable,
```

Techatrix (changes requested):

Techatrix (approved):

PR #2356, opened 2025-06-10
Print build on save loop stderr

SuperAuguste:
Helpful for diagnosing build on save issues.

### DIFF
--- a/src/features/diagnostics.zig
+++ b/src/features/diagnostics.zig
@@ -563,11 +563,11 @@ pub const BuildOnSave = struct {
             const header = transport.receiveMessage(null) catch |err| switch (err) {
                 error.EndOfStream => {
                     log.debug("zig build runner process has exited", .{});
-                    return;
+                    break;
                 },
                 else => {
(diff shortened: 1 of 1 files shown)


src/features/diagnostics.zig
```diff
@@ -581,14 +581,22 @@ pub const BuildOnSave = struct {
...
             }
         }
+
+        const stderr = if (child_process.stderr) |stderr|
+            stderr.readToEndAlloc(allocator, 16 * 1024 * 1024) catch ""
+        else
+            "";
+        defer allocator.free(stderr);
+
+        log.debug("build on save stderr:\n{s}", .{stderr});
```
Techatrix, line 599:
```suggestion
        if (stderr.len != 0) {
            log.debug("build on save stderr:\n{s}", .{stderr});
        }
```

Techatrix (changes requested):

techatrix:
I coincidentally applied similar changes in #2400. The difference is that stderr is only printed if `error.EndOfStream` is returned which is probably for the better.

PR #2359, opened 2025-06-14
update tests for zig parser change

xdBronch:
https://github.com/ziglang/zig/pull/24160 oops i made zls worse :see_no_evil:

### DIFF
--- a/tests/lsp_features/semantic_tokens.zig
+++ b/tests/lsp_features/semantic_tokens.zig
@@ -2025,12 +2025,6 @@ test "weird code" {
         .{ "bar", .variable, .{} },
     });
     try testSemanticTokens(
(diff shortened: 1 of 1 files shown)


Techatrix (changes requested):
These tests do not care about the output. They only ensure that no crash occurs. Instead of removing the test, the expected output should be set to whatever the actual output is.

xdBronch:
including no output? this test doesnt give any tokens anymore

techatrix:
> including no output? this test doesnt give any tokens anymore

The presence of the test guarantees that the input doesn't cause a crash. That's the entire point of the test in the first place.

Techatrix (approved):

PR #2368, opened 2025-06-20
fix some breaking changes with build runner and `NativePaths.detect`

leecannon:
https://github.com/ziglang/zig/pull/24205 renamed a field on `std.Build.Step.ConfigHeader`

Fixes https://github.com/zigtools/zls/issues/2366

PR #2369, opened 2025-06-22
Fix native paths https://github.com/ziglang/zig/pull/24193

leecannon:
`NativePaths.detect` was changed to take `std.Target` by pointer in https://github.com/ziglang/zig/commit/917640810e7f3e18daff9e75b5ecefe761a1896c (https://github.com/ziglang/zig/pull/24193)


leecannon:
Change will be made in https://github.com/zigtools/zls/pull/2368

PR #2372, opened 2025-06-24
resolve relative paths from workspace/didChangeConfiguration

dotcarmen:
ok admittedly i made the last changes to #2331 without thorough testing.

when I tried to use the feature from #2331 just now, i found out that it wasn't working with the following `.zed/settings.json` i have for my ziglang/zig checkout:

```jsonc
// see the documentation: https://zed.dev/docs/configuring-zed#settings-files
{
  "lsp": {
    "zls": {
      "settings": {
        "builtin_path": "./src/Builtin.zig",
        "zig_lib_path": "./lib",
        "zig_exe_path": "./build/stage3/bin/zig",
        "build_runner_path": "./lib/compiler/build_runner.zig",
      }
    }
  }
}
```

(ignore the incorrect `builtin_path` pls)

after further debugging, i found 2 issues:
1. zed is sending these settings as part of `workspace/didChangeConfiguration` notifications. Right now zls only correctly resolves relative paths that are a response to `workspace/configuration` server-side requests (in other words, client messages with `i_haz_configuration` id)
2. because #2331 only used `std.fs.path.join` to resolve the relative paths, ZLS seemed to have some double-accounting of the paths. at some point, I had a log message indicating that both `..././src/Builtin.zig` and `.../src/Builtin.zig` were being watched by ZLS. which is redundant and presumably a (minor) waste of memory

this PR accounts for both of these issues:
- logic for resolving relative paths have been moved back to `updateConfiguration` method on Server
  - this only happens when `options.resolve_relative_paths` argument is true
  - this is directly opposed to [this review comment](https://github.com/zigtools/zls/pull/2331#pullrequestreview-2876877037) which was addressed in #2331.
    - the review comment mentions that only `handleConfiguration` should be receiving messages with relative configuration paths. As noted above, this is not true - it should also handle relative paths in `workspace/didChangeConfiguration` notifications, which is handled by `didChangeConfigurationHandler`
    - I think it's better to handle this behavior in `updateConfiguration` since `handleConfiguration` and `didChangeConfigurationHandler` have different handling of the message
    - later on, when zls receives the `i_haz_configuration` response from the client, it seems Zed is not re-sending the updated relative paths. while the `workspace/configuration` model seems to be the more modern feature, it seems Zed still expects `workspace/didChangeConfiguration` to handle the initial values from `.zed/settings.json`
- instead of using `std.fs.path.join` to resolve relative paths, use `std.fs.path.resolve`.

### DIFF
--- a/src/Server.zig
+++ b/src/Server.zig
@@ -754,42 +754,9 @@ fn handleConfiguration(server: *Server, json: std.json.Value) error{OutOfMemory}
         }
     }
 
-    const maybe_root_dir: ?[]const u8 = if (server.workspaces.items.len == 1) dir: {
-        const uri = std.Uri.parse(server.workspaces.items[0].uri) catch |err| {
-            log.err("failed to parse root uri for workspace {s}: {!}", .{
-                server.workspaces.items[0].uri, err,
-            });
-            break :dir null;
-        };
-        break :dir try uri.path.toRawMaybeAlloc(arena);
-    } else null;
-
-    check_relative: inline for (&.{
-        "zig_exe_path",
-        "builtin_path",
-        "build_runner_path",
-        "zig_lib_path",
-        "global_cache_path",
-    }) |fieldName| {
-        const field: *?[]const u8 = &@field(new_config, fieldName);
-        if (field.*) |maybe_relative| resolve: {
-            if (maybe_relative.len == 0) break :resolve;
-            if (std.fs.path.isAbsolute(maybe_relative)) break :resolve;
-
-            const root_dir = maybe_root_dir orelse {
-                log.err("relative path only supported for {s} with exactly one workspace", .{fieldName});
-                break :check_relative;
-            };
-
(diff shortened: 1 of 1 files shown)


dotcarmen:
@Techatrix let me know if there's another way this should be handled. sorry for not testing again before merging, last few weeks were busy and i was focused on other projects 😬

techatrix:
According to Zed's client capabilities, they claim to support `workspace/configuration` so this workaround shouldn't be necessary in theory. But they also send configuration value with `workspace/didChangeConfiguration` and ZLS does not handle this well. I will need to investigate this further to figure out what to do here. I will keep you updated once I have something more concrete to say.

techatrix:
With the changes in #2416, ZLS will now properly use `workspace/configuration` with Zed and various other editors instead of relying on the outdated `workspace/didChangeConfiguration` request to receive config options. This means that the changes in this PR should no longer be needed to get this working. Feel free to let me know if #2416 works well for you.

PR #2375, opened 2025-06-25
Improvements for switch captures

FnControlOption:
## Fix resolved type of tag capture in inline switch case

```zig
const Enum = enum {
    foo,
    bar,
    baz,
    qux,
};

const TaggedUnion = union(Enum) {
    foo: i32,
    bar: bool,
    baz: f64,
    qux: void,
};

const some_tagged_union: TaggedUnion = .{ .baz = 3.14 };

const switch_tagged_union_inline = switch (some_tagged_union) {
    inline .foo => |a, b| .{ a, b },
    //                 ^ Enum
    inline .bar => |a, b| .{ a, b },
    //                 ^ Enum
    inline else => |a, b| .{ a, b },
    //                 ^ Enum
};
```

## Fix resolved type of switch capture for non-enum/union types

```zig
const some_u8: u8 = 'e';

const switch_u8 = switch (some_u8) {
    'x' => |a| a,
    //      ^ u8
    'y' => |a| a,
    //      ^ u8
    else => |a| a,
    //       ^ u8
};
```

## Resolve type of else capture in tagged union switch

```zig
const switch_tagged_union = switch (some_tagged_union) {
    .foo => |a| a,
    .bar => |a| a,
    else => |a| a,
    //       ^ TaggedUnion
};
```

### DIFF
--- a/src/DocumentScope.zig
+++ b/src/DocumentScope.zig
@@ -114,8 +114,11 @@ pub const Declaration = union(enum) {
         condition: Ast.Node.OptionalIndex,
     },
     assign_destructure: AssignDestructure,
-    // a switch case capture
+    /// - `switch (condition) { .case => |value| {} }`
     switch_payload: Switch,
+    /// - `switch (condition) { inline .case => |_, tag| {} }`
+    /// - `switch (condition) { inline else => |_, tag| {} }`
+    switch_inline_tag_payload: Switch,
     label: struct {
         identifier: Ast.TokenIndex,
         block: Ast.Node.Index,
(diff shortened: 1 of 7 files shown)


src/DocumentScope.zig
```diff
@@ -156,6 +156,7 @@ pub const Declaration = union(enum) {
     };
 
     pub const Switch = struct {
+        identifier: Ast.TokenIndex,
```
Techatrix, line 159: This unnecessarily increases the size of each DocumentScope Declaration from 8 to 12 bytes for the untagged union.

Having two different fields in `Declaration` should work for this:

```zig
/// - `switch (condition) { .case => |value| {} }`
switch_payload: Switch,
/// - `switch (condition) { inline .case => |_, tag| {} }`
/// - `switch (condition) { inline else => |_, tag| {} }`
switch_inline_tag_payload: Switch,
```

Techatrix (changes requested):
The test cases should ideally also cover the `_` prong on non-exhaustive enums.
```zig
const E = enum(u8) {
    foo,
    bar,
    _,
};
const some_enum: E = .foo;
const what = switch (some_enum) {
    .foo, .bar => {},
    _ => |tag| {},
};
```

Techatrix (approved):

PR #2376, opened 2025-06-26
Resolve type of string/vector element

### LINKED ISSUE #1973: capturing u8 from a string in a for loop has unknown type
### Zig Version

0.14.0-dev.367+a57479afc

0.14.0-dev.54+adf3b83

nvim 0.10.0 with nvim-lspconfig

```zig
pub fn main() void {
    const string = "Hello, world!";
    for (string) |byte| {
        _ = byte;
    }

    for ("hello") |byte| {
        _ = byte;
    }
}
```
image
image

### Expected Behavior

Type of the capture should be `u8`.

### Relevant log output

_No response_

FnControlOption:
Resolves #1973


Techatrix (approved):

techatrix:
Reverted because the following code snippet will cause infinite recursion:

```zig
const foo: type = undefined;
const bar = foo[5];
```

PR #2381, opened 2025-06-29
fix: `NativePaths.detect` expects a `Target` not a `*const Target`

robbielyman:
this commit also fixes an apparent typo in the required zig version? at least, the currently available-for-download version of zig master is 0.15.0-dev.847, but this repo claims a minimum required version of 0.15.0-dev.864.


leecannon:
The latest version of zig master available is actually `0.15.0-dev.877+0adcfd60f` https://ziglang.org/download/index.json.

There is an issue with the website not updating its download page https://github.com/ziglang/www.ziglang.org/issues/480.

In the latest zig master the target does need to be passed as a pointer.

PR #2384, opened 2025-06-30
fix build config exchange in invalidateBuildFileWorker

### LINKED ISSUE #2383: build on save broken
### Zig Version

0.15.0-dev.885+e83776595

### ZLS Version

0.15.0-dev.250+3e1d1a11

vscode 1.101.2 vscode-zig 0.6.10

ZLS is broken for https://github.com/CascadeOS/CascadeOS after https://github.com/zigtools/zls/pull/2377.

There are intermittent seg faults on ZLS starting and if it does start without crash running 'Stop Language Server' will have a seg fault.

When ZLS starts without crashing packages are not resolved. Running the build runner manually gets all the expected packages.

This [log](https://github.com/user-attachments/files/20986847/zls.log.txt) (had to attach as log field does not accept a log this big) has many crashes on startup at `DocumentStore.zig:1976` but then at the end successfully started (without packages) then the final seg fault is after a call to 'Stop Language Server'.

### Expected Behavior

ZLS starts and stops without seg fault and resolves packages.

### Relevant log output

```shell

```

Techatrix:
fixes #2383

### DIFF
--- a/src/DocumentStore.zig
+++ b/src/DocumentStore.zig
@@ -1014,11 +1014,12 @@ fn invalidateBuildFileWorker(self: *DocumentStore, build_file: *BuildFile) void 
         switch (build_file.impl.build_runner_state) {
             .idle => unreachable,
             .running => {
-                build_file.impl.build_runner_state = .idle;
+                var old_config = build_file.impl.config;
                 build_file.impl.config = build_config;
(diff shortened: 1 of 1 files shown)


leecannon (approved):
:+1:

PR #2387, opened 2025-07-02
add `Unnecessary` tag to unused var diagnostics

The-Briel-Deal:
Problem: Currently unused var and const declarations produce a diagnostic error without `lsp.types.DiagnosticTag.Unnecessary`. This means that editors can't visually display that a variable is unused by dimming the highlight on the variable.

Solution: When `ast-check` produces a "unused local constant" or "unused local variable" diagnostic, we should add the Unnecessary tag to the LSP diagnostic.


Techatrix (changes requested):

src/DiagnosticsCollection.zig
```diff
@@ -354,11 +354,23 @@ fn convertErrorBundleToLSPDiangostics(
             break :blk lsp_notes;
         };
 
+        var tags: std.ArrayListUnmanaged(lsp.types.DiagnosticTag) = .empty;
+
+        const diag_msg = eb.nullTerminatedString(err.msg);
+        const unnecessary_prefix = "unused";
+
+        if (diag_msg.len >= unnecessary_prefix.len and
+            std.mem.eql(u8, diag_msg[0 .. unnecessary_prefix.len], unnecessary_prefix))
```
Techatrix, line 363:
```suggestion

        if (std.mem.startsWith(u8, diag_msg, "unused "))
```

Techatrix (changes requested):

The-Briel-Deal:
It looks to me like CI was failing on the last run because I didn't have the version tags on my fork. All tests run from `zig build test` pass locally.

techatrix:
There has been a recent breaking change in Zig which has caused the CI to fail. This will be addressed soon by #2394.

The-Briel-Deal:
> There has been a recent breaking change in Zig which has caused the CI to fail. This will be addressed soon by #2394.

Ok thats good to hear, I just wanted to make sure it wasn't a problem on my side.

Techatrix (approved):

PR #2390, opened 2025-07-07
Fix `nodeChildrenRecursiveAlloc` and `nodeChildrenAlloc`

KurtWagner:
zls is amazing, thanks for developing it. I noticed this while developing zlinter which uses zls for a bunch of code analysis

- Since 0.15 index is an enum (so .root == 0 or `@enumFromInt(0)`).
- Both methods were also using unmanaged structures so should be passing the allocator to deinit and toOwned slice
- Added a basic test for both that replicated problem being fixed

### DIFF
--- a/src/ast.zig
+++ b/src/ast.zig
@@ -1520,15 +1520,38 @@ pub fn nodeChildrenAlloc(allocator: std.mem.Allocator, tree: Ast, node: Ast.Node
         children: *std.ArrayListUnmanaged(Ast.Node.Index),
         fn callback(self: @This(), ast: Ast, child_node: Ast.Node.Index) error{OutOfMemory}!void {
             _ = ast;
-            if (child_node == 0) return;
+            if (child_node == .root) return;
             try self.children.append(self.allocator, child_node);
         }
     };
 
     var children: std.ArrayListUnmanaged(Ast.Node.Index) = .empty;
-    errdefer children.deinit();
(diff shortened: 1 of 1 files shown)


src/ast.zig
```diff
@@ -1520,15 +1520,38 @@ pub fn nodeChildrenAlloc(allocator: std.mem.Allocator, tree: Ast, node: Ast.Node
         children: *std.ArrayListUnmanaged(Ast.Node.Index),
         fn callback(self: @This(), ast: Ast, child_node: Ast.Node.Index) error{OutOfMemory}!void {
             _ = ast;
-            if (child_node == 0) return;
+            if (child_node == .root) return;
```
Techatrix, line 1523: The old meaning of `0` isn't always the same as `.root`. Previously, `0` could mean either `.root` or a "null" node. These two possibilities are now differentiated between through the type system with `Ast.Node.OptionalIndex`. It is impossible that the child node is `.root` so this check is no longer necessary:

```suggestion
```

src/ast.zig
```diff
@@ -1540,17 +1563,42 @@ pub fn nodeChildrenRecursiveAlloc(allocator: std.mem.Allocator, tree: Ast, node:
...
```
Techatrix, line 1566:
```suggestion
```

Techatrix (changes requested):

Issue #2100, opened 2024-11-30
make organize imports more flexible about import placement


Techatrix:
The current implementation of `source.organizeImports` (#2051) always places the imports at the top of the file. Sometimes you find code that has all imports at the bottom of the file or some imports between code. It would be nice if ZLS could this.

I propose either of these two solutions:
- add a special case where ZLS will place all imports at the bottom of the file (**no config option**)
- never move imports to a different location, only sort imports that are next to each other i.e. not separated by an empty line

Srekel:
+1 to keeping imports together.

Here is an example where it's moved lines from the second "paragraph" of consts into the first paragraph of imports. I like to group imports by having a space inbetween.

![image](https://github.com/user-attachments/assets/2e101e3d-181a-4a71-8b4d-612e643dd667)

nihklas:
Why not making it configurable? 
I think it would help enforce code style beyond what `zig fmt` does. I don't know how common it is to have imports both at the top and the bottom in the same codebase.

PR #2391, opened 2025-07-07
add support for organizing imports at the bottom of file

tristanpemble:
this adds an `import_organization` config that allows `auto`, `top`, and `bottom`. it defaults to `auto`.

in auto mode, if the last decl of the file is an import, it uses `bottom`, otherwise it uses `top`.

I struggled with whitespace/newlines, as you can see in the test cases. I'm putting the PR up in the hope that either this is acceptable, since `zig fmt` will remove the duplicate newlines, or that if it is not acceptable, someone can assist me in fixing it.


tristanpemble:
in testing this in my own codebase I am running into some edge cases, so I want to iron those out.

tristanpemble:
this is now ready for review.

src/features/code_actions.zig
```diff
@@ -587,6 +589,52 @@ fn handleVariableNeverMutated(builder: *Builder, loc: offsets.Loc) !void {
...
+    // non-import declarations between them.
+    //
+    // If there aren't, the file is only imports, so we can choose "top" to avoid unnecessary newlines.
+    // If there are; there being an import at the bottom is a strong signal that that is what they want.
+    if (!has_gaps) {
+        return .top;
+    } else {
+        return .bottom;
+    }
+}
```
Techatrix, line 636: This function is more complicated than it needs to be:

- `imports` are already sorted.
- The usage `@intFromEnum` is unnecessary because you are comparing `Ast.Node.Index` which already works with `==` and `!=`.

```suggestion
fn analyzeImportPlacement(tree: Ast, imports: []const ImportDecl) ImportPlacement {
    const root_decls = tree.rootDecls();

    if (root_decls.len == 0 or imports.len == 0) return .top;

    const first_import = imports[0];
    const last_import = imports[imports.len - 1];

    const first_decl = root_decls[0];
    const last_decl = root_decls[root_decls.len - 1];

    const starts_with_import = first_decl == first_import.var_decl;
    const ends_with_import = last_decl == last_import.var_decl;

    if (starts_with_import and ends_with_import) {
        // In the ambiguous case, there is an import at both the top and bottom. The deciding factor is 
        // if there are other non-import declarations between them.
        //
        // If there aren't, the file is only imports, so we can choose "top" to avoid unnecessary newlines.
        // If there are; there being an import at the bottom is a strong signal that that is what they want.
        const has_gaps = root_decls.len != imports.len;
        return if (has_gaps) .bottom else .top;
    }
    return if (!starts_with_import and ends_with_import) .bottom else .top;
}
```

Techatrix (changes requested):
As mentioned in #2100, this feature should not have a config option. Remove it.

> I struggled with whitespace/newlines, as you can see in the test cases. I'm putting the PR up in the hope that either this is acceptable, since `zig fmt` will remove the duplicate newlines, or that if it is not acceptable, someone can assist me in fixing it.

Relying on `zig fmt` is acceptable. This could always be improved in the future.

tristanpemble:
Thank you for the feedback. I've removed the config option, and simplified `analyzeImportPlacement` as you've suggested.

The key was that I somehow completely missed the [sort call directly above](https://github.com/zigtools/zls/blob/af2acc31059f0ba2b1c81fc71a7e5569ef5500dd/src/features/code_actions.zig#L655) 🤦 . These checks needed to on the original index, not post-sort.

Anyways, thanks to your input, it's much cleaner now.

Techatrix (approved):

PR #2412, opened 2025-07-19
start reference search in the current file rather than the decl file

xdBronch:
in practice this means that documentHighlight will properly work for things declared in other files :tada:. before we were searching the file where the symbol was declared and then just [throwing it all away](https://github.com/zigtools/zls/blob/45b855f7ec3dccea3c9a95df0b68e27dab842ae4/src/features/references.zig#L735), incredible!
it appears we've been doing this for... [literally ever](https://github.com/zigtools/zls/commit/8bc608764602d4f9591a827c349b447233d16f88)


xdBronch:
unsure what to do about the windows failure, trying to run it locally with wine im getting panics in the build system and `error.Unexpected`s...

PR #2413, opened 2025-07-19
update zig

antlilja:
Depends on https://github.com/zigtools/lsp-kit/pull/23

### DIFF
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -21,8 +21,8 @@
             .hash = "diffz-0.0.1-G2tlIQrOAQCfH15jdyaLyrMgV8eGPouFhkCeYFTmJaLk",
         },
         .lsp_kit = .{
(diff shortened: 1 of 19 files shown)


techatrix:
There are some issues with this PR that would need to be addressed to be merged:

- do not let `error.WriteFailed` traverse through the entire program. The error should be catched wherever the underlying writer has been created. 
- Do not use `adaptToNewApi` when you can use the new API directly
- semantic tokens for the new asm clobbers are missing
- Calling `tree.extraData(extra_index, Node.Asm)` on a `.asm_legacy` node is incorrect ([here](https://github.com/zigtools/zls/blob/44f342988b65615a2e0451056756e5226e66f44f/src/ast.zig#L693C27-L693C64))
- `tree.asmLegacy(node)` should be `ast.asmLegacy(tree, node)` ([here](https://github.com/zigtools/zls/blob/44f342988b65615a2e0451056756e5226e66f44f/src/ast.zig#L1475))

I can save you the effort of addressing these issues as I had already made these changes locally. See  #2415

PR #2417, opened 2025-07-20
Iterate child `fn_proto_*` under `fn_decl`

KurtWagner:
`fn_decl` contains a `fn_proto_*` so should be iterated above the proto when walking children, as they're not quite the same thing.

It's possible that this is working as intended in the context of ZLS.

I noticed this when comparing total nodes to nodes walked using the iterate method.

### DIFF
--- a/src/ast.zig
+++ b/src/ast.zig
@@ -1126,6 +1126,44 @@ pub fn iterateChildren(
     }
 }
 
+test "iterateChildren - fn_proto_* inside of fn_proto" {
+    const allocator = std.testing.allocator;
+
+    var tree = try std.zig.Ast.parse(
+        allocator,
+        \\pub fn nextAge(age: u32) u32 {
+        \\  return age + 1;
+        \\}
+    ,
+        .zig,
+    );
+    defer tree.deinit(allocator);
+
+    var children_tags = std.ArrayList(Ast.Node.Tag).init(allocator);
+    defer children_tags.deinit();
+
+    const Context = struct {
+        accumulator: *std.ArrayList(Ast.Node.Tag),
+        fn callback(self: @This(), ast: Ast, child_node: Ast.Node.Index) !void {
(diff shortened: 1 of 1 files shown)


Techatrix (changes requested):
The current approach does work correctly in ZLS. The idea is to make it easier to switch on the `.fn_decl` and `.fn_proto*` tags together, use `tree.fullFnProto` to get the function prototype and then handle the function body separately. However this behaviour is non obvious so I can understand changing this to handle `.fn_decl` separately.

There are two locations where we depend on the previous behaviour which needs to be adjusted:

DocumentScope.zig:
- https://github.com/zigtools/zls/blob/957eb87c355cf1917be23a709eeedee301d78d97/src/DocumentScope.zig#L629-L630
- https://github.com/zigtools/zls/blob/957eb87c355cf1917be23a709eeedee301d78d97/src/DocumentScope.zig#L956-L959

semantic_tokens.zig:
- https://github.com/zigtools/zls/blob/957eb87c355cf1917be23a709eeedee301d78d97/src/features/semantic_tokens.zig#L390
- https://github.com/zigtools/zls/blob/957eb87c355cf1917be23a709eeedee301d78d97/src/features/semantic_tokens.zig#L461-L463

KurtWagner:
Thanks for pointing these out. For document scope, are we 

1. wanting to keep the same scope output (i.e., `function` scope is `fn_decl` or `fn_proto_*`), thus the change is more about being consistent in how ZLS walks the tree by first looking at `fn_decl`; or are we

2. wanting a new scope that encapsulates a `fn_proto` and `fn_decl` separately (e.g., a `fn_decl` would scope like `function -> function_proto -> block` and `fn_proto` would scope as `function_proto`)

I'm not sure (2) adds a lot of value in this context so would lean towards the former as lets friction but maybe I'm missing something?

PR #2438, opened 2025-08-07
Use ArrayListUnmanaged instead of BoundedArray

Joey152:
Fix the removal of BoundedArray that happened in zig master here: https://github.com/ziglang/zig/commit/196e36bbb27b0f0ebd7cd7a866b85f477b3662fb#diff-cbd8cbbc17583cb9ea5cc0f711ce0ad447b446e62ea5ddbe29274696dce89e4f

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -6,7 +6,7 @@ const builtin = @import("builtin");
 const zls_version: std.SemanticVersion = .{ .major = 0, .minor = 15, .patch = 0, .pre = "dev" };
 
 /// Specify the minimum Zig version that is required to compile and test ZLS:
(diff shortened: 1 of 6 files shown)


mochalins (changes requested):

mochalins:
@Joey152 This must also update the `build.zig` accordingly; namely, the `minimum_build_zig_version` (along with the PR name that caused the breaking change).

The same change should be reflected in `build.zig.zon` as well.

leecannon:
Thanks for the PR @Joey152 

I have applied @mochalins suggestions and fixed up a test that was missed.

leecannon (approved):

PR #2457, opened 2025-08-22
Use silversquirl/zig-flake

silversquirl:
Shameless plug ofc, but my flake has a number of advantages over zig-overlay :)

- very frequent updates - each zig version should be available in the flake within half an hour of being published to ziglang.org
- fetches from community mirrors - this increases availability when ziglang.org has downtime, and also avoids DoSing ZSF
- includes a setup hook and a `makePackage` helper to make packaging extremely easy
- includes a dependency fetcher that doesn't need a `deps.nix` (just update a single `depsHash` instead)
- zero dependencies (no pointless `flake-utils` ^^')
- actively maintained - zig-overlay has open PRs that haven't even been given feedback in over a year


silversquirl:
also, a note on versioning: it would be very nice if zls chose a specific nightly version to use (eg. `zig_0_16_0_dev_9`) rather than simply following the `nightly` alias. this means other flakes that depend on zls can have it follow their copy of `zig-flake`, and still run `nix flake update` without any fear of zls breaking due to the new version.

This would be much appreciated even if you choose to reject this PR and stick with zig-overlay! ^-^

techatrix:
Took me some time but I managed to convince myself to agree with switching to silversquirl/zig-flake.

> * fetches from community mirrors - this increases availability when ziglang.org has downtime, and also avoids DoSing ZSF

Some time after this PR has been opened, zig-overlay did start [fetching from community mirrors](https://github.com/mitchellh/zig-overlay/pull/79/files) but they used `fetchurl` with multiple urls which are [fetched in order](https://github.com/NixOS/nixpkgs/blob/11829932852b028b2ef13120aadc0f3a8379ba49/doc/build-helpers/fetchers.chapter.md?plain=1#L197) so it just always tries `https://pkg.machengine.org/zig`. Your flake appears to fetch from mirrors at random which is nice. I do have a possible issue about how to deal with [community-mirrors.txt](https://ziglang.org/download/community-mirrors.txt) which can be found in https://github.com/silversquirl/zig-flake/issues/2. Would be nice to get that resolved before switching.

> * includes a setup hook and a `makePackage` helper to make packaging extremely easy
> 
> * includes a dependency fetcher that doesn't need a `deps.nix` (just update a single `depsHash` instead)

That sure looks convenient. But I'd like to minimize the amount of things that have been offloaded to a different flake so packaging ZLS and fetching it's dependencies should remain as is. Just with zig-overlay, a different flake can provide us with Zig but that's it.

I have implemented an updated version of the ZLS flake here:
https://github.com/zigtools/zls/tree/techatrix/nix

I also pinned the Zig version just like you suggested and made some other improvements. Let me know if think that's an acceptable solution or you have further suggestions.

Techatrix:
I switched to `silversquirl/zig-flake` in 75e0b965e4ecbaab08bd72630f7e59ef77654923. As suggested, the exact Zig version has also been pinned in e59e84150d1f28606920e3718e9082b5505a5310.

PR #2468, opened 2025-08-27
Remove the concept of top level snippets

Hejsil:
The limit of being at the beginning of the line or at the first character of a line is too limiting. In the editor I use (`kakoune`), snippets are part of the auto completion menu that pops up as I type. If I type `print`, the `print` snippet can be selected, but not if I type `std` because of this limitation.

`print`
image

`std`
image

I could not find a reason for why it works like this in any commits or the PR that implemented it, so I suggest removing it and just always provide these snippets.

### DIFF
--- a/src/features/completions.zig
+++ b/src/features/completions.zig
@@ -859,15 +859,6 @@ pub fn completionAtIndex(
     };
     const source = handle.tree.source;
 
-    // Provide `top_level_decl_data` only if `offsets.lineSliceUntilIndex(handle.tree.source, source_index).len` is
-    // 0 => Empty new line, manually triggered
-    // 1 => This is the very first char on a given line
-    const at_line_start = offsets.lineSliceUntilIndex(source, source_index).len < 2;
-    if (at_line_start) {
-        try populateSnippedCompletions(&builder, &snipped_data.top_level_decl_data);
-        return .{ .isIncomplete = false, .items = builder.completions.items };
-    }
-
     const pos_context = try Analyser.getPositionContext(arena, handle.tree, source_index, false);
 
     switch (pos_context) {
--- a/src/snippets.zig
+++ b/src/snippets.zig
@@ -8,29 +8,6 @@ pub const Snipped = struct {
     text: ?[]const u8 = null,
 };
 
-pub const top_level_decl_data = [_]Snipped{
-    .{ .label = "std", .kind = .Snippet, .text = "const std = @import(\"std\");" },
-    .{ .label = "root", .kind = .Snippet, .text = "const root = @import(\"root\");" },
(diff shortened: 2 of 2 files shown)


techatrix:
Does this patch work for you?

```diff
diff --git a/src/features/completions.zig b/src/features/completions.zig
index a7408529..731779e6 100644
--- a/src/features/completions.zig
+++ b/src/features/completions.zig
@@ -872,11 +872,9 @@ pub fn completionAtIndex(
     };
     const source = handle.tree.source;
 
-    // Provide `top_level_decl_data` only if `offsets.lineSliceUntilIndex(handle.tree.source, source_index).len` is
-    // 0 => Empty new line, manually triggered
-    // 1 => This is the very first char on a given line
-    const at_line_start = offsets.lineSliceUntilIndex(source, source_index).len < 2;
-    if (at_line_start) {
+    const line_until_index = offsets.lineSliceUntilIndex(source, source_index);
+
+    if (line_until_index.len == 0 or std.zig.isValidId(line_until_index)) {
         try populateSnippedCompletions(&builder, &snipped_data.top_level_decl_data);
         return .{ .isIncomplete = false, .items = builder.completions.items };
     }
```

Some of the top level snippets collide with the normal snippets so they can't just be copied over without further care.

Hejsil:
> Does this patch work for you?
> 
> ```diff
> diff --git a/src/features/completions.zig b/src/features/completions.zig
> index a7408529..731779e6 100644
> --- a/src/features/completions.zig
> +++ b/src/features/completions.zig
> @@ -872,11 +872,9 @@ pub fn completionAtIndex(
>      };
>      const source = handle.tree.source;
>  
> -    // Provide `top_level_decl_data` only if `offsets.lineSliceUntilIndex(handle.tree.source, source_index).len` is
> -    // 0 => Empty new line, manually triggered
> -    // 1 => This is the very first char on a given line
> -    const at_line_start = offsets.lineSliceUntilIndex(source, source_index).len < 2;
> -    if (at_line_start) {
> +    const line_until_index = offsets.lineSliceUntilIndex(source, source_index);
> +
> +    if (line_until_index.len == 0 or std.zig.isValidId(line_until_index)) {
>          try populateSnippedCompletions(&builder, &snipped_data.top_level_decl_data);
>          return .{ .isIncomplete = false, .items = builder.completions.items };
>      }
> ```

Yep, this seems to work, though I do wonder why `panic` is a completion option here
image

techatrix:
> I do wonder why `panic` is a completion option here

It's a snippet for the panic handler:
https://github.com/zigtools/zls/blob/173ad110c180145b048647a9fbf0e37b3b553138/src/snippets.zig#L26-L30

techatrix:
Closing in favor of #2482

PR #2469, opened 2025-08-28
Add flake support for some older versions of zls

Ratakor:
I think this is useful for when nixpkgs doesn't have up to date zls or when a project uses an older version of zig.
Currently there is 3 versions: master, 0.15.0 and 0.14.0 but I can add more if needed.

Depends on https://github.com/zigtools/zls/pull/2457.
Related issue https://github.com/NixOS/nixpkgs/issues/392896.


Ratakor:
Made two typos but now it works as intended, see [here](https://github.com/Ratakor/zfs-restore/blob/master/flake.nix) for an example.

techatrix:
Providing older ZLS versions is out of scope for this flake. Consider pinning your ZLS input flake.

PR #2510, opened 2025-11-18
more accurately model how build system modules work

Techatrix:
ZLS had an inaccurate view of how modules work in the build system. It would previously merge every module in the entire build graph into one single module which can easily lead to incorrect results in any non trivial project that contains multiple modules. This PR will keep track of each individual module so that analysis can differentiate between them just like how the compiler does it.

Some examples of what has been improved:
- `@import("root")` now actually figures out in which module a given file is in. 
- auto complete for `@import` will only show the modules that have been directly added to the module that the current file belongs to
- `@import("some_name")` will resolve the correct module if there are multiple modules that have imported different modules with `"some_name"` as their name.

One issue this PR doesn't fully address is when there are multiple modules with the same root source file that have conflicting definitions. This is only achievable using multiple compilation units because Zig fortunately disallows sharing a root source file within a single compilation unit.

Related PR #2308

### DIFF
--- a/src/DocumentStore.zig
+++ b/src/DocumentStore.zig
@@ -1436,7 +1436,7 @@ fn createAndStoreDocument(
     };
     errdefer new_handle.deinit();
 
-    if (supports_build_system and isBuildFile(uri) and !isInStd(uri)) {
+    if (supports_build_system and lsp_synced and isBuildFile(uri) and !isInStd(uri)) {
         if (self.getBuildFile(uri)) |build_file| {
             self.invalidateBuildFile(build_file.uri);
         } else {


dotcarmen:
> `@import("root")` now actually figures out in which module a given file is in.

fyi - it seems that can't really be relied upon

https://gist.github.com/dotcarmen/7f186613f1f839427a839c33306af9dd <- the output is:
```
hello, world!
exe says hello
```

what happens when a module is imported by multiple roots? i think it's best to not resolve `@import("root")` to _anything_ to avoid any footguns

techatrix:
> fyi - it seems that can't really be relied upon
> 
> https://gist.github.com/dotcarmen/7f186613f1f839427a839c33306af9dd <- the output is:

I turns out that I have misunderstood the semantic of `@import("root")` as I assumed it would be the root source file a given file is inside but it is root source file of the compilation unit(s) a given file is in. I have updated the PR to handle this which should fix the issue in the example you provided.

> what happens when a module is imported by multiple roots?

Resolve it to all of them. ZLS has something called "either types" that can be used to represent a type that can be one of a given set of types but it's unknown which one it will be.

So `const root = @import("root")` would be internally represented as:
```zig
const root = switch (unknown) {
    0 => @import("main.zig"),
    1 => @import("main2.zig"),
};
```
If `main.zig` and `main2.zig` if both root modules that import the file we are in.

complete example: https://gist.github.com/Techatrix/80b625e5965ca6177ff4360713ffded3

dotcarmen:
i've been running this build pretty successfully, however, as of `0.16.0-dev.1484+d0ba6642b`, i'm getting an error about `LazyPath.getPath` being called during the configure phase

this only occurred after i added `b.addTranslateC` step

dotcarmen:
per https://github.com/zigtools/zls/pull/2564#issuecomment-3738052084:

> I have a similar change in https://github.com/zigtools/zls/pull/2510 (https://github.com/zigtools/zls/commit/c8098a4783a1dc9ae9fc7e6b7dc9076202f907f9) which has not been merged yet, and has gone out of date again.

would it be helpful if i gave you merge commits @Techatrix? I'm not sure what's blocking merging of this PR, but I've been making manual merge commits against master to have the nice QoL improvements from this PR in my daily usage

if there's anything needed to help push this PR over the line, i'd be happy to help.

techatrix:
Thanks for the offer to help. I already have locally rebased this PR which should get it ready to be merged.

I wanted to wait until I could investigate some issues related to [ziglang/translate-c](https://codeberg.org/ziglang/translate-c) which is likely what you mentioned here. The issues were already present before and are unrelated to this PR so they don't have to block it. They are tracked in #2572 and #2573. 

The change in 4e59770c29bc433c1c6e8dc75c076051917bffa4 should at least make the errors about ziglang/translate-c not appear unless you manually navigate to the package in the zig cache.

PR #2516, opened 2025-11-26
update to Zig `0.16.0-dev.1458+755a3d957`

Techatrix:


src/configuration.zig
```diff
@@ -694,14 +688,13 @@ pub const DidConfigChange = blk: {
             .alignment = 0,
```
xdBronch: not too much going on with these old variables anymore

PR #2522, opened 2025-11-29
Fix hover/definition for pointer decl literals

vivekstills:
- Dereference expected decl-literal types when resolving fields, so pointers returned from decl literals (and pointer-typed consts) resolve correctly.
- Add hover/definition coverage for pointer decl literals (const and function return).

### DIFF
--- a/tests/add_analysis_cases.zig
+++ b/tests/add_analysis_cases.zig
@@ -26,7 +26,7 @@ pub fn addCases(
     });
 
     // https://github.com/ziglang/zig/issues/20605
-    var dir = std.fs.openDirAbsolute(b.pathFromRoot(cases_path_from_root), .{ .iterate = true }) catch |err|
+    var dir = std.fs.cwd().openDir(cases_path_from_root, .{ .iterate = true }) catch |err|
         std.debug.panic("failed to open '{s}': {}", .{ cases_path_from_root, err });
     defer dir.close();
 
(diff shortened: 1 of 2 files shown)


vivekstills:
@pusewicz @alexrp @prime31 @kcbanner

vivekstills:
@Techatrix  Could you please review this PR when you get a moment? All checks have passed.

src/analysis.zig
```diff
@@ -5873,6 +5873,7 @@ pub fn lookupSymbolFieldInit(
         const unwrapped =
             try analyser.resolveUnwrapErrorUnionType(container_type, .payload) orelse
             try analyser.resolveOptionalUnwrap(container_type) orelse
+            try analyser.resolveDerefType(container_type) orelse
             break;
         container_type = unwrapped;
     }
```
Techatrix: We can simplify this by using `resolveDeclLiteralResultType` which does exactly what is needed for decl literals. The entire while loop can be replaced with the following code:
```zig
container_type = try container_type
    .resolveDeclLiteralResultType()
    .instanceTypeVal(analyser) orelse container_type;
```

vivekstills:
@Techatrix Updated the code as suggested, let me know if anything else is needed.

vivekstills:
@Techatrix is it ready to be merged?

Issue #2525, opened 2025-12-06
Build error against current `zig` master branch


lukeflo:
### Zig Version

0.16.0-dev.1569+db15df5da

### ZLS Version

60cff3d69f6072b840aa588fb8454fac9a95fbe1

### Steps to Reproduce and Observed Behavior

I build `zig` from source with `llvm` and `clang` 21 also built from source. That went well.

Cloning `zls` repo and building with `zig` results in following error:

```
❯ zig build -Doptimize=ReleaseSafe
install
└─ install zls
   └─ compile exe zls ReleaseSafe native 5 errors
src/DocumentScope.zig:769:10: error: no field named 'asm_legacy' in enum 'zig.Ast.Node.Tag'
        .asm_legacy,
        ~^~~~~~~~~~
/home/lukeflo/Documents/packages/zig/build/stage3/lib/zig/std/zig/Ast.zig:2990:21: note: enum declared here
    pub const Tag = enum {
                    ^~~~
referenced by:
    init: src/DocumentScope.zig:547:25
    create: src/DocumentStore.zig:348:62
    14 reference(s) hidden; use '-freference-trace=16' to see all references
src/analysis.zig:2929:10: error: no field named 'asm_legacy' in enum 'zig.Ast.Node.Tag'
        .asm_legacy,
        ~^~~~~~~~~~
/home/lukeflo/Documents/packages/zig/build/stage3/lib/zig/std/zig/Ast.zig:2990:21: note: enum declared here
    pub const Tag = enum {
                    ^~~~
src/ast.zig:694:10: error: no field named 'asm_legacy' in enum 'zig.Ast.Node.Tag'
        .asm_legacy => {
        ~^~~~~~~~~~
/home/lukeflo/Documents/packages/zig/build/stage3/lib/zig/std/zig/Ast.zig:2990:21: note: enum declared here
    pub const Tag = enum {
                    ^~~~
src/ast.zig:1521:10: error: no field named 'asm_legacy' in enum 'zig.Ast.Node.Tag'
        .asm_legacy,
        ~^~~~~~~~~~
/home/lukeflo/Documents/packages/zig/build/stage3/lib/zig/std/zig/Ast.zig:2990:21: note: enum declared here
    pub const Tag = enum {
                    ^~~~
src/features/semantic_tokens.zig:740:10: error: no field named 'asm_legacy' in enum 'zig.Ast.Node.Tag'
        .asm_legacy => {
        ~^~~~~~~~~~
[...]

PR #2526, opened 2025-12-07
fix(build): remove `asm_legacy` references for latest Zig

neoto:
Resolves https://github.com/zigtools/zls/issues/2525

The minimum version now corresponds to [this](https://codeberg.org/ziglang/zig/commit/5d96a58f1d02d162fdff3e0ef0d742ee1d5074a7) commit - which matches what I'm currently running - but the build actually started failing since [#30018](https://codeberg.org/ziglang/zig/pulls/30018). I don't use Nix so things related to that probably need to be addressed additionally.

I guess I should note that the formatter reordered the imports, but I can revert these changes if necessary...

### DIFF
--- a/.github/workflows/artifacts.yml
+++ b/.github/workflows/artifacts.yml
@@ -39,7 +39,7 @@ jobs:
       - uses: mlugg/setup-zig@v2
         if: env.SKIP_DEPLOY != 'true'
         with:
-          version: 0.16.0-dev.1484+d0ba6642b
+          version: master
 
       - name: Install APT packages
         if: env.SKIP_DEPLOY != 'true'
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -16,7 +16,7 @@ jobs:
 
       - uses: mlugg/setup-zig@v2
         with:
-          version: 0.16.0-dev.1484+d0ba6642b
+          version: master
 
       - name: prefetch dependencies (workaround)
         run: |
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -20,7 +20,7 @@ jobs:
 
       - uses: mlugg/setup-zig@v2
         with:
-          version: 0.16.0-dev.1484+d0ba6642b
+          version: master
 
       - name: Run zig fmt
         if: matrix.os == 'ubuntu-22.04'


Techatrix (changes requested):
The following two function are now unused and should be removed:
- https://github.com/neoto/zls/blob/e8e542a2ce47ce73a6562edf586d8f63d65e1e2c/src/ast.zig#L120C4-L120C23
- https://github.com/neoto/zls/blob/e8e542a2ce47ce73a6562edf586d8f63d65e1e2c/src/ast.zig#L206

The following test should also be removed:
- https://github.com/neoto/zls/blob/e8e542a2ce47ce73a6562edf586d8f63d65e1e2c/tests/lsp_features/semantic_tokens.zig#L1880

> The minimum version now corresponds to [this](https://codeberg.org/ziglang/zig/commit/5d96a58f1d02d162fdff3e0ef0d742ee1d5074a7) commit

The `minimum_zig_version` should be an exact version that is provided on [ziglang.org](https://ziglang.org/download/). The prebuilt Zig binaries are currently not updating so this PR cannot be merged until then.

> I don't use Nix so things related to that probably need to be addressed additionally.

I can take care of that before merging.

> I guess I should note that the formatter reordered the imports, but I can revert these changes if necessary...

Yes, that should be reverted.

neoto:
Thank you for the review!

> The following two function are now unused and should be removed:
> 
> * https://github.com/neoto/zls/blob/e8e542a2ce47ce73a6562edf586d8f63d65e1e2c/src/ast.zig#L120C4-L120C23
> * https://github.com/neoto/zls/blob/e8e542a2ce47ce73a6562edf586d8f63d65e1e2c/src/ast.zig#L206
> 
> The following test should also be removed:
> 
> * https://github.com/neoto/zls/blob/e8e542a2ce47ce73a6562edf586d8f63d65e1e2c/tests/lsp_features/semantic_tokens.zig#L1880

Done

> > The minimum version now corresponds to [this](https://codeberg.org/ziglang/zig/commit/5d96a58f1d02d162fdff3e0ef0d742ee1d5074a7) commit
> 
> The `minimum_zig_version` should be an exact version that is provided on [ziglang.org](https://ziglang.org/download/). The prebuilt Zig binaries are currently not updating so this PR cannot be merged until then.

I've reverted that change because the rest is otherwise valid, plus someone building ZLS from scratch will probably be using a newer Zig version anyway. Once the binaries start being updated again, the version change can be revisited - what do you think?
 
> > I guess I should note that the formatter reordered the imports, but I can revert these changes if necessary...
> 
> Yes, that should be reverted.

Done. Would be nice if the repo used a config with import sorting disabled, but that currently doesn't seem to be supported with zls.json.

techatrix:
> I've reverted that change because the rest is otherwise valid, plus someone building ZLS from scratch will probably be using a newer Zig version anyway. Once the binaries start being updated again, the version change can be revisited - what do you think?

It doesn't matter if the `minimum_zig_version` is updated or not. This PR will fix building ZLS against master branch of Zig but in return break building against the latest prebuilt binary. I prefer to support the latter until they are back in sync at which point there is no longer a reason to withhold from updating the minimum_zig_version.

neoto:
> It doesn't matter if the `minimum_zig_version` is updated or not. This PR will fix building ZLS against master branch of Zig but in return break building against the latest prebuilt binary. I prefer to support the latter until they are back in sync at which point there is no longer a reason to withhold from updating the minimum_zig_version.

Alrighty. I see Andrew mentioned on Zulip that he'll update the script soonish so I hope we don't have to wait too long.

neoto:
Looks like this should be good to go now that the binaries are being updated again. Consequently, I've bumped the minimum version, which matches [this](https://codeberg.org/ziglang/zig/commit/6b9125cbe662d530160e0732c856aa0da86894c0) commit.

I'll let you handle the Nix stuff, along with the workflow versions.

PR #2535, opened 2025-12-24
remove deprecated pool logic and replaced with async

pollend:
The current 0.16 branch of zig has deprecated Pool in favor of Io.Thread and Io.Group . This just updates all the threading logic with this change.

https://codeberg.org/ziglang/zig/commit/985a3565c6130c7279319e9c36642f0b958e6944#diff-f192d18f39c389fafe62a3a341b75c760701fcae

### DIFF
--- a/src/DocumentStore.zig
+++ b/src/DocumentStore.zig
@@ -22,7 +22,7 @@ allocator: std.mem.Allocator,
 /// the DocumentStore assumes that `config` is not modified while calling one of its functions.
 config: Config,
 lock: std.Thread.RwLock = .{},
-thread_pool: *std.Thread.Pool,
+thread_pool: *std.Io.Threaded,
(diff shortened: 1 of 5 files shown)


techatrix:
Thanks for the PR. It looks like the Io interface is being used in way that doesn't quite match what I would have expected. Here are some examples:
- Most of the code is directly using a specific Io implementation like `std.Io.Threaded` instead of using the `std.Io` interface.
- There are multiple instances of `std.Io.Threaded` being used. The main function in `main.zig`, `build_runner.zig` and `analysis_check.zig` already created one at the start.
- Do not discard the return value of `std.Io.async`, it must be awaited or cancelled
- Use `std.testing.io` to get an io instance in tests

I can save you the effort of addressing these issues since I've already implemented this is #2537.

PR #2546, opened 2026-01-01
Io.Group.wait replaced with Io.Group.awaitUncancelable

jozip:
`Io.Group.wait` doesn't exist as of (at least) `zig-0.16.0-dev.1889+4ad8bc341`. My guess is that `Io.Group.awaitUncancelable` is the correct analogue, since all the tests pass and everything seem to work properly. Whether or not it's desirable having (a)wait being "uncancelable", I leave it as an exercise for those better in the know.

Edit: Whoops, there is no such thing as `std.Group`.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -56,11 +56,11 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1766687554,
-        "narHash": "sha256-DegN7KD/EtFSKXf2jvqL6lvev6GlfAAatYBcRC8goEo=",
+        "lastModified": 1767051569,
+        "narHash": "sha256-0MnuWoN+n1UYaGBIpqpPs9I9ZHW4kynits4mrnh1Pk4=",
         "owner": "NixOS",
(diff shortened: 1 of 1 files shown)


xdBronch:
it just got introduced but fyi `awaitUncancelable` is being removed in https://codeberg.org/ziglang/zig/pulls/30634 :smile:

jozip:
> it just got introduced but fyi `awaitUncancelable` is being removed in https://codeberg.org/ziglang/zig/pulls/30634 😄

Haha, I should've known 😄 Thanks for the heads-up! I guess I can become better-in-the-know myself by taking a stab at how cancellation is supposed to work here.

jozip:
From what I gather, we don't care about `await` returning `error.Cancelable` the way it's used here. It feels weird to just ignore the error, but right now I can't see how it makes sense to handle it any other way.

techatrix:
As far as I know, you should generally let the cancellation error bubble up the call stack instead of ignoring it.

PR #2552, opened 2026-01-05
feat: inlay hints for decl literal functions

dotcarmen:
This started when i noticed a bunch of output in the lsp logs that say `cannot deduce fn expression with tag '.enum_literal'`

This PR accomplishes the following:
- opaque types are now also searched for decl literals
- enum literal types are now resolved if there's a binding type
- inlay hints are now printed for declaration literal function calls


dotcarmen:
another breaking change in std... patching for latest master now

Techatrix (approved):

PR #2555, opened 2026-01-06
feat: support 0.16.0-dev.1976+8e091047b

dotcarmen:
i'm not sure exactly _which_ master release is associated with which changes, so this is best-guess

- `std.process.SpawnOptions.StdIo.Ignore` -> `std.process.SpawnOptions.StdIo.ignore`
- use `std.process.Init.Minimal`
  - `build_runner.zig` detects whether this usage is necessary or not and selects an entrypoint function based on the zig release version
- `std.Build.Graph` has a new field `cwd` for the path of the cwd
- `std.Build.Watch` also gained a new param `cwd_path` which is now propagated

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -345,7 +345,7 @@ fn getVersion(b: *Build) std.SemanticVersion {
         "git", "-C", b.pathFromRoot("."), "--git-dir", ".git", "describe", "--match", "*.*.*", "--tags",
     };
     var code: u8 = undefined;
-    const git_describe_untrimmed = b.runAllowFail(argv, &code, .Ignore) catch |err| {
+    const git_describe_untrimmed = b.runAllowFail(argv, &code, .ignore) catch |err| {
         const argv_joined = std.mem.join(b.allocator, " ", argv) catch @panic("OOM");
         std.log.warn(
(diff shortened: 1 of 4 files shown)


dotcarmen:
for some reason there's a segfault in `std.Build.create`, i _believe_ it's an upstream issue but i'm looking into it

xdBronch:
that segfault is happening with a `0xaa` pointer and youre setting the graph to `undefined`, looks very related

dotcarmen:
🤦🏻‍♀️ yup, you're right. i got distracted with other things, i'll fix that in a little bit...

dotcarmen:
fixed, though now there's an error in `lsp-kit` since it uses the deprecated args iterator api. i see https://github.com/zigtools/lsp-kit/pull/32 is already opened, once that's merged then i can update `build.zig.zon` and then this should be mergeable :)

dotcarmen:
ah i forgot the `nix flake update` step, i unfortunately deleted nix from my system, if a maintainer could do that before merging that'd be great :)

techatrix:
#2556 got merged instead.

PR #2564, opened 2026-01-10
[WIP] Fixes to the build runner as per zig/ziglang#30754

jozip:
The recent merge of https://codeberg.org/ziglang/zig/pulls/30754 broke our bespoke build runner. It has diverged significantly from `lib/compiler/build_runner.zig` and it is starting to ache pretty bad working with it. So I figured it would be a good time to pull the recent runner from the compiler and smush the customizations on top and start documenting what goes where so it will be a bit easier to track and fix breaking changes in the future.

Everything builds, but the tests fail due to no output from the runner. I'm sure it's something trivial that I've missed, but I'm now at a point where I do random pokes. So I think it's for the best to let this mess get some eyeballs other than mine. (Assuming this is a good approach to begin with.)

### DIFF
--- a/src/build_runner/build_runner.zig
+++ b/src/build_runner/build_runner.zig
@@ -1,7 +1,7 @@
 //! PLEASE READ THE FOLLOWING MESSAGE BEFORE EDITING THIS FILE:
 //!
 //! This build runner is targeting compatibility with the following Zig versions:
-//!   - 0.15.1 or later
+//!   - 0.16.0 or later
 //!
 //! Handling multiple Zig versions can be achieved with one of the following strategies:
 //!   - use `@hasDecl` or `@hasField` (recommended)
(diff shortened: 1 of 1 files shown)


TibboddiT:
I'm not familiar with the build runner customizations made by zls, but would it be possible to "just" replace `printStepFailure` with something that serves the errors the way `serveWatchErrorBundle` did? (maybe what I'm describing is more or less what is already going on)

That would limit the customizations mainly to this function, and maybe make proposing to the zig core team a mechanism to provide a user-defined implementation feasible?

llogick:
Always follow the golden rule:
Make It Work, Make It Right, Make It Fast

and right now it isn't working :)

[minimal diff](https://github.com/zigtools/zls/commit/385ee2193c5ca9aab54522b6ab38a6a724f55a6c?diff=unified)

jozip:
> Always follow the golden rule: Make It Work, Make It Right, Make It Fast
> 
> and right now it isn't working :)
> 
> [minimal diff](https://github.com/zigtools/zls/commit/385ee2193c5ca9aab54522b6ab38a6a724f55a6c?diff=unified)

It works? It works! Get that sucker on a PR, stat! 😄 I guess there is no need to make an effort to track the upstream (side-stream?) build runner then. I'll close this PR.

llogick:
> It works? It works! Get that sucker on a PR, stat! 😄 ...

@jozip 
Do you mind taking over? You can have the commit/diff

I like writing code, but can't stand the slow, back and forth nature of the PR process

jozip:
> > It works? It works! Get that sucker on a PR, stat! 😄 ...
> 
> @jozip Do you mind taking over? You can have the commit/diff
> 
> I like writing code, but can't stand the slow, back and forth nature of the PR process

Sure, I'll get to it.

techatrix:
Rebasing the ZLS build runner based on changes from the Zig build runner is generally a welcome contribution. I have a similar change in #2510  (c8098a4783a1dc9ae9fc7e6b7dc9076202f907f9) which has not been merged yet, and has gone out of date again. Also note that any unused or unnecessary code should be removed from the build runner instead of being commented out.

jozip:
> Rebasing the ZLS build runner based on changes from the Zig build runner is generally a welcome contribution. I have a similar change in #2510 ([c8098a4](https://github.com/zigtools/zls/commit/c8098a4783a1dc9ae9fc7e6b7dc9076202f907f9)) which has not been merged yet, and has gone out of date again. Also note that any unused or unnecessary code should be removed from the build runner instead of being commented out.

Perhaps it's a dumb question, but why not write a purpose-built build runner for ZLS? Or is it strictly necessary that it remain a subset of Zig's?

techatrix:
The code in the ZLS build runner is not a strict subset of the code in the Zig build runner. We do have a bunch of code that is specific to what ZLS needs. But there is also a lot of shared logic that is easier to keep in sync with Zig rather than re implementing it ourselves. We wouldn't gain anything from doing so and only make maintaining the build runner even more cumbersome.

jozip:
Alright, I see.

PR #2570, opened 2026-01-11
fix: param placeholders for types with right brace

theofabilous:
Completions for functions with snippet placeholders are broken when an argument's type is rendered with a right brace (`}`). The right brace symbol is rendered unescaped, which causes snippet parsers to interpret it as the end of the placeholder, leaving an extra trailing `}`.

Example:

```zig
fn foo(x: struct { x: i32 }) void {
    _ = x;
}
```
```
When expanding the completion item for a function call of `foo` with snippet placeholders,
we get the following:

     selected region
     ______________
    |              |
foo(|x: struct {...|})


The snippet is parsed as follows:

      __ placeholder start
     |
foo(${1:x: struct {...}})
                  ^   ^^
                 /    | \_____ not part of snippet placeholder
            ignored   |
                     placeholder end

```

This occurs (at least) in the following cases:
1. type is an anonymous container literal (`enum { ... }`)
2. type's name is a raw identifier with a `}` (`@"Evil}Name"`)
3. type is a generic function with a type argument that is itself problematic (`std.ArrayList(struct {})`, ...)

Fixed by adding an `escape_r_braces` option to `FormatOptions` and escaping '`}`' symbols (as specified by [LSP protocol - Snippet syntax](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#snippet_syntax)) in `Type.rawStringify()` when this option is set to `true`. When `rawStringifyParameter` calls this function, it sets `options.escape_r_braces` to `true` if snippet placeholders are enabled.

In `Type.rawStringify`, I only specifically handle `escape_r_braces` in cases where I've encountered this bug. There are a few lines in that function that render strings as is (without potentially escaping r-braces) which I didn't touch because I haven't encountered errors related to those code paths (and in most cases, wasn't sure if those code paths are reachable with snippet placeholders and/or can ever yield `}` symbols).

I did this so that the changes map nicely onto the added test cases (removing any change causes at least one of the added tests to fail).

### DIFF
--- a/src/analysis.zig
+++ b/src/analysis.zig
@@ -237,6 +237,7 @@ fn rawStringifyParameter(
         try info.type.rawStringify(writer, analyser, .{
             .referenced = referenced,
             .truncate_container_decls = true,
+            .escape_r_braces = options.snippet_placeholders,
         });
     }
 
@@ -4193,6 +4194,7 @@ pub const Type = struct {
     pub const FormatOptions = struct {
         referenced: ?*ReferencedType.Set = null,
         truncate_container_decls: bool,
+        escape_r_braces: bool = false,
     };
 
     fn rawStringify(
@@ -4247,7 +4249,11 @@ pub const Type = struct {
                     }
                     try elem_ty.rawStringify(writer, analyser, options);
                 }
-                try writer.writeAll(" }");
+                if (options.escape_r_braces) {
+                    try writer.writeAll(" \\}");
+                } else {
+                    try writer.writeAll(" }");
+                }
             },
             .optional => |child_ty| {
                 try writer.writeByte('?');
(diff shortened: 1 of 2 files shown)


techatrix:
Thank you for looking into this. Escaping `}` wherever it is written while stringifying types can resolve some of the common cases where this issue is encountered, this however doesn't handle every possible edge case. As you mentioned, rendered strings or identifiers may include characters that need to be escaped as well. The snippet syntax also has other characters that need to be escaped like `$` and `\`.

Coincidentally I had also worked on a fix for the same issue. I already handled most of those edge cases I mentioned so I can save you the effort of updating this PR accordingly. See #2574.

theofabilous:
Cool, glad to see that a fix was merged!

PR #2582, opened 2026-01-23
feat(goto): goto definition prefers builtin This()

### LINKED ISSUE #1596: Go to definition on file struct should look for `@This()`.
```zig
// If I go to definition on `File` -
const File = @import("path/to/File.zig");
// ZLS should search `File.zig` for a `@This()` assigned to the same name.
const File = @This();
```

dotcarmen:
closes #1596

consider goto on `Io` from `std.Io`. goto definition currently goes to the beginning of the file, but it's typically more helpful to go to `Io` in its internal declaration `const Io = @This()`.

contrary to the issue description, this PR also goes-to such definitions for non-file container definitions. Note that goto-declaration still goes to the outer declaration, while goto-definition now goes to the `@This()` declaration if it exists.

### DIFF
--- a/src/analysis.zig
+++ b/src/analysis.zig
@@ -4168,9 +4168,33 @@ pub const Type = struct {
 
     pub fn typeDefinitionToken(self: Type) ?TokenWithHandle {
         return switch (self.data) {
-            .container => |info| .{
-                .token = info.scope_handle.handle.tree.firstToken(info.scope_handle.toNode()),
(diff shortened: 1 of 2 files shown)


xdBronch:
im not sure i like this happening for all structs, i have a local change that is similar but instead stores the `const x = @This();` node on `DocumentStore.Handle`s which can be used for whatever, f.e. stringification or collecting doc comments without needing to recalculate it for everything.

PR #2608, opened 2026-02-15
restrain cross-file symbol references within it's build system module

### LINKED ISSUE #1071: Broken cross-file references

Techatrix:
fixes #1071

When requesting references on a file that is within a build system module, the search will be constrained within the given module. This should give more predictable results compared to flawed approach that is preceded it (see #1071). However, this does not yet account from references across modules (or projects) which can be implemented in the future.

This change also fixes an issue with #2339 where symbol references would load all build files inside of the workspace. This is especially problematic with projects that have many build files like the Zig compiler codebase.

Followup Issues:

- #2613
- #2614
- #2612

### DIFF
--- a/src/features/references.zig
+++ b/src/features/references.zig
@@ -108,7 +108,7 @@ const Builder = struct {
             builder.target_symbol.nameToken(),
         );
 
-        switch (tree.nodeTag(node)) {
+        var candidate: Analyser.DeclWithHandle, const name_token = candidate: switch (tree.nodeTag(node)) {
             .identifier,
             .test_decl,
             => |tag| {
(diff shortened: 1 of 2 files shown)


xdBronch:
```zig
// a.zig
const b = @import("b.zig");
const Foo = b.Foo;
comptime {
    _ = Foo;
}

// b.zig
pub const Foo = struct {};
```
this fails to find the `_ = Foo;`

techatrix:
> this fails to find the `_ = Foo;`

Fixed.

PR #2618, opened 2026-02-20
fix completions replacing punctuation

xdBronch:
basically the same as https://github.com/zigtools/zls/pull/2605, the `(`,`[`,`{` would get replaced when completions were manually triggered
this affected a lot more than what I added tests for but it seems unlikely that any one in particular would uniquely regress so I just did some more likely cases to keep it small

### DIFF
--- a/src/features/completions.zig
+++ b/src/features/completions.zig
@@ -487,7 +487,6 @@ fn prepareCompletionLoc(tree: *const Ast, source_index: usize) offsets.Loc {
             std.debug.assert(token_loc.start <= source_index and source_index <= token_loc.end);
             return offsets.identifierIndexToLoc(tree.source, token_loc.start, if (tag == .builtin) .name else .full);
         },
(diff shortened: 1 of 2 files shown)


techatrix:
This does not handle "replace" text edits correctly in certain cases:

```zig
const foo = 5;
const bar = foo(f<cursor>oo);
```
The "replace" text edit should change the line to the following:
```zig
const bar = foo(foo);
```
but instead it does the same as the "insert" text edit:
```zig
const bar = foo(foooo);
```

Test Case:
```zig
    try testCompletionTextEdit(.{
        .source =
        \\const foo = 5;
        \\const bar = foo(<cursor>foo);
        ,
        .label = "foo",
        .expected_insert_line = "const bar = foo(foofoo);",
        .expected_replace_line = "const bar = foo(foo);",
    });
```

Since I had some related fixes to completion text edits, I went ahead and fixed this in #2621.

PR #2623, opened 2026-02-24
sort deprecated items to the bottom

### LINKED ISSUE #2155: The LSP should put deprecated suggestions after non-deprecated suggestions
I think that the LSP might be putting suggestions with a compiler error before other suggestions, although it is also possible that this is a issue with my text editor (helix).

Image

Image

nkbelov:
Closes #2155

Screenshot 2026-02-24 at 13 54 40

### DIFF
--- a/src/features/completions.zig
+++ b/src/features/completions.zig
@@ -665,7 +665,8 @@ fn kindToSortScore(kind: types.completion.Item.Kind) []const u8 {
 }
 
 fn itemSortScore(item: types.completion.Item) []const u8 {
-    const deprecated: bool = if (item.tags) |tags|
+    // Completion items have two ways to mark deprecation; we need to check both.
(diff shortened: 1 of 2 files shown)


src/features/completions.zig
```diff
@@ -664,6 +664,19 @@ fn kindToSortScore(kind: types.completion.Item.Kind) []const u8 {
     };
 }
 
+fn itemSortScore(item: types.completion.Item) []const u8 {
+    const deprecated: bool = if (item.tags) |tags|
```
Techatrix: The LSP protocols has two ways to mark a completion item as deprecated. The new one is with the `tags` field but there is also the legacy `deprecated` field that should also be checked.

```suggestion
    const deprecated: bool = item.deprecated orelse if (item.tags) |tags|
```

Techatrix (changes requested):

PR #2633, opened 2026-03-04
Improve type resolution of builtin functions

FnControlOption:
e.g. some of the parameters for `@Fn`, `@Struct`, `@Union`

### DIFF
--- a/src/analysis.zig
+++ b/src/analysis.zig
@@ -4456,6 +4456,90 @@ pub fn resolveImportString(analyser: *Analyser, handle: *DocumentStore.Handle, i
 fn resolveLangrefType(analyser: *Analyser, type_str: []const u8) Error!?Type {
     if (try analyser.resolvePrimitive(type_str)) |primitive|
         return Type.fromIP(analyser, primitive, null);
+
+    // e.g. `?type`
+    if (std.mem.startsWith(u8, type_str, "?")) {
+        const elem_str = type_str[1..];
+        const elem_ty = try analyser.resolveLangrefType(elem_str) orelse return null;
+        return .{
+            .data = .{
+                .optional = try analyser.allocType(try elem_ty.typeOf(analyser)),
+            },
+            .is_type_val = false,
+        };
+    }
+
+    // e.g. `*const anyopaque`
+    if (std.mem.startsWith(u8, type_str, "*")) {
+        var elem_str = type_str[1..];
+        const is_const = std.mem.startsWith(u8, elem_str, "const ");
+        if (is_const)
+            elem_str = elem_str[6..];
+        const elem_ty = try analyser.resolveLangrefType(elem_str) orelse return null;
+        return .{
+            .data = .{
+                .pointer = .{
+                    .size = .one,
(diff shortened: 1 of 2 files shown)


xdBronch:
its currently a big fat hack but locally ive played with the idea of creating a dummy file that contains declarations mimicking the parameters of the builtin, f.e. `@Fn` would look something like this
```zig
const param_types: []const type = call.args[0];
const param_attrs: *const [param_types.len]std.builtin.Type.Fn.Param.Attributes = call.args[1];
const ReturnType: type = call.args[2];
const attrs: std.builtin.Type.Fn.Attributes = call.args[3];
```
this means it can be parsed and analysed just like any other code which makes it quite accurate with no need to special-case what any of the builtins currently do. i dont love how ive done this but i think its something worth exploring at some point

techatrix:
@xdBronch That is a very fascinating idea. But I don't think this is a practical solution to resolve builtins over just manually implementing the analysis code for each builtin. Relying on the language reference has been helpful when builtins were constantly changing but as the language stabilizes we can just bite the bullet and hardcore all the builtins.

Just imagine how to resolve the RLS based pointer type in `@fieldParentPtr`. Or running `@Int(.unsigned, @log2(@bitSizeOf(T)))` at comptime to resolve the second parameter of `@shlExact`. Truly fascinating...

xdBronch:
yeah fair enough, hardcoding is definitely better in the long run. maybe we can use fake files like this for other fun stuff :)

FnControlOption:
Days since last GitHub incident: 0 💀

FnControlOption:
I agree that hardcoding builtin support will likely be best in the long term, but I also agree that using the language reference is a good stopgap until Zig 1.0 is released.

<details><summary>List of types used by builtin functions</summary>

```
?*StackTrace
?Element
?T
?type
(field)
[:0]const u8
[]const []const u8
[]const type
[]const u8
@TypeOf(value)
@Vector(a_len, E)
@Vector(b_len, E)
@Vector(len, bool)
@Vector(len, T)
@Vector(mask_len, E)
@Vector(mask_len, i32)
*const [field_names.len]TagInt
*const [field_names.len]type
*const [field_names.len]Type.StructField.Attributes
*const [field_names.len]Type.UnionField.Attributes
*const [N:0]u8
*const [param_types.len]Type.Fn.Param.Attributes
*const anyopaque
*const T
*T
*VaList
anyerror
anytype
AtomicOrder
AtomicRmwOp
bool
BranchHint
CallModifier
comptime_int
DestType
E
ExportOptions
ExternOptions
FloatMode
isize
Log2T
noreturn
PrefetchOptions
ReduceOp
Signedness
SourceLocation
std.meta.Int(.unsigned, @bitSizeOf(anyerror))
struct { @TypeOf(a, b), u1 }
struct { @TypeOf(a), u1 }
T
type
Type
Type.ContainerLayout
Type.Enum.Mode
Type.Fn.Attributes
Type.Pointer.Attributes
Type.Pointer.Size
u1
u16
u32
Union
usize
VaList
void
```

</details>

PR #2634, opened 2026-03-06
Resolve type of `std.meta.{Tag,ArgsTuple}` call

FnControlOption:
```zig
const EnumA = enum {
    foo,
    bar,
};

const TaggedUnionA = union(EnumA) {
    foo: u8,
    bar: i32,
};

const TagA = std.meta.Tag(TaggedUnionA);
//    ^^^^ (type)(EnumA)

const ArgsTupleA = std.meta.ArgsTuple(fn (u8, i32) void);
//    ^^^^^^^^^^ (type)(struct { u8, i32 })
```

- Skipped `std.meta.{Child,Elem,Sentinel}` for now since they require interacting with `InternPool` and overlap with `resolveDerefType`, `resolveBracketAccessType`, etc.
- Skipped `std.meta.{FieldEnum,DeclEnum}` since we can't yet create enums with `InternPool`
- Skipped support for calling `std.meta.Tag` with an enum since (a) the enum tag can be implicit and (b) our type analysis of integer types currently isn't very helpful anyways

### DIFF
--- a/src/analysis.zig
+++ b/src/analysis.zig
@@ -1916,6 +1916,21 @@ fn resolveTypeOfNodeUncached(analyser: *Analyser, options: ResolveOptions) Error
 
             func_ty = try analyser.resolveFunctionTypeFromCall(handle, call, func_ty);
             const func_info = func_ty.data.function;
+            const func_uri = func_info.handle.uri.raw;
+
+            if (std.mem.endsWith(u8, func_uri, "/std/meta.zig") and func_info.name != null) {
+                const func_name = func_info.name.?;
+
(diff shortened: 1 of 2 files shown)


techatrix (approved):

BraynStorm:
`std.meta.FieldEnum` would be awesome, as `MultiArrayList` relies on it, among other things.

I tried hacking around to get it working, but it seems the InternPool is setup in a way that if I don't have an AST node for an `enum` I cannot emit a container that is an enum... or maybe I'm missing something, I just spent an afternoon and that's the best I got. I could reliably "lie" to ZLS that the thing is a `struct` and would get field-like completion, but that's not great.

Issue #2346, opened 2025-05-30
Tracking issue for Zig AST parser limitations


Techatrix:
The Zig parser provided by the standard library `std.zig.Ast` has poor error resilience. This issue is going to keep track of the various issues that have been caused by it. In the future, we will have to look into forking the existing parser or writing our own one from scratch.

psznm:
Do you have some thoughts on how to proceed with this? Going with any solution will be huge commitment, but I think this is holding zls back, so some progress on this would be huge.

Issue #1845, opened 2024-03-30
sort completions based on the result location type at the current cursor position


Techatrix:
In the following code snippet, in which order should the completions be ordered?
```zig
const large_number: u64 = 0;
const small_number: u32 = 0;
const byte: u8 = 0;
const string: []const u8 = "";
fn foo(_: u32) u32 {}
test {
    foo(<cursor>);
}
```
`byte`, `small_number` and `foo()` should come first because only they can coerce to the result location type `u32`.
I can imagine many possible sorting rules on how to use RLS to improve completion and this example is just one of many.

PR #2637, opened 2026-03-07
complete dot after `try` keyword

### LINKED ISSUE #2627: Trigger code completion after keyword `try`
In this case, `.` can trigger code completion:
Image

But if it follows the `try` keyword, it won't work:
Image

Even if you continue typing, it cannot trigger:
Image

## Idea

For the code `const a: T =`, I think that when I type `.` to trigger code autocompletion, it should list functions with return types `T` and `!T` at the top. If it's after `const a: T = try`, then functions with return type `!T` should be listed first.

nkbelov:
Closes #2627

Looks like simply ignoring `try` and looking back at the token before that achieves the goal.

I considered implementing the second request in the issue (to sort fallible functions to be first in the completion list if we're in the context of a `try` keyword), but it looks to be a bit less trivial since it would require storing something like

```zig
const ItemWithExtras = struct {
    item: types.completion.Item,
    is_err_fn: bool,
};
```
instead of plain `types.completion.Item` in the `Builder` for it to produce correct sorting, which is a bit invasive and might warrant a separate PR, so I've left it for now.

### DIFF
--- a/src/features/completions.zig
+++ b/src/features/completions.zig
@@ -1117,6 +1117,8 @@ fn getEnumLiteralContext(
     // Allow using `1.` (parser workaround)
     var token_index = if (tree.tokenTag(dot_token_index - 1) == .number_literal)
         (dot_token_index - 2)
+    else if (tree.tokenTag(dot_token_index - 1) == .keyword_try)
+        (dot_token_index - 2)
     else
         (dot_token_index - 1);
     if (token_index == 0) return null;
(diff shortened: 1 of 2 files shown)


techatrix:
Sorting completion items based on the context around the cursor position is a non trivial issue that can have many different cases beyond the one presented here with `try` (related issue #1845). The current implementation of code completions for decl literals is too inflexible to serve this use case well and must be first rewritten based using a [error resilient parser](https://github.com/zigtools/zls/issues/2346) before work on this can begin.

Techatrix (approved):

PR #2645, opened 2026-03-11
Fixed resolved types related to mutable bracket access

### LINKED ISSUE #2534: zls can give incorrect type hints with slices
### Zig Version

0.15.2

### ZLS Version

0.15.1

Zed and VSCode

_(Looking back on this, this is probably more of an enhancement)_
The interesting thing with almost of these type hints, is, that zls picks a type that a general slice can be coerced to.

Example 1:
```zig
const the_string2 = try std.fmt.allocPrintSentinel(alloc, "{s}", .{the_string}, 0); //zls thinks this is []u8
std.debug.print("type: {any}\n", .{@TypeOf(the_string2)}); //however it is [:0]u8
```
Example 2:
```zig
const attempt2 = the_string2[0..the_string2.len] //zls thinks this is []const u8
std.debug.print("type: {any}\n", .{@TypeOf(attempt2)}); //however it is []u8
```
Example 3:
```zig
const attempt2 = the_string2[0.. :0] //zls thinks this is []u8
std.debug.print("type: {any}\n", .{@TypeOf(attempt2)}); //however it is [:0]u8
```
Example 4:
```zig
const attempt2 = the_string2[0..the_string2.len :0] //zls thinks this is [:0]const u8
std.debug.print("type: {any}\n", .{@TypeOf(attempt2)}); //however it is [:0]u8
```
The problem with these types being slightly incorrect is that the confusion stacks. 
Take for example the following:
```zig
const the_string2 = try std.fmt.allocPrintSentinel(alloc, "{s}", .{the_string}, 0); //returns [:0]u8
const attempt2 = the_string2[0..the_string2.len :0] //this is fine.
```

```zig
const the_string2: []u8 = try std.fmt.allocPrintSentinel(alloc, "{s}", .{the_string}, 0); //changed to what zls assumes the return type is.
const attempt2 = the_string2[0..the_string2.len :0] //this is illegal behavior leading to a panic (ArrayIndexOutOfBounds).
```
This can be confusing for newer people.

### Expected Behavior

`
`

### Log Output

```shell

```

FnControlOption:
Closes #2534

~~Requires #2644~~

### DIFF
--- a/src/analysis.zig
+++ b/src/analysis.zig
@@ -1074,7 +1074,8 @@ pub const BracketAccess = union(enum) {
 /// - `lhs[start..]` (open)
 /// - `lhs[start..end]` (range)
 pub fn resolveBracketAccessType(analyser: *Analyser, lhs: Type, rhs: BracketAccess) error{OutOfMemory}!?Type {
-    return analyser.resolveBracketAccessTypeFromBinding(.{ .type = lhs, .is_const = false }, rhs);
+    const binding = try analyser.resolveBracketAccess(.{ .type = lhs, .is_const = false }, rhs) orelse return null;
+    return binding.type;
 }
 
 // TODO: copy indexing logic from Zig compiler to InternPool, and then delete bracketAccessTypeFromIPIndex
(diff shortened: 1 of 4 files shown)


src/analysis.zig
```diff
@@ -3123,17 +3138,17 @@ pub const Type = struct {
         /// - `[*c]T`
         pointer: struct {
             size: std.builtin.Type.Pointer.Size,
-            /// `.none` means no sentinel
-            sentinel: InternPool.Index,
+            /// `.none` means no sentinel, `null` means unknown sentinel
+            sentinel: ?InternPool.Index,
```
Techatrix: The internpool can already represent unknown types and values so why is `null` being instead?

Techatrix (changes requested):

PR #2654, opened 2026-03-16
fix: make never mutated local vars const

neoto:
Resolves not being able to compile ZLS with [latest Zig](https://codeberg.org/ziglang/zig/pulls/31534).


xdBronch:
this is a backwards compatible change, it shouldnt be necessary to change the required zig version

neoto:
> this is a backwards compatible change, it shouldnt be necessary to change the required zig version

I thought about this as well, but still ended up changing the version in the end. I'll yank it out since that makes more sense, though.

PR #3172, opened 2026-04-11
preserve allowzero, align, volatile, and addrspace on pointer types

### LINKED ISSUE #2603: Inline hint of the pointer type is incomplete.
Image

Output(Except):
```
*u8
*allowzero u8
*align(4) u8
```

barry3406:
The `ptr_type` resolver in `analysis.zig` only read `const_token` from `ptr_info`, and `Data.createPointer` only set `is_const` on the `InternPool.pointer_type` flags. So `*allowzero u8`, `*align(4) u8`, `*volatile u8` and friends all collapsed to `*u8` in inlay hints (and anywhere else that runs through `stringifyTypeOf`).

This reads `volatile_token`, `allowzero_token`, and the literal integer in `ast.align_node` from `ptr_info`, plumbs them through `createPointerType` / `Data.createPointer` into the IP flags and the `.pointer` Data fallback, and updates `rawStringify` so the fallback path matches the InternPool printer.

Address-space resolution from arbitrary expressions still defers to `.generic`; the field is plumbed through but only the literal alignment integer is currently read.

Regression coverage added in `tests/lsp_features/inlay_hints.zig`.

Fixes #2603.

### DIFF
--- a/src/analysis.zig
+++ b/src/analysis.zig
@@ -908,7 +908,7 @@ pub fn resolveOrelseType(analyser: *Analyser, lhs: Type, rhs: Type) error{OutOfM
 
 pub fn resolveAddressOf(analyser: *Analyser, is_const: bool, ty: Type) error{OutOfMemory}!Type {
     const elem_ty = try ty.typeOf(analyser);
-    const pointer_ty = try Type.createPointerType(analyser, .one, .none, is_const, elem_ty);
+    const pointer_ty = try Type.createPointerType(analyser, .one, .none, is_const, false, false, 0, .generic, elem_ty);
     return try pointer_ty.instanceUnchecked(analyser);
 }
 
@@ -1231,12 +1231,12 @@ pub fn resolveBracketAccess(analyser: *Analyser, lhs_binding: Binding, rhs: Brac
     switch (result) {
         .array => |elem_count| {
             const array_ty = try Type.createArrayType(analyser, elem_count, sentinel, elem_ty.*);
-            const pointer_ty = try Type.createPointerType(analyser, .one, .none, is_const, array_ty);
+            const pointer_ty = try Type.createPointerType(analyser, .one, .none, is_const, false, false, 0, .generic, array_ty);
             const pointer_instance = try pointer_ty.instanceUnchecked(analyser);
             return .{ .type = pointer_instance, .is_const = true };
         },
         .slice => {
-            const slice_ty = try Type.createPointerType(analyser, .slice, sentinel, is_const, elem_ty.*);
+            const slice_ty = try Type.createPointerType(analyser, .slice, sentinel, is_const, false, false, 0, .generic, elem_ty.*);
             const slice_instance = try slice_ty.instanceUnchecked(analyser);
             return .{ .type = slice_instance, .is_const = true };
         },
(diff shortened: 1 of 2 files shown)


tests/lsp_features/inlay_hints.zig
```diff
@@ -286,6 +286,33 @@ test "var decl" {
     , .{ .kind = .Type });
 }
 
+test "var decl - pointer modifiers" {
```
Techatrix, line 289: These tests shouldn't need to go through inlay hints. Instead, add new tests to [`tests/analysis/pointer.zig`](https://github.com/zigtools/zls/blob/master/tests/analysis/pointer.zig) which directly target the analysis backend without going through LSP requests.

Example:

```diff
diff --git a/tests/analysis/pointer.zig b/tests/analysis/pointer.zig
index 4086f8d4..5aa7acab 100644
--- a/tests/analysis/pointer.zig
+++ b/tests/analysis/pointer.zig
@@ -270,6 +270,9 @@ const ManyPointerWithSentinel = [*:.{}]addrspace(.generic) const packed struct {
 //                                 ^ (packed struct {})()
 //                                               ^^^^^^^^ (AddressSpace)()
 
+const VolatilePointer = *volatile u32;
+//    ^^^^^^^^^^^^^^^ (type)(*volatile u32)
+
 var runtime_index: usize = 5;
 var runtime_u8: u8 = 1;
 var runtime_i8: i8 = -1;

```

src/analysis.zig
```diff
@@ -3769,10 +3810,24 @@ pub const Type = struct {
         size: std.builtin.Type.Pointer.Size,
         sentinel: InternPool.Index,
         is_const: bool,
+        is_volatile: bool,
+        is_allowzero: bool,
+        alignment: u16,
+        address_space: std.builtin.AddressSpace,
         elem_ty: Type,
```
Techatrix, line 3817: We already have a type that can store most pointer attributes. So the number of function parameters can be reduced here:
```suggestion
        elem_ty: Type,
        sentinel: InternPool.Index,
        flags: InternPool.Key.Pointer.Flags,
```

src/analysis.zig
```diff
@@ -3184,6 +3201,11 @@ pub const Type = struct {
             /// `.none` means no sentinel, `.unknown_unknown` means unknown sentinel
             sentinel: InternPool.Index,
             is_const: bool,
+            is_volatile: bool = false,
+            is_allowzero: bool = false,
+            /// `0` means default alignment.
+            alignment: u16 = 0,
+            address_space: std.builtin.AddressSpace = .generic,
```
Techatrix, line 3208: We already have a type that can store most pointer attributes.

```zig
elem_ty: *Type,
/// `.none` means no sentinel, `.unknown_unknown` means unknown sentinel
sentinel: InternPool.Index,
flags: InternPool.Key.Pointer.Flags,
```

src/analysis.zig
```diff
@@ -3290,6 +3312,10 @@ pub const Type = struct {
             size: std.builtin.Type.Pointer.Size,
             sentinel: InternPool.Index,
             is_const: bool,
+            is_volatile: bool,
+            is_allowzero: bool,
+            alignment: u16,
+            address_space: std.builtin.AddressSpace,
             elem_ty: Type,
```
Techatrix, line 3319: We already have a type that can store most pointer attributes. So the number of function parameters can be reduced here:

```suggestion
            elem_ty: Type,
            sentinel: InternPool.Index,
            flags: InternPool.Key.Pointer.Flags,
```

Techatrix (changes requested):

PR #3175, opened 2026-04-14
feat: add type size/alignment hover information

### LINKED ISSUE #1677: show size/alignment/padding in tooltip for type annotations
^

ShadowCurse:
Add a basic version of size/alignment calculations for types and print these values as a part of the hover tool tip.
There are 2 main paths to get this info:
1. `InternPool` can calculate size/alignment for types it contain, but it does not contain many things
2. AST path can calculate sizes as well, but has some nuances

In general, this machinery seems to function pretty well from my testing. There are some limitations though:
- anything requiring `comptime` does not show any info. It also recursively prevents parent types from showing any info as well
- `extern` structs cannot have proper size calculations since AST can mess up the field ordering

Maybe `extern` can be solved if someone more knowledgeable about internals can take a look at the problem.

Fixes: #1677

### DIFF
--- a/src/analyser/InternPool.zig
+++ b/src/analyser/InternPool.zig
@@ -3604,6 +3604,348 @@ pub fn getNamespace(ip: *InternPool, ty: Index) NamespaceIndex {
     };
 }
 
+pub fn typeAlignment(ip: *InternPool, ty: Index, target: std.Target) ?u16 {
+    return switch (ty) {
+        .bool_type => 1,
+        .void_type => 1,
+        .u1_type, .u8_type, .i8_type => 1,
+        .u16_type, .i16_type, .f16_type => 2,
+        .u29_type, .u32_type, .i32_type, .f32_type => 4,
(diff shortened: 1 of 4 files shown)


Techatrix:
I appreciate your work on this issue, however I do not believe that a good solution to #1677 is reimplement the size calculation logic of the Zig compiler in ZLS at this time. Especially when dealing with properties like size and alignment, you would like to put your trust into tools to always tell you the correct answer. But this implementation is too buggy and maintenance heavy that I'd rather wait until we can [query the compiler](https://github.com/ziglang/zig/issues/615) to get reliable results from the source of truth. There is no way around this anyway since some types like auto structs have implementation defined size. The compiler is free to reorder fields, add padding or insert "invisible" fields however it may wish. Any standalone implementation of this in ZLS would be an approximation at best.

ShadowCurse:
I see your points and do agree with them. This compiler query feature you mentioned seems like the exact solution we need here indeed.
Anyway, I appreciate you taking a look at this.

PR #3187, opened 2026-04-26
collect module imports of public modules

iceghost:
Currently, zls only collects module imports of Step.Compile. For public modules, it only collect the root_modules, not their module imports.

This commit makes the build_runner also collects the module imports and also adds one regression test case when one does `addOptions` on a public module, which is my original motivation (and confusion why my `@import("options")` was not working).

The generated options.zig file is generated in the `c/` instead of `o/` namespace, so I also removed the namespace assertion. I assume the original assertion is only for reminding reader of what that component is, and not an actual useful property of the path, so I keep the intent as a comment.

---

Please let me know if this is the correct fix, or was there a reason `b.modules` are treated differently from `Step.Compile`.

Also, without the fix, the added test case with fail with:

```
getPath() was called on a GeneratedFile that wasn't built yet.
  source package path: /workspace/zls/tests/build_runner_cases
  Is there a missing Step dependency on step 'options'?
```

### DIFF
--- a/src/build_runner/build_runner.zig
+++ b/src/build_runner/build_runner.zig
@@ -1078,9 +1078,11 @@ fn extractBuildInformation(
         var modules: std.array_hash_map.Auto(*std.Build.Module, void) = .empty;
         defer modules.deinit(gpa);
 
-        try modules.ensureUnusedCapacity(gpa, b.modules.count());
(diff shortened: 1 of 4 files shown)


Techatrix (approved):
👍

PR #3189, opened 2026-05-01
fix: prevent renaming aliased functions during rename request

### LINKED ISSUE #3184: Rename symbol renames the value instead of the constant binding
### Zig Version

0.16.0

### ZLS Version

0.16.0

nvim v0.11.6 with 'neovim/nvim-lspconfig'

1. Create a file:

```zig
const std = @import("std");
const testName = std.debug.print;

pub fn main() void {
    testName("hello\n", .{});
}
```
2. Rename `testName` into `newTestName`.
3. The result is that the `std.debug.print` function is renamed instead of the `testName` constant, directly in the Zig standard library.
```zig
const std = @import("std");
const testName = std.debug.newTestName;

pub fn main() void {
    testName("hello\n", .{});
}
```

### Expected Behavior

The `testName` constant should be renamed instead of the `std.debug.print`:

```zig
const std = @import("std");
const newTestName = std.debug.print;

pub fn main() void {
    newTestName("hello\n", .{});
}
```

### Attach ZLS log output

<details><summary>zls.log</summary>

```log
info  ( main ): Starting ZLS      0.16.0 @ 'zls'
info  ( main ): Log File:         /home/oleh/.cache/zls/zls.log (info)
info  (server): Client Info:      Neovim (0.11.6)
info  (server): Set config option 'builtin_path' to "/home/oleh/.cache/zls/builtin.zig"
info  (server): Set config option 'zig_lib_path' to "/opt/zig-bin-0.16.0/lib"
info  (server): Set config option 'zig_exe_path' to "/opt/zig-bin-0.16.0/zig"
info  (server): Set config option 'build_runner_path' to "/home/oleh/.cache/zls/build_runner/0536d83f23cda18cf0a02d9cd4e6dce4/build_runner.zig"
info  (server): Set config option 'global_cache_path' to "/home/oleh/.cache/zls"

```

</details>

nikhilcodes777:
Fixes #3184 

### Description
This fixes an issue where triggering a `textDocument/rename` on a local alias of a function (e.g., `const testName = std.debug.print;`) would incorrectly resolve the alias to its source and rename the original declaration (e.g., renaming `print` inside the Zig standard library).

**Root Cause:**
During a rename request, `referencesHandler` and the reference-gathering `Builder` were aggressively resolving aliases using `analyser.resolveVarDeclAlias`. While this is desired for `textDocument/references` or `highlight`, it results in cross-file destruction during a rename.

**Changes:**
- Added a `resolve_aliases` boolean flag to the `Builder` struct in `src/features/references.zig`.
- Conditionally bypassed `resolveVarDeclAlias` in both `referencesHandler` and `Builder.referenceNode` if `request == .rename`.
- This ensures that a rename operation strictly targets the literal identifier the user initiated the rename on, preserving the underlying source declaration.

### Environment / Tested On
- `zig 0.16.0`
- `zig 0.17.0-dev.215+8c5542bd3`
- `emacs/neovim`

### Test Case Used
```zig
const std = @import("std");

const foo = struct {
    pub fn print(comptime fmt: []const u8, args: anytype) void {
        std.debug.print(fmt, args);
    }
};

const fein = foo.print;

pub fn main() void {
    // Renaming `fein` here no longer renames `foo.print`
    fein("Hello world \n", .{});
}
```

### DIFF
--- a/src/features/references.zig
+++ b/src/features/references.zig
@@ -70,6 +70,7 @@ const Builder = struct {
     did_add_target_symbol: bool = false,
     analyser: *Analyser,
     encoding: offsets.Encoding,
+    resolve_aliases: bool = true,
 
     fn add(self: *Builder, handle: *DocumentStore.Handle, token_index: Ast.TokenIndex) error{OutOfMemory}!void {
         if (self.target_symbol.handle == handle and
@@ -235,8 +236,9 @@ const Builder = struct {
             },
             else => return,
         };
-
-        candidate = try builder.analyser.resolveVarDeclAlias(candidate) orelse candidate;
+        if (builder.resolve_aliases) {
+            candidate = try builder.analyser.resolveVarDeclAlias(candidate) orelse candidate;
+        }
 
         if (builder.target_symbol.eql(candidate)) {
             try builder.add(handle, name_token);
@@ -296,6 +298,7 @@ fn symbolReferences(
         .target_symbol = target_symbol,
         .local_only_decl = local_node != null,
         .encoding = encoding,
+        .resolve_aliases = request != .rename,
     };
 
     blk: {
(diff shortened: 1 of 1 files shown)


techatrix:
This change is as expected besides this particular case:

```zig
const S = struct {
    fn foo() void {}
    // ^^^ rename here
};
comptime {
    const foo = S.foo;
    //    ^^^ should be changed
    foo(); // <-- should also changed
}
```

The behavior should be different depending on if the rename request has been done directly on the declaration compared to an alias. While comparing this PR with #3190 to figure out the right behavior, I ended up implementing this myself in #3199 which has test coverage and handles the previously mentioned edge-case.

PR #3190, opened 2026-05-02
fix: handle analysis of decls more correctly

### LINKED ISSUE #3184: Rename symbol renames the value instead of the constant binding

neoto:
Handles references to aliased decls more correctly by distinguishing between containers and runtime types. Compared to
https://github.com/zigtools/zls/pull/3189 which special cases renames, this _should_ address the root cause, fixing both renaming and highlighting symbols. I also added regression tests that fail without the changes on this branch.

Closes https://github.com/zigtools/zls/issues/3184

### DIFF
--- a/src/analysis.zig
+++ b/src/analysis.zig
@@ -755,6 +755,15 @@ pub fn resolveVarDeclAlias(analyser: *Analyser, decl: DeclWithHandle) Error!?Dec
     }
 }
 
+/// Like `resolveVarDeclAlias`, but only follows the alias when the resolved
+/// symbol is a type value (things like containers and imported modules).
+pub fn resolveVarDeclAliasType(analyser: *Analyser, decl: DeclWithHandle) Error!?DeclWithHandle {
+    const resolved = try analyser.resolveVarDeclAlias(decl) orelse return null;
+    const ty = try resolved.resolveType(analyser) orelse return null;
+
+    return if (ty.is_type_val) resolved else null;
+}
+
 /// resolves `@field(lhs, field_name)`
 pub fn resolveFieldAccess(analyser: *Analyser, lhs: Type, field_name: []const u8) Error!?Type {
     const binding = try analyser.resolveFieldAccessBinding(.{ .type = lhs, .is_const = false }, field_name) orelse return null;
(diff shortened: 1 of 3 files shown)


techatrix:
I prefer the behavior provided by #3189 over this solution. Because of alises, renaming should not necessarily behave identical to references. Consider this example:

```zig
const S = struct {
    fn foo() void {}
};
comptime {
    const foo = S.foo;
    foo();
}
```
Works as expected:

1. Renaming `foo()` should not rename `fn foo` and `S.foo`

Doesn't work as expected:

2. Asking for references on `foo()` should also provide `fn foo` and `S.foo`
3. Asking for references on `fn foo` should also provide `const foo` or `foo()`
3. Renaming `fn foo` should also rename `const foo` or `foo()`

The preferred behavior of the second item could be debated but 3 & 4 I feel more confident about.

neoto:
Aight. That still leaves highlighting borked, which that PR doesn't fix

PR #3192, opened 2026-05-04
fix ArrayList getLast calls

l1yefeng:
https://codeberg.org/ziglang/zig/commit/d02d0b879c143e1495fc360845d6feec1ac293ed has updated the prototype of `getLast` and `getLastOrNull`.

### DIFF
--- a/src/features/document_symbol.zig
+++ b/src/features/document_symbol.zig
@@ -69,7 +69,7 @@ pub fn getDocumentSymbols(
             },
         };
 
-        try stack.append(arena, stack.getLast());
+        try stack.append(arena, stack.items[stack.items.len - 1]);
         const stack_entry: *StackEntry = &stack.items[stack.items.len - 1];
 
         const symbol: Symbol = switch (tree.nodeTag(node)) {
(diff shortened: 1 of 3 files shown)


src/features/document_symbol.zig
```diff
@@ -69,7 +69,7 @@ pub fn getDocumentSymbols(
             },
         };
 
-        try stack.append(arena, stack.getLast());
+        try stack.append(arena, stack.items[stack.items.len - 1]);
```
nektro:
```suggestion
        try stack.append(arena, stack.getLast().?);
```

mochalins:
@l1yefeng Did you test locally with `zig build test`? I do believe that this patch should also include a change to `src/build_runner/build_runner.zig` in order to pass tests.

l1yefeng:
Thanks @mochalins, you're right. Updated and passed the tests.

Techatrix (approved):
@l1yefeng Make sure to create and push to a different branch branch when making a pull request instead of using the default "master" branch.

PR #3195, opened 2026-05-08
fix inefficient semanticTokens/range at the end of files

### LINKED ISSUE #1977: inefficient semanticTokens/range on the end of the document
Apply the following patch to `semanticTokens/range` and print which ast nodes will be traversed to generate semantic tokens.
```patch
diff --git a/src/Server.zig b/src/Server.zig
index 567de571..fec62c81 100644
--- a/src/Server.zig
+++ b/src/Server.zig
@@ -640,7 +640,7 @@ fn initializeHandler(server: *Server, arena: std.mem.Allocator, request: types.I
             },
             .semanticTokensProvider = .{
                 .SemanticTokensOptions = .{
-                    .full = .{ .bool = true },
+                    // .full = .{ .bool = true },
                     .range = .{ .bool = true },
                     .legend = .{
                         .tokenTypes = std.meta.fieldNames(semantic_tokens.TokenType),
index dfd7b1a0..69fd55d8 100644
@@ -1048,6 +1048,7 @@ pub fn writeSemanticTokens(
 
     // reverse the ast from the root declarations
     for (nodes) |child| {
+        std.debug.print("node: {d}\n", .{child});
         try writeNodeTokens(&builder, child);
     }
```

Then open a large file like `Sema.zig` or `AstGen.zig` and the move to the end of document.
ZLS will generate tokens for the entire document.
```
node: 3
node: 5
node: 7
node: 9

...

node: 140721
node: 140880
node: 140903
node: 140907
node: 140911
node: 140915
node: 140919
node: 140923
debug (server): Took 6527ms to process request-15-textDocument/semanticTokens/range on Thread 79111
```

xdBronch:
realized the same thing was happening with inlay hints as well
closes #1977

### DIFF
--- a/src/features/inlay_hints.zig
+++ b/src/features/inlay_hints.zig
@@ -565,15 +565,13 @@ pub fn writeRangeInlayHint(
         .hover_kind = hover_kind,
     };
 
-    const nodes = try ast.nodesAtLoc(arena, &handle.tree, loc);
-
-    for (nodes) |child| {
-        try writeNodeInlayHint(&builder, &handle.tree, child);
-
-        var walker: ast.Walker = try .init(arena, &handle.tree, child);
-        defer walker.deinit(arena);
-        while (try walker.nextIgnoreClose(arena, &handle.tree)) |node| {
+    var walker: ast.Walker = try .init(arena, &handle.tree, .root);
+    defer walker.deinit(arena);
+    while (try walker.nextIgnoreClose(arena, &handle.tree)) |node| {
+        if (offsets.locIntersect(loc, offsets.nodeToLoc(&handle.tree, node))) {
             try writeNodeInlayHint(&builder, &handle.tree, node);
+        } else {
+            walker.skip();
         }
     }
 
--- a/src/features/semantic_tokens.zig
+++ b/src/features/semantic_tokens.zig
@@ -89,6 +89,8 @@ const Builder = struct {
     encoding: offsets.Encoding,
     limited: bool,
     overlappingTokenSupport: bool,
+    /// Only emit tokens for the given source range.
+    loc: ?offsets.Loc,
 
     fn add(self: *Builder, token: Ast.TokenIndex, token_type: TokenType, token_modifiers: TokenModifiers) error{OutOfMemory}!void {
         const tree = &self.handle.tree;
@@ -284,6 +286,10 @@ fn colorIdentifierBasedOnType(
 fn writeNodeTokens(builder: *Builder, node: Ast.Node.Index) Analyser.Error!void {
     const handle = builder.handle;
     const tree = &handle.tree;
+
+    if (builder.loc) |loc| {
+        if (!offsets.locIntersect(loc, offsets.nodeToLoc(tree, node))) return;
+    }
 
     const main_token = tree.nodeMainToken(node);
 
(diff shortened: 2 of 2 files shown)


xdBronch:
~~err, somehow just realizing that this is actually pretty far from ideal... it just happens to be that a couple of the root decls are usually what you want which is what this'll give~~
ok i did something a lot simpler it works just fine

src/features/semantic_tokens.zig
```diff
@@ -1164,18 +1167,21 @@ pub fn writeSemanticTokens(
...
         .limited = limited,
         .overlappingTokenSupport = overlappingTokenSupport,
+        .loc = loc orelse offsets.nodeToLoc(tree, .root),
     };
 
-    var nodes = if (loc) |l| try ast.nodesAtLoc(arena, &handle.tree, l) else handle.tree.rootDecls();
+    var nodes = if (loc) |l| try ast.nodesAtLoc(arena, tree, l) else tree.rootDecls();
     if (nodes.len == 1 and nodes[0] == .root) {
-        nodes = handle.tree.rootDecls();
+        nodes = tree.rootDecls();
```
Techatrix: If `writeNodeTokens` is going to skip nodes on it's own now, why bother doing this initial `nodesAtLoc` call? Doesn't seem like it has any advantage over starting at `tree.rootDecls()` every time.

Did some basic benchmarking and didn't appear like this extra logic still improved performance significantly.

src/features/semantic_tokens.zig
```diff
@@ -1164,18 +1167,21 @@ pub fn writeSemanticTokens(
...
+    const tree = &handle.tree;
+
+    var builder: Builder = .{
         .arena = arena,
         .analyser = analyser,
         .handle = handle,
         .encoding = encoding,
         .limited = limited,
         .overlappingTokenSupport = overlappingTokenSupport,
+        .loc = loc orelse offsets.nodeToLoc(tree, .root),
```
Techatrix: This should be an optional field. The intersection check in `writeNodeTokens` should be skipped in that case because everything intersects with the root node anyway.

src/features/inlay_hints.zig
```diff
@@ -411,6 +412,7 @@ fn writeNodeInlayHint(
     tree: *const Ast,
     node: Ast.Node.Index,
 ) Analyser.Error!void {
+    if (!offsets.locIntersect(builder.loc, offsets.nodeToLoc(tree, node))) return;
```
Techatrix: This check could be moved up the call stack into the `ast.Walker` iteration where we could skip iterating over child nodes as well. This would also make usage of `nodesAtLoc` unnecessary.

```diff
diff --git a/src/features/inlay_hints.zig b/src/features/inlay_hints.zig
index 74a80581..de5cae79 100644
--- a/src/features/inlay_hints.zig
+++ b/src/features/inlay_hints.zig
@@ -161,7 +161,6 @@ const Builder = struct {
     handle: *DocumentStore.Handle,
     hints: std.ArrayList(InlayHint) = .empty,
     hover_kind: types.MarkupKind,
-    loc: offsets.Loc,
 
     fn appendParameterHint(
         self: *Builder,
@@ -412,7 +411,6 @@ fn writeNodeInlayHint(
     tree: *const Ast,
     node: Ast.Node.Index,
 ) Analyser.Error!void {
-    if (!offsets.locIntersect(builder.loc, offsets.nodeToLoc(tree, node))) return;
     switch (tree.nodeTag(node)) {
         .call_one,
         .call_one_comma,
@@ -565,18 +563,15 @@ pub fn writeRangeInlayHint(
         .config = config,
         .handle = handle,
         .hover_kind = hover_kind,
-        .loc = loc,
     };
 
-    const nodes = try ast.nodesAtLoc(arena, &handle.tree, loc);
-
-    for (nodes) |child| {
-        try writeNodeInlayHint(&builder, &handle.tree, child);
-
-        var walker: ast.Walker = try .init(arena, &handle.tree, child);
-        defer walker.deinit(arena);
-        while (try walker.nextIgnoreClose(arena, &handle.tree)) |node| {
+    var walker: ast.Walker = try .init(arena, &handle.tree, .root);
+    defer walker.deinit(arena);
+    while (try walker.nextIgnoreClose(arena, &handle.tree)) |node| {
+        if (offsets.locIntersect(loc, offsets.nodeToLoc(&handle.tree, node))) {
             try writeNodeInlayHint(&builder, &handle.tree, node);
+        } else {
+            walker.skip();
         }
     }
```

Techatrix (changes requested):

xdBronch:
sorry for the poor timing again, won't be home to work on this for a couple weeks

techatrix:
No worries. I could take over If you'd like. Also happy to wait until your back.

xdBronch:
feel free, I'd probably just be applying your suggestions anyway 😄

PR #3198, opened 2026-05-11
Allow C pointers as valid self parameters in method detection

fomgleb:
The language server was not recognizing methods with C pointer receivers (`[*c]T`) as valid methods on their container types. This meant that when working with C interop code, functions like `fuse_opt_parse` declared with `[*c]struct_fuse_args` parameters were not being offered as method completions on instances of `struct_fuse_args`.

Example of generated `cimport.zig`:
```zig
pub const struct_fuse_args = extern struct {
    argc: c_int = 0,
    argv: [*c][*c]u8 = null,
    allocated: c_int = 0,
    pub const fuse_opt_parse = __root.fuse_opt_parse;
    pub const parse = __root.fuse_opt_parse;
};
// Here, generated with `[*c]` pointer.
pub extern fn fuse_opt_parse(args: [*c]struct_fuse_args, data: ?*anyopaque, opts: [*c]const struct_fuse_opt, proc: fuse_opt_proc_t) c_int;
```

```zig
var fuse_args: c.struct_fuse_args = .{};

// This compiles, but no method completion was offered:
_ = fuse_args.parse(&options, &option_spec, null);
```

## Solution

Updated firstParamIs to treat .c pointers the same as .one pointers when matching the container type.

### DIFF
--- a/src/analysis.zig
+++ b/src/analysis.zig
@@ -461,16 +461,16 @@ pub fn firstParamIs(
 
     const deref_type = switch (resolved_type.data) {
         .pointer => |info| switch (info.size) {
-            .one => info.elem_ty.*,
-            .many, .slice, .c => return false,
+            .one, .c => info.elem_ty.*,
+            .many, .slice => return false,
         },
         else => resolved_type,
     };
 
     const deref_expected_type = switch (expected_type.data) {
         .pointer => |info| switch (info.size) {
-            .one => info.elem_ty.*,
-            .many, .slice, .c => return false,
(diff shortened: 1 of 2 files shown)


techatrix:
Please send pull requests to the default branch (i.e. "master") instead of `0.16.x` unless the change does not apply to master branch. Given that I deem this change eligible to be cherry picked into `0.16.x` anyway, it's not a problem in this particular case.

Techatrix (approved):

PR #3219, opened 2026-06-28
Custom server endpoint that returns the list of top-level steps

FrancoisAucoin:
## Quick summary

**Commit message:** Implemented a custom server endpoint that returns the list of top-level steps defined in the specified workspace's root `build.zig` file.

**Note:** I made my changes relative to the [0.16.0](https://github.com/zigtools/zls/tree/0.16.0) release tag, so I will need to rebase against the latest branch and fix the conflicts. But I still wanted to get some feedback to see whether the maintainers would consider this feature a contribution.

## Detailed summary

Let me start by being transparent about why I worked on this feature: I thought that it would be cool to implement a task provider in the [vscode-zig](https://codeberg.org/ziglang/vscode-zig) extension ([here's my fork](https://codeberg.org/FrancoisAucoin/vscode-zig/src/branch/feat/build-steps-task-provider)).

That being said, I browsed the ZLS codebase and realized that pretty much everything was already in place for this to work, as this project's custom build runner is doing the heavy lifting and storing the required information inside a structure called [BuildConfig](https://github.com/zigtools/zls/blob/494486203c3a48927f2383aa3d5ce5fca112186d/src/build_runner/shared.zig#L6).

So here is how the feature works at a high level:

* The LSP client sends a request to `customZig/listBuildSteps` with the following `params` object: `{ workspaceUri: string }`;
* The server, in its private `processMessage` method's [`.other` prong](https://github.com/FrancoisAucoin/zls/blob/ab6db25421b516a6b257490ebd5a5ef91729d1e6/src/Server.zig#L1874), handles the new method;
* The [new handler](https://github.com/FrancoisAucoin/zls/blob/ab6db25421b516a6b257490ebd5a5ef91729d1e6/src/custom_features/list_build_steps.zig#L50) tries to retrieve the appropriate `DocumentStore.BuildFile` instance from the server's `document_store` property; and
* The found `BuildFile` instance's `tryLockConfig` method gets called and, upon success, returns a reference to a `BuildConfig` object, from which the top-level steps (names and descriptions) are extracted and returned from the handler, which the server then sends as a response to the client.

But to make this work decently, I had to make a few extra modifications:

* I added a public method to `DocumentStore` named [primeBuildFile](https://github.com/FrancoisAucoin/zls/blob/ab6db25421b516a6b257490ebd5a5ef91729d1e6/src/DocumentStore.zig#L842), which internally calls the private [getOrLoadBuildFile](https://github.com/FrancoisAucoin/zls/blob/ab6db25421b516a6b257490ebd5a5ef91729d1e6/src/DocumentStore.zig#L810) method for its side effects, because my understanding is that the `build.zig` files are analyzed lazily. Without this, no information would become available to the client until either the root `build.zig` file was opened in the editor, or a user action like code navigation was attempted. While the first call to the endpoint will likely return nothing, the client implementation would still, in theory, be able to prime the server with a dummy request upon connection. I know this isn't ideal, but that's how I solved the problem for now (but I am obviously open to suggestions).
* I modified the [BuildConfig.top_level_steps](https://github.com/FrancoisAucoin/zls/blob/ab6db25421b516a6b257490ebd5a5ef91729d1e6/src/build_runner/shared.zig#L15) property from a `[]const []const u8` type to a `[]const BuildConfig.TopLevelStepInfo` type, where `TopLevelStepInfo` is a simple structure with two fields: `name` and `description` (the goal was to hold the step's description in addition to its name). This schema modification meant that a bunch of `.json` files used for the build runner's test cases also needed to be modified.
* Finally, I added a [brief subsection in the README.md](https://github.com/FrancoisAucoin/zls/blob/ab6db25421b516a6b257490ebd5a5ef91729d1e6/README.md?plain=1#L45) to describe the new endpoint and how to use it.

## How to test the feature (in VS Code)

As mentioned above, I have [forked](https://codeberg.org/FrancoisAucoin/vscode-zig/src/branch/feat/build-steps-task-provider) the [official vscode-zig repository](https://codeberg.org/ziglang/vscode-zig) and implemented a task provider that directly uses this new ZLS endpoint.

To test out the feature:

1. Clone that extension branch.
2. Run `npm install` inside the project root.
3. Open the project in VS Code and launch a debugging session.

**IMPORTANT:** The new feature in my fork of the `vscode-zig` extension is gated behind enablement flags that intentionally default to `false`. I chose this default because I didn't want the server's build runner to automatically spin up in the background without user consent. A user who wants this feature can easily enable it, but that should be their explicit choice.

Additionally, you must configure the extension to point directly to the ZLS binary built from this branch. Your VS Code `settings.json` should look like this:

```json
{
    "zig.zls.path": "/Users/your-username/path/to/your/zls-clone-at-the-pr-branch/zig-out/bin/zls",
    "zig.zls.enableBuildStepTaskProvider": true,
    "zig.zls.enableEagerAnalysis": true
}
```

Here is what the task picker looks like for a basic Zig project in VS Code:

tasks-in-vs-code-command-palette

If the current PR eventually ends up being approved, I will also submit a PR for my [vscode-zig fork](https://codeberg.org/FrancoisAucoin/vscode-zig/src/branch/feat/build-steps-task-provider).

## Thank you for your time!

Thank you for taking some of your precious time to look at this PR. I very much appreciate it! :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,12 @@ The following LSP features are supported:
 - Selection ranges
 - Folding regions
 
+## Custom features
+
+Custom (non-standard) server endpoints:
+
+- **`customZig/listBuildSteps`**: Can be used by the client to request the list of top-level steps for a given workspace. The endpoint expects a mandatory `{workspaceUri: string}` as the `params` object. Internally, the `workspaceUri` property is used to identify the root `build.zig` file for the workspace, from which the top-level steps (and their corresponding descriptions) are extracted and returned as `{name: string, description: string}[]`. The main use case for this endpoint is to allow client IDEs to generate automatic tasks (e.g., to implement a `vscode.TaskProvider` in VS Code) based on the top-level steps in a workspace's root `build.zig` without the IDEs having to implement a custom build runner themselves (since that job is already being done by ZLS).
+
 ## Related Projects
 
 - [`sublime-zig-language` by @prime31](https://github.com/prime31/sublime-zig-language)
--- a/src/DocumentStore.zig
+++ b/src/DocumentStore.zig
@@ -835,6 +835,22 @@ fn getOrLoadBuildFile(self: *DocumentStore, uri: Uri) error{ Canceled, OutOfMemo
     return new_build_file;
 }
 
+/// When a new workspace containing build.zig files is added, those files are loaded lazily.
+/// While this is desirable for most cases, certain edge cases benefit from having
+/// `BuildConfig` objects readily available. This method primes the internal worker process
+/// to immediately analyze the target build.zig file.
+pub fn primeBuildFile(self: *DocumentStore, build_file_uri: Uri) error{ Canceled, OutOfMemory, InvalidBuildFileUri, DocumentStoreDoesNotSupportBuildSystem }!void {
+    if (!isBuildFile(build_file_uri)) {
+        return error.InvalidBuildFileUri;
+    }
+
+    if (!supports_build_system) {
+        return error.DocumentStoreDoesNotSupportBuildSystem;
+    }
+
+    _ = try self.getOrLoadBuildFile(build_file_uri);
+}
+
 /// Opens a document that is synced over the LSP protocol (`textDocument/didOpen`).
 /// **Not thread safe**
 pub fn openLspSyncedDocument(self: *DocumentStore, uri: Uri, text: []const u8) error{ Canceled, OutOfMemory }!void {
--- a/src/Server.zig
+++ b/src/Server.zig
@@ -35,6 +35,8 @@ const hover_handler = @import("features/hover.zig");
 const selection_range = @import("features/selection_range.zig");
 const diagnostics_gen = @import("features/diagnostics.zig");
 
+const list_build_steps = @import("custom_features/list_build_steps.zig");
+
 const BuildOnSave = diagnostics_gen.BuildOnSave;
 const BuildOnSaveSupport = build_runner_shared.BuildOnSaveSupport;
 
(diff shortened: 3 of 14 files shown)


techatrix:
Thank you fork your work on this. This would be a great addition for the VS Code Zig extension. There are however some recent changes  since Zig 0.16.0 that will require a different approach:

The Zig build system has been re-architectured in [ziglang/zig#35428](https://codeberg.org/ziglang/zig/pulls/35428) into separate "maker" and "configurer" processes instead of a single build runner. The configurer is tasked with running the project's `build.zig` and serializing the resulting build graph into `std.Build.Configuration` that can then consumed and executed by the maker. This also allows third-party tools like ZLS or vscode-zig to inspect the build graph directly.

The custom build runner approach is no longer possible but instead, you can implement this directly in the vscode-zig extension without needing ZLS as a proxy. The extension can run `zig build --print-configuration` or `zig build --print-configuration-path` and then parse the output to get the list of top level steps. Unfortunately, documentation on these build system internal is scarce so reading the source code of the reworked build system is vital ([Maker.zig](https://codeberg.org/ziglang/zig/src/branch/master/lib/compiler/Maker.zig), [configurer.zig](https://codeberg.org/ziglang/zig/src/branch/master/lib/compiler/configurer.zig), etc.).

A related issue to unblock ZLS and provide better build system integration for editors can be found here in [ziglang/zig#35538](https://codeberg.org/ziglang/zig/issues/35538). One of my goals for that issue is to integrate that into vscode-zig similar to what is being done here.

Let me know if you have any further questions on this. I'd also suggest to open on issue over at [ziglang/vscode-zig](https://codeberg.org/ziglang/vscode-zig) to keep track of this task if you wish to continue.

PR #3225, opened 2026-07-25
Port to recent Zig 0.17.0-dev std APIs

by965738071:
Adapt to the std breaking changes that landed since the previously pinned Zig version (0.17.0-dev.387):

- lowercase OptimizeMode members (.Debug -> .debug)
- @backingInt/@fromBackingInt replace @intFromEnum/@enumFromInt
- @typeInfo reflection reshape (field_names/field_types, pointer attrs, FieldAttributes)
- Ast.parse now takes ParseOptions (.{ .mode = ... })
- std.Build API changes (LazyPath/Cache.Path, Step.Fail, findProgram options, addPassthruArgs); Graph.zig_lib_directory and Build.cache_root were removed
- new builtin tags (SpirvType, div_ceil, backing_int, from_backing_int) in inlay hints exclusion list
- new std.lang.Type tag 'spirv' handled in error_msg switch
- PtrType gained a duplicate_token field
- relax the max supported Zig version check temporarily so ZLS builds with current dev builds

Also cover previously missed call sites in print_ast.zig, offsets.zig, tests/utility and tests/ErrorBuilder.zig.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -69,8 +69,14 @@ pub fn build(b: *Build) !void {
         test_options.step.name = "ZLS test options";
 
         test_options.addOptionPath("zig_exe_path", .{ .cwd_relative = b.graph.zig_exe });
-        test_options.addOptionPath("zig_lib_path", .{ .cwd_relative = b.fmt("{f}", .{b.graph.zig_lib_directory}) });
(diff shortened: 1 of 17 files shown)


xdBronch:
thank you for the contributions, but your changes are already addressed by the https://github.com/zigtools/zls/commits/techatrix/build-runner-process/ branch which should be merged soon and takes an alternative approach to module resolution. Any change made before that branch are likely just going to delay it by causing conflicts

PR #3226, opened 2026-07-25
Fix/package import resolution

by965738071:


xdBronch:
see https://github.com/zigtools/zls/pull/3225#issuecomment-5079743911

